PR #12, opened 2015-04-09
[WIP] Add multithreaded world manager

mastercoms:
I am attempting to create a multi threaded world manager so that the worlds load and save concurrently. PaperSpigot's async chunk loader would be extremely useful.

There a lot of errors thrown as the server doesn't expect to be multi threaded, but I guess they can be ignored. However, before this is merged, we need to figure out how to fix the errors.

World generation is a bit buggy, but it is more efficient. Apparently, region files are reduced by about 6k, and worlds would load even faster if the server knew how to handle the faster chunk and world loading.

And I added 1 unrelated change: I fixed some mistakes with the library manager.

Also, I reordered some stuff in GlowWorld to make it more efficient, and actually listen to keep-spawn-loaded completely.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -487,8 +487,8 @@ public final class GlowServer implements Server {
             checkTransfer(name, "_the_end", Environment.THE_END);
             createWorld(WorldCreator.name(name + "_the_end").environment(Environment.THE_END).seed(seed).type(type).generateStructures(structs));
         }
-        
-        chunksLoader.worldLoad.countDown();
+
+        chunksLoader.start();
 
         // Finish loading plugins
         enablePlugins(PluginLoadOrder.POSTWORLD);
@@ -1372,7 +1372,7 @@ public final class GlowServer implements Server {
     @Override
     public boolean unloadWorld(String name, boolean save) {
         GlowWorld world = getWorld(name);
-        return world != null && new UnloadWorld(this, world, save).unloadWorld();
+        return world != null && unloadWorld(world, save);
     }
 
     @Override
--- a/src/main/java/net/glowstone/ChunkManager.java
+++ b/src/main/java/net/glowstone/ChunkManager.java
@@ -187,7 +187,7 @@ public final class ChunkManager {
     /**
      * Populate a single chunk if needed.
      */
-    private void populateChunk(int x, int z, boolean force) {
+    public void populateChunk(int x, int z, boolean force) {
         GlowChunk chunk = getChunk(x, z);
         // cancel out if it's already populated
         if (chunk.isPopulated()) {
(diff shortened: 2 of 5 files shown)


mastercoms:
This has been vastly updated.

mastercoms:
Oh um @jimmikaelkael, I just realized this breaks world generation, maybe because chunks are being loaded in a weird way.

jimmikaelkael:
Each chunk can normally be generated independently, this is the goal of the biomegrid and noise, so yes to answer your question only coordinates are taken into account.

mastercoms:
I fixed it somehow. :stuck_out_tongue:

jimmikaelkael:
Great!

mastercoms:
Ok, apparently, Glowstone has an async chunk saver that is not used for some reason...

mastercoms:
Oh nevermind :sob:
chunks are always weird when generated on a thread.

mastercoms:
I finally fixed everything!

I just need to make it efficient by skipping chunks that have already been loaded, and change world loading logic to wait until chunks are done loading.

mastercoms:
Closing this as the gains are minimal at best, and sometimes, it is it slower.

However, I will make a new PR to populate chunks if they are anchored in spawn.

phase:
+1 for keeping the branch, I want to mess around with this and see if I can make it faster/better.

mastercoms:
Well, I think the main issue is the [anvil IO service](https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/8dde1c55105d722c22daa46026aa340e69040ae4/src/main/java/net/glowstone/io/anvil/AnvilChunkIoService.java), and just our handling of region files in general, which should be improved.

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1316,25 +1323,50 @@ private ChunkGenerator getGenerator(String name, Environment environment, WorldT
...
+    GlowWorld nether = null;
+    GlowWorld end = null;
+
     /**
      * Add a world to the internal world collection.
      * @param world The world to add.
      */
     void addWorld(GlowWorld world) {
-        worlds.addWorld(world);
+        int worldIndex = 0;
```
mastercoms: Very :hankey: .

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1316,25 +1323,50 @@ private ChunkGenerator getGenerator(String name, Environment environment, WorldT
...
+        } else if (worldIndex == 2 && world.getEnvironment() == Environment.NETHER) {
+            nether = world;
+        }
+        if (worldIndex == 1 && nether != null) {
+            worlds.addWorld(nether);
+        }
+        if (worldIndex == 2 && end != null) {
+            worlds.addWorld(end);
+        }
+
```
mastercoms: End :hankey:.

Issue #16, opened 2015-04-10
The Nether is flat


jimmikaelkael:
The Nether look is quite boring...
If you agree I should be able to make a relevant PR soon.

mastercoms:
It is not just aesthetics, survival mode players can't really navigate the nether by using its cliffs and archs, since there aren't any.

deathcap:
:+1: would be great to have a real nether

PR #17, opened 2015-04-12
Add Nether generator

jimmikaelkael:
This replaces the old nether generation.
Related issue: #16

### DIFF
--- a/src/main/java/net/glowstone/generator/decorators/overworld/DesertWellDecorator.java
+++ b/src/main/java/net/glowstone/generator/decorators/overworld/DesertWellDecorator.java
@@ -1,7 +1,8 @@
-package net.glowstone.generator.decorators;
+package net.glowstone.generator.decorators.overworld;
 
 import java.util.Random;
 
+import net.glowstone.generator.decorators.BlockDecorator;
 import net.glowstone.generator.structures.GlowDesertWell;
 import net.glowstone.generator.structures.util.StructureBoundingBox;
 import net.glowstone.util.BlockStateDelegate;
--- a/src/main/java/net/glowstone/generator/populators/overworld/BiomePopulator.java
+++ b/src/main/java/net/glowstone/generator/populators/overworld/BiomePopulator.java
@@ -7,7 +7,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 
-import net.glowstone.generator.decorators.DesertWellDecorator;
 import net.glowstone.generator.decorators.overworld.*;
 import net.glowstone.generator.decorators.overworld.FlowerDecorator.FlowerDecoration;
 import net.glowstone.generator.decorators.overworld.TreeDecorator.TreeDecoration;


jimmikaelkael:
It's complete. Just missing physics to flow the lava and some nether fortress structures, but for now it's good.

mastercoms:
I'll merge https://github.com/GlowstoneMC/Glowstone/pull/636 this first.

mastercoms:
Tested this. Beautiful. Nice work.

Except, there is random gravel at around Y: 64.
and biomes are not set properly, though it isn't that important.

src/main/java/net/glowstone/generator/NetherGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+    @Override
+    public short[][] generateExtBlockSectionsWithData(World world, Random random, int chunkX, int chunkZ, BiomeGrid biomes) {
+        final short[][] buf = generateRawTerrain(world, chunkX, chunkZ);
+
+        int cx = chunkX << 4;
+        int cz = chunkZ << 4;
+
+        final double[] surfaceNoise = ((PerlinOctaveGenerator) getWorldOctaves(world).get("surface")).fBm(cx, cz, 0, 0.5D, 2.0D);
+        final double[] soulsandNoise = ((PerlinOctaveGenerator) getWorldOctaves(world).get("soulsand")).fBm(cx, cz, 0, 0.5D, 2.0D);
+        final double[] gravelNoise = ((PerlinOctaveGenerator) getWorldOctaves(world).get("gravel")).fBm(cx, 0, cz, 0.5D, 2.0D);
```
mastercoms: This seems to be generating the gravel, though you probably knew that already :)

jimmikaelkael:
@mastercoms things must have changed, here's what I found in a freshly generated world with a seed:
http://i.imgur.com/kPSRuDr.png

Screenshot taken from vanilla minecraft 1.8.3 in solo.

mastercoms:
Yeah, [you're right](http://minecraft.gamepedia.com/Nether#Naturally_generated). Sorry for not investigating it further.

But this feels wrong to me:
gravel slice in nether

jimmikaelkael:
Like you said before, giving the user more control  on the noises is a how to handle such thing.

mastercoms:
Yes but default values should be good.

Does gravel only spawn near Y: 64, or am I missing something again?

jimmikaelkael:
Yeah, from Y 63 to 65.

mastercoms:
What would happen if that range were to be expanded?

mastercoms:
@deathcap any concerns or suggestions before we merge?

jimmikaelkael:
I don't really undertand your last question... More gravel I guess :)

mastercoms:
Well would those gravel patches appear on a larger variation of heights, with the same thickness or would the gravel patches just get thicker?

jimmikaelkael:
Same tickness on larger height. It's just a noise taken at XZ, if noise treshold is reach Y is between 63 and 65 it places gravel and there's netherrack, but Y isn't influencing this noise.

mastercoms:
Ok. So is there a way to expand the range but decrease the probability of gravel?

src/main/java/net/glowstone/generator/NetherGenerator.java
jimmikaelkael: There, 0.5 is the frequency and 2.0 the amplitude of the fractal.

mastercoms:
I'll test some new values and get back to you, then.

jimmikaelkael:
For expanding the range it's required to modify the loop that build terrain columns.

mastercoms:
So amplitude cannot be increased on its own?

jimmikaelkael:
Amplitude will still influence the fractional brownian motion, but it's 2d noise so Y isn't used. 
But by range I concluded you meant height ?
Maybe I was wrong...

mastercoms:
Ah, I misunderstood what you meant by amplitude.

mastercoms:
I expanded the height range, but I couldn't get the frequency right, it was either too high or too low, so I'm giving up until tomorrow.

src/main/java/net/glowstone/generator/NetherGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+        octaves.put("surface", gen);
+
+        gen = new PerlinOctaveGenerator(seed, 4, 16, 16, 1);
+        gen.setXScale(SURFACE_SCALE / 2.0);
+        gen.setYScale(SURFACE_SCALE / 2.0);
+        octaves.put("soulsand", gen);
+
+        gen = new PerlinOctaveGenerator(seed, 4, 16, 1, 16);
+        gen.setXScale(SURFACE_SCALE / 2.0);
+        gen.setZScale(SURFACE_SCALE / 2.0);
```
jimmikaelkael: The noise scales can be played with too.

jimmikaelkael:
As commented above you can play with noise scales too. Keep in mind that this noise is not very detailed has it only have 4 octaves. More octaves would give more details but less performances.

PR #20, opened 2015-04-12
Liquid physics

mastercoms:
This is from https://github.com/GlowstoneMC/Glowstone/pull/636

Adds flowing water and lava.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1648,7 +1648,7 @@ public final class GlowWorld implements World {
                 GlowBlock block = this.getBlockAt(entry.getKey());
                 BlockType notifyType = itemTable.getBlock(block.getTypeId());
                 if (notifyType != null)
-                    notifyType.recievePulse(block);
+                    notifyType.receivePulse(block);
             }
         }
     }


mastercoms:
I have some major concerns with how this is implemented. I will be moving a lot of code around and changing stuff.

mastercoms:
@deathcap @jimmikaelkael any concerns?

src/main/java/net/glowstone/block/blocktype/BlockDoublePlant.java
```diff
@@ -27,7 +27,6 @@ public boolean canPlaceAt(GlowBlock block, BlockFace against) {
         return false;
     }
 
-    @Override
```
deathcap: Should this override be kept, but add GlowBlockStatecto the parameters?

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -11,13 +14,17 @@
...
+    private static final BlockFace[] dirNESWD = new BlockFace[]{BlockFace.NORTH, BlockFace.EAST, BlockFace.SOUTH, BlockFace.WEST, BlockFace.DOWN};
+    private static final BlockFace[] dirNESWUD = new BlockFace[]{BlockFace.NORTH, BlockFace.EAST, BlockFace.SOUTH, BlockFace.WEST, BlockFace.UP, BlockFace.DOWN};
+
     protected BlockLiquid(Material bucketType) {
         this.bucketType = bucketType;
     }
 
     ////////////////////////////////////////////////////////////////////////////
     // Public accessors
-
```
deathcap: Unnecessary whitespace change

src/main/java/net/glowstone/block/blocktype/BlockDoublePlant.java
jimmikaelkael: Yeah, keep it.

mastercoms:
I did some additional tests and there seems to be some inconsistencies with vanilla.

Now, it's fine for now but it will be important to note that trying to place a liquid on a block face with a attachment block cancels the liquid placement. In vanilla minecraft, the attachment block is removed.

Also, vanilla minecraft liquid does not always flow in all directions, unlike this PR. It has a tendency to follow the depression in the land, or 'flow downwards':

liquids flowing down

mastercoms:
completely* :blush:

SamNosliw:
Hmmm, I completely forgot about that.. That is an interesting mechanic, I'll have to play around with my physics a little more.. Also, one thing I forgot to add in my code, is when water and cobblestone mix there should be a sound effect, I used FIZZ, it sounds correct to me.

Additional bug: When a single bucket of lava is placed in a pool of water, 1 in 5 times the lava will spread before the water and create cobblestone surrounding it. I forgot to fix that one.. HaHa, sorry D:

Also, there are two values in the liquid, each decides the spread speed for lava, and for water, I don't know if they are correct. But I think in vanilla, lava should disappear slower than it spreads? Because that I didn't account for..

Ohh, and I wanted to refractor into LiquidLava/LiquidWater a little more but I didn't ;P

RandomAltThing:
Look at CraftBukkit source for an idea :P

http://github.com/RandomAltThing/CraftBukkit

mastercoms:
That is a 404 page. Also, having CraftBukkit violates DMCA. Also, Glowstone should ideally be a clean room implementation. This liquid physics PR is already pretty far ahead and does not require inspiration or an idea.

RandomAltThing:
Alright

RandomAltThing:
Apparently the repo was taken down as someone on Reddit reported it. I found the source online, however it was only a suggestion to see how it works.

mastercoms:
Spigot hosts its derivation of CraftBukkit on their server.

RandomAltThing:
@mastercoms I mean the pure CraftBukkit source. Also is you **did** want an idea there I's the open source (Obviously don't steal code but you know what I mean) [mc-dev](http://github.com/bukkit/mc-dev)

PR #23, opened 2015-04-12
tweak nether soulsand and gravel gen

mastercoms:
This pull request has been opened to have open discourse about height values and frequency of soul sand and gravel generation in the nether.


deathcap:
Hows this compare to the official Minecraft server?

mastercoms:
I have no idea, the values I put are not exact at all. Just made an issue to discuss what values would be vanilla like.

mastercoms:
So can we agree on some default values?

mastercoms:
I'm doing a rewrite of the world manager and I'll come back to this during that.

PR #24, opened 2015-04-12
Use more modern build status icon

mastercoms:
I feel like the default build status picture looks better, but it is up to @deathcap.


deathcap:
Looks good to me

Issue #27, opened 2015-04-13
Lava and Water is not flowing automatically.


jimmikaelkael:
I noticed lava is not flowing automatically in the nether: trigerring manually an adjacent block update updates physics but I forgot to ask for physic updates in the LavaDecorator class.

The same goes for overworld: needs to fix FlowingLiquidsDecorator class.

Will try to do a related PR soon.

PR #30, opened 2015-04-13
Ask for physics update when placing flowing liquids from biome decorator...

jimmikaelkael:
Fix attempt for #27.
Currently untested.

### DIFF
--- a/src/main/java/net/glowstone/ChunkManager.java
+++ b/src/main/java/net/glowstone/ChunkManager.java
@@ -212,9 +212,9 @@ public final class ChunkManager {
         Random random = new Random(world.getSeed());
         long xRand = random.nextLong() / 2 * 2 + 1;
         long zRand = random.nextLong() / 2 * 2 + 1;
+        random.setSeed((long) x * xRand + (long) z * zRand ^ world.getSeed());
 
         for (BlockPopulator p : world.getPopulators()) {
-            random.setSeed((long) x * xRand + (long) z * zRand ^ world.getSeed());
             p.populate(world, random, chunk);
         }
 


jimmikaelkael:
Just tested it, this fix doesn't work...
@deathcap @mastercoms  any idea ?

mastercoms:
receivePulse needs to be called.

jimmikaelkael:
Shouldn't this be done automatically when calling org.bukkit.block.BlockState update method and asking for physics update ?

mastercoms:
Yeah, we should modify it in glowkit, but I'll pull request a quick fix for now.

mastercoms:
Closing this as a better fix can be done in Glowkit.

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -452,5 +461,6 @@ public void applyPhysics(Material oldType, int newTypeId, byte oldData, byte new
         if (type != null) {
             type.onBlockChanged(this, oldType, oldData, newType, newData);
         }
+        type.receivePulse(this);
```
jimmikaelkael: Not what I expected. But it should work. Any reason we pulse even if applyPhysics is set to false ?

mastercoms:
I don't know the test case for this, so could you test for me and tell the result?

jimmikaelkael:
Not currently I'm not at home :)
On my side I was thinking about inserting the pulse in GlowBlock, after line 455 (current HEAD), and maybe line 462.

I think that's how it worked in bukkit but I'm really not sure.

src/main/java/net/glowstone/block/GlowBlock.java
mastercoms: Well `.update()` with no applyPhysics still tries to update the block, so we should still pulse the block when it is updated.

jimmikaelkael:
Well I'll try the current fix, let me a few minutes to install all what's needed.

mastercoms:
Ok, I pushed your suggestion. When you're ready, merge it into your branch.

mastercoms:
I made an oopsie. Instead of using notifyType, I used type, which didn't exist. I pushed a new thing, pull request it. Sorry.

jimmikaelkael:
Yeah :) It can happen, no problem.

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -451,6 +461,8 @@ public void applyPhysics(Material oldType, int newTypeId, byte oldData, byte new
         BlockType type = itemTable.getBlock(oldType);
         if (type != null) {
             type.onBlockChanged(this, oldType, oldData, newType, newData);
+            type.receivePulse(this);
         }
+
```
mastercoms: also accidental new line

jimmikaelkael:
I'm testing, if it works I'll remove the newline ;)

jimmikaelkael:
Had an exception loop triggering stack overflow, no way to see what's the error :)
I'm reverting to your fix.

jimmikaelkael:
@mastercoms  Well, your initial fix is what works, the other method I suggested either doesn't work, either throw a loop of exceptions.

The current results:
flowing liquids

I guessed why they make lava flowing farther and faster in the nether: I think it's to be able to form the lava falls more quickly to the ground once a chunk is populated, so that the player see it well formed.

I fixed one of my oddities in the latest commit, that was when playing with structures. The only thing I haven't be able to check is the structures that fit on several chunks (there is currently only one: the desert temple). But anyway, that will be the structures that will need a fix if it's broken.

jimmikaelkael:
I've rebased and  cleaned up the PR.

Issue #36, opened 2015-04-14
Player placed leaves should not be decayable


jimmikaelkael:
Currently, the leaves placed by playes are decayable, this shouldn't be the case:
http://minecraft.gamepedia.com/Leaves

This currently have no real incidence, but I'm planning leaf decay so this will need to be fixed.

mastercoms:
Fixed in 53ebe9f67642ba4c7394fc54c5fffda905c3fa85. Thanks!

PR #37, opened 2015-04-14
Mark player placed leaves blocks as non decayable

jimmikaelkael:
Fix for #36.

Please don't pull until I verified it matches vanilla behavior.


jimmikaelkael:
Tested: it works as expected.

mastercoms:

Issue #40, opened 2015-04-15
Mycelium block should spread.


jimmikaelkael:
The mycel block should receive random block ticking, then fade or spread accounting light level.
BlockFadeEvent and BlockSpreadEvent should be called accordingly.

References:
http://minecraft.gamepedia.com/Mycelium
http://minecraft.gamepedia.com/Tick

Will file a PR soon.

TekkitCommando:
:+1:

mastercoms:
+1

thank you for contributing so much to this project!

PR #41, opened 2015-04-15
Fix random block ticking

jimmikaelkael:
There are several issues that needs to be adressed:
- [x] Crops/stems does not uproot when light level is too low (at least in the code, as it'll require to handle light-level from chunks)
- [x] Mushrooms does not uproot when light level is too high
- [x] No fire block tick
- [x] Ice and snow should be melting if there is enough light
- [x] Water should freeze if the light level is 11 or less, there is a solid block nearby, the block above it is air, and the biome temperature is 0.15 or less
- [x] Leaves should decay assuming there is no nearby wood
- [x] Cacti and sugar cane should increase in age; at 15, it will grow another cactus or sugar cane on top as long as the height does not exceed 3 blocks
- [x] Mycelium should spread to a random adjacent dirt block or die
- [x] Lava should attempt to ignite a random block in its vincinity
- [x] Lit redstone ore should turn off
- [x] Vines should spread

Related issues:
- #40 

Notes:
This PR adds light opacities, flammability and  fire resistance for all blocks, hence it'll be possible to use opacities to calculate lighting.

Sources:
http://minecraft.gamepedia.com/Tick

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockCrops.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockCrops.java
@@ -7,7 +7,6 @@ import net.glowstone.entity.GlowPlayer;
 import org.bukkit.CropState;
 import org.bukkit.Material;
 import org.bukkit.block.BlockFace;
-import org.bukkit.event.block.BlockFadeEvent;
 import org.bukkit.event.block.BlockGrowEvent;
 import org.bukkit.inventory.ItemStack;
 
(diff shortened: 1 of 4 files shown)


mastercoms:
http://www.minecraftforge.net/wiki/Plants#Growing_.28Ticks.29

jimmikaelkael:
Growing tick is already implemented in head. The way they are describing is quite confusing as it is valid only for crops. It might be outdated as growth rate calculation is far to be that simplistic :)

mastercoms:
Alright, just thought it would've helped. Sorry.

mastercoms:
Is the next section, automatic breaking, of any use?

jimmikaelkael:
Yeah, could explain why I can't see mushrooms uproot naturally in vanilla.

jimmikaelkael:
@deathcap @mastercoms  Should be good to be pulled.

TekkitCommando:
yay :smile:

PR #63, opened 2015-05-18
Pulse flow in world.pulse()

### LINKED ISSUE #60: Implement tile tick format
At the moment, pulse requests are stored within the GlowWorld, and if the server shuts down while something is waiting for an update, then it will forget this. So if water is spreading or the pulse request is used for something else later (e.g. redstone) then if the server shuts down or the world unloads, when it reloads it will freeze in place and wont update until a player manually updates it.

This isn't a super high priority issue, but (as far as I can see), this can be fixed in 2 ways:

1) When the world unloads, save all of the active pulse requests in the world save.

2) When a chunk is loaded, force pulse request every block that would accept pulse requests. (Although, this wouldn't perfectly replicate vanilla behavior as floating water doesn't update itself when chunks load, and it might be a little in-efficient)
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15468788-when-pulse-requests-unload?utm_campaign=plugin&utm_content=tracker%2F14691067&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14691067&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mastercoms:
Update BlockLiquid, BlockWater, and BlockLava to use the world.pulse() method to be pulsed automatically.

Fixes #60. (and I tested it)

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
@@ -224,10 +224,7 @@ public abstract class BlockLiquid extends BlockType {
     
     @Override
     public void updateBlock(GlowBlock block) {
-        if (isStationary(block.getType())) {
-            block.setType(getOpposite(block.getType()), block.getData(), false);
-        }
-        block.getWorld().requestPulse(block, isWater(block.getType()) ? TICK_RATE_WATER : TICK_RATE_LAVA);
+        updatePhysics(block);
     }
     
     @Override


SamNosliw:
Ummm, how does this help?

mastercoms:
In here, 
https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/GlowWorld.java#L505

It calls updateBlock, not updatePhysics, so that when a lava or water block is called in the next tick, it also calls the liquid updateBlock method, which actually flows the block.

Again, I tested this by placing a block of water, and just after it spread to the 4 adjacent water blocks, I stopped the server. Then, I started the server again, and the liquid flowed.

SamNosliw:
Oooh, didn't see that line there, my bad. Although, that is based on random rick updates? Doesn't that mean it wont activate instantly, this could cause issues with non-survival gameplay? 

Some custom maps create and use floating fluids to their advantage, this would cause them to update regardless?

mastercoms:
Yes, it is random tick, but it works instantly in all my tests, probably because the random tick checks is the block is applicable for a tick first.

PR #64, opened 2015-05-18
Added partial Redstone support.

SamNosliw:
This is a base platform that should pave the way for functioning redstone, it only partially implements GlowBlock methods.

There is _FULL_ vanilla behaviour for:
  -> Redstone Wire
  -> Redstone Torch
  -> Redstone Block
  -> Redstone Lamp
 -> Wood/Stone Buttons
 -> Levers
 -> Noteblocks
 -> Wood/Iron Doors/Trapdoors & Fencegates
 -> TNT?!!

There is _PARTIAL_ support for:
 -> Redstone Repeater/Diode  _\- (Missing 'locking' when powered from side)_

There is _NOT_ support for:
 -> Comparators _\- (I don't even know how these work, LOL)_
 -> Daylight Detectors
 -> Wood/Stone/Iron/Gold Pressure Plates
 -> Droppers/Dispensers & Hoppers
 -> Tripwire & Hooks

Known bugs:
 -> The second door in a double set opens in reverse. _\- (Not sure how to detect what side)_
 -> Noteblocks don't play sound when .play() is called.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
@@ -198,11 +198,12 @@ public class BlockRedstone extends BlockNeedsAttached {
             if (target.getType().isSolid()) {
                 for (BlockFace face2 : ADJACENT) {
                     GlowBlock target2 = target.getRelative(face2);
-                    if (target2.getFace(block) == null) {
-                        BlockType notifyType = itemTable.getBlock(target2.getTypeId());
-                        if (notifyType != null) {
+                    BlockType notifyType = itemTable.getBlock(target2.getTypeId());
+                    if (notifyType != null) {
+                        if (target2.getFace(block) == null) {
                             notifyType.onNearBlockChanged(target2, BlockFace.SELF, block, block.getType(), block.getData(), block.getType(), block.getData());
                         }
+                        notifyType.onRedstoneUpdate(target2);
                     }
                 }
             }
--- a/src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
@@ -94,11 +94,12 @@ public class BlockRedstoneTorch extends BlockNeedsAttached {
         if (target.getType().isSolid()) {
             for (BlockFace face2 : ADJACENT) {
                 GlowBlock target2 = target.getRelative(face2);
-                if (target2.getFace(block) == null) {
-                    BlockType notifyType = itemTable.getBlock(target2.getTypeId());
-                    if (notifyType != null) {
+                BlockType notifyType = itemTable.getBlock(target2.getTypeId());
+                if (notifyType != null) {
+                    if (target2.getFace(block) == null) {
                         notifyType.onNearBlockChanged(target2, BlockFace.SELF, block, block.getType(), block.getData(), block.getType(), block.getData());
                     }
+                    notifyType.onRedstoneUpdate(target2);
                 }
             }
         }
--- a/src/main/java/net/glowstone/block/blocktype/BlockButton.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockButton.java
@@ -62,11 +62,12 @@ public class BlockButton extends BlockAttachable {
         if (target.getType().isSolid()) {
             for (BlockFace face2 : ADJACENT) {
                 GlowBlock target2 = target.getRelative(face2);
-                if (target2.getFace(block) == null) {
-                    BlockType notifyType = itemTable.getBlock(target2.getTypeId());
-                    if (notifyType != null) {
+                BlockType notifyType = itemTable.getBlock(target2.getTypeId());
+                if (notifyType != null) {
+                    if (target2.getFace(block) == null) {
                         notifyType.onNearBlockChanged(target2, BlockFace.SELF, block, block.getType(), block.getData(), block.getType(), block.getData());
                     }
+                    notifyType.onRedstoneUpdate(target2);
                 }
             }
         }
--- a/src/main/java/net/glowstone/block/blocktype/BlockDoor.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockDoor.java
@@ -149,24 +149,14 @@ public class BlockDoor extends BlockType {
 
         return true;
     }
-
+    
     @Override
-    public void afterPlace(GlowPlayer player, GlowBlock block, ItemStack holding, GlowBlockState oldState) {
-        updatePhysics(block);
-    }
-
-    @Override
-    public void onNearBlockChanged(GlowBlock block, BlockFace face, GlowBlock changedBlock, Material oldType, byte oldData, Material newType, byte newData) {
-        updatePhysics(block);
-    }
-
-    @Override
-    public void updatePhysics(GlowBlock me) {
-        GlowBlockState state = me.getState();
+    public void onRedstoneUpdate(GlowBlock block) {
+        GlowBlockState state = block.getState();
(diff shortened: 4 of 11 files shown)


src/main/java/net/glowstone/block/blocktype/BlockLamp.java
```diff
@@ -0,0 +1,34 @@
...
+        updatePhysics(block);
+    }
+
+    @Override
+    public void onNearBlockChanged(GlowBlock block, BlockFace face, GlowBlock changedBlock, Material oldType, byte oldData, Material newType, byte newData) {
+        updatePhysics(block);
+    }
+
+    @Override
+    public void updatePhysics(GlowBlock me) {
```
mastercoms: Why updatePhysics instead of updateBlock?

src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
```diff
@@ -0,0 +1,121 @@
...
+
+        if (connections.isEmpty()) {
+            value.addAll(Arrays.asList(SIDES));
+        } else {
+            value.addAll(connections);
+            if (connections.size() == 1) {
+                value.add(connections.get(0).getOppositeFace());
+            }
+        }
+
```
mastercoms: I think you forgot to remove from connections, because redstone dust never turns off.

src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
```diff
@@ -0,0 +1,61 @@
...
+    protected BlockFace getAttachedFace(GlowBlock me) {
+        return ((RedstoneTorch) me.getState().getData()).getAttachedFace();
+    }
+    
+    public static BlockFace getAttachedBlockFace(GlowBlock block) {
+        return ((RedstoneTorch) block.getState().getData()).getAttachedFace();
+    }
+
+    @Override
+    public void updatePhysics(GlowBlock me) {
```
mastercoms: Make sure you also call `super.updatePhysics(me)` so that it properly breaks when the block it is attached to breaks.

src/main/java/net/glowstone/block/blocktype/BlockLamp.java
SamNosliw: Doesn't updateBlock only call when the block itself is changed in some way? I want to get when redstone around the block changes..

src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
SamNosliw: Connections is just for calculating, its not returned at all afterwards. Redstone dust visually doesn't turn off, for the player, this is a bug that I have fixed on my test server and I have not committed to this fork yet..

src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
SamNosliw: HaHa, oops, didn't think about that.

SamNosliw:
Ok, fixed a bunch of bugs, and added repeaters, click the above comment for a full list ;D

TekkitCommando:
Sweet :+1: I don't have access to my computer right now, or i would test this i wanna try it out.

src/main/java/net/glowstone/block/blocktype/BlockLamp.java
mastercoms: There are moments in ticking and stuff where it calls updateBlock. I guess you could add updateBlock, and have it call updatePhysics

src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
mastercoms: :+1:

mastercoms:
Don't add super.updatePhysics to everything, only blocks that extend something other than BlockType.

mastercoms:
You can't place redstone on slabs, but vanilla, you can.

SamNosliw:
All of them extend (or should have extended) BlockNeedsAttatched. I guess I forgot redstone.
Hmmm, torch loops work but only if the torch is on the opposite side of redstone, because redstone doesnt seem to update that far.. I think..

SamNosliw:
Alright, all fixed up ;D

mastercoms:
Thank you for your quick response and development! I'll be looking for more bugs.

mastercoms:
The [10 Hz redstone clock](https://www.youtube.com/watch?v=POlozq29mLs) does not work.

src/main/java/net/glowstone/block/blocktype/BlockLamp.java
```diff
@@ -0,0 +1,36 @@
...
+        super.updatePhysics(me);
```
mastercoms: This does not need an update from super.

SamNosliw: Aha, thanks fixed this for next.

SamNosliw:
Added buttons and levers.
Hmmm, I know why the clock doesn't work, I have an idea to fix this. On a side-note, after redstone is complete, it would be a really good idea to go through this and make it more efficient, I feel its a little sloppy in that way.

mastercoms:
Once the WIP label is off this, I'll go through the code to optimize it.

SamNosliw:
Ok, I think I have fixed repeaters, I was able to create the clock that you have used as an example first try, although I used a lamp instead of a noteblock as they don't work *yet*.. Also, placing a redstone torch ontop of a redstone block creates a 1 tick pulse, as a quick method for starting the above clock if you want to test it out yourself. (Both vanilla and in my version)

mastercoms:
Any idea on the vertical torch loop?

SamNosliw:
That still doesn't work? Torch clocks I have built seem to work fine.

Based on your image, I think you built it wrong, it looks like there is no connection to the top right, unless there is redstone ontop of the gold blocks. also the top left doesn't connect to the bottom left in any way?

mastercoms:
Yes, there is redstone on the top right, but I forgot the bottom left redstone. Sorry for being so stupid :blush:

SamNosliw:
Oops, my bad, redstone torches and repeaters were not causing an 'extra' update when they changed state. These are redstone specific updates that cause an update to more than just the immediate surrounding blocks, I've used it before.

I'm sorry that I use this same code in multiple places, I'll refactor it later to be nice and tidy ;D

SamNosliw:
Added redstone support TNT, Noteblocks and all the doors (Trapdoors, Iron Doors, Wood Doors, Fencegates).

Known bugs:
-> The second door in a double set opens in reverse. - (Not sure how to detect what side)
-> Doors/Trapdoors/Fencegates update regardless of redstone update. - (I know a roundabout way to fix this)

SamNosliw:
Ok, fixed the bug "Doors/Trapdoors/Fencegates update regardless of redstone update" by adding an 'onRedstoneUpdate' method to BlockType and calling it for redstone related updates. I'll switch other redstone related blocks to this later. As well as figuring out the second bug and refactoring the 'extraUpdate' method, which *REALLY* needs to be done.

TekkitCommando:
Holy Crap good job on this :+1:

TekkitCommando:
You have done a lot thank you

mastercoms:
Yeah, I can't emphasize this enough: _thank you_

mastercoms:
I feel like everyone really wants some redstone, this PR is not perfect, but at least it is something. I didn't see any noticeable performance loss, so I will be merging.

gdude2002:
I noticed that [this PR was referenced](https://github.com/GlowstoneMC/Glowstone/issues/116#issuecomment-109689532) - If you guys think it appropriate, feel free to PR the main project as well

PR #77, opened 2015-07-15
Liquid Flow Redux

### LINKED ISSUE #21: Liquid physics are not perfectly vanilla-like
https://github.com/deathcap/GlowstonePlusPlus/pull/20#issuecomment-92104993

mastercoms:
This is a redo of Glowstone++'s current liquid flow implementation to optimize it and fix #21.

Right now, I feel performance could be a bit better, but the functionality is partially there.

Todo:
- [x] Liquid flowing
- [x] Lava + water mixing
- [x] Lava in nether differences (flows faster and farther)
- [x] Water source block creation
- [x] Water attachment to source block (right now, even if the source block is gone, water still flows)

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
@@ -199,7 +199,7 @@ public abstract class BlockLiquid extends BlockType {
                 return;
             }
         }
-        block.getWorld().requestPulse(block, isWater(block.getType()) ? TICK_RATE_WATER : TICK_RATE_LAVA);
+        block.getWorld().requestPulse(block, isWater(block.getType()) || block.getBiome() == Biome.HELL ? TICK_RATE_WATER : TICK_RATE_LAVA);
     }
     
     @Override


mastercoms:
For some reason, liquids always flow in all directions. Could someone please tell me why? I'm probably doing something very stupid.

mastercoms:
I figured out the issue. Short circuiting if statement.

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -71,143 +68,86 @@ public void receivePulse(GlowBlock block) {
...
+                        case NORTH:
+                        case WEST:
+                            m *= -1;
+                            break;
+                    }
+                    // check for a depression in the land
+                    if (block.getWorld().getBlockAt(block.getX() + (face == EAST || face == WEST ? m : 0), block.getY() - 1, block.getZ() + (face == NORTH || face == SOUTH ? m : 0)).getType() == Material.AIR) {
+                        if (!flowed && calculateTarget(block.getRelative(face), face, block.getType(), state.getRawData())) {
+                            // mark that we found a match already
+                            flowed = true;
```
mastercoms: This doesn't work. Any ideas?

mastercoms:
I got the correct flowing down, and mixing. All I need to do is draining and I'll push this for review.

mastercoms:
Could someone test this? @gdude2002, maybe? I want another pair of eyes on this PR to ensure it is good.

gdude2002:
Looks pretty good, a few things to note though.

Water seems to take some time to calculate before spreading or placing. Additionally, removing the source block acts plain weird, disappearing most of its water at once and slowly removing the rest.

This can cause some interesting-looking glitches. Take a look at the videos I took below.

* http://vurl.com/mZ8Kr
* http://vurl.com/n57LE

Additionally, and I might be wrong on this, doesn't the water/lava reaction *always* produce cobblestone, aside from the times it produces obsidian? Overall, the lava/water stuff looks really weird to me.

* http://vurl.com/pYBNN

mastercoms:
Yeah, I didn't spend much time on the draining to make it look good. Basically, a liquid block sees if there is a liquid block above it, or a liquid block stronger than it horizontally. If not, it sets itself to air.

gdude2002:
Sure, but it should do so at the same rate as everything else. I'm not entirely sure if liquids use blockphysics in vanilla, but honestly it would appear not.

By the way, water/lava rates seem a bit slow, might just be me though. Or maybe it's the calculations taking too long? Had you planned out your methods of efficiency?

mastercoms:
Yeah, I think it might be the rates.

Also, here are the rules for mixing: http://minecraft.gamepedia.com/Lava#Lava_and_water

gdude2002:
Your rules still seem to be a little incorrect then, since it seems like water is solidifying lava without directly flowing into it? I mean, the flows at the end of that lava video there don't really seem to make sense at all

mastercoms:
These are my rules:
* If water flows into lava, and the lava is a source block, create obsidian
* If water flows down into any other lava block, it creates cobblestone
* If lava flows down into water, it creates stone
* If lava flows horizontally into water, it creates cobblestone

mastercoms:
This needs to be optimized before it is merged.

PR #88, opened 2015-08-20
Various Redstone fixes

xxmicloxx:
This PR fixes two bugs:
- Fluids breaking redstone components will now actually cause a redstone update
- Redstone torches can now also be placed when there is a redstone dust on the opposite side


mastercoms:
:+1: I'll test this out and merge it if it works.

mastercoms:
Works, merging.

PR #91, opened 2015-08-31
Entities

mastercoms:
Adds entities, thanks to https://github.com/GlowstoneMC/Glowstone/pull/282. I also added monster egg support, but it only spawns pigs right now.


RandomAltThing:
Maybe we could do something like this: (I'm not sure the API but basic idea)
```javaswitch(entity.getMetadata()){
	case PIG_METADATA:
		GlowPig pig = new GlowPig(target.getLocation());
		break;
	case SHEEP_METADATA:
		GlowSheep sheep = new GlowSheep(target.getLocation()):
		//Then we could make it spawn colours randomly 
		break;
}
```

RandomAltThing:
@mastercoms ^

RandomAltThing:
Maybe, instead we could make it so you do
```java
Entity entity = new Entity(entityId, location);
```
This way we can then send an add entity packet with the entityId. This will then spawn that mob.

@mastercoms

PR #106, opened 2015-10-23
Saves player data when they are kicked.

tsengia:
This is to fix issue #622 from Glowstone: https://github.com/GlowstoneMC/Glowstone/issues/622


mastercoms:
I'm not sure if this bug is in Glowstone++. I'll check in a moment.

mastercoms:
Well, that was not a moment, but yes, this bug exists. Thanks for your pull request.

mastercoms:
I am experiencing a rather strange bug with this. My TNT entity detection code work on a freshly created world, which has not experienced a player save. However, once it is started for a second time, TNT can no longer detect entities.

PR #113, opened 2015-12-07
makes it possible to run a GlowServer as part of another application

jlirochon:
Example of minimal external application :

``` java
package com.acme;

import net.glowstone.GlowServer;

public class AcmeApplication {
    public static void main(String[] args) {
        GlowServer server = GlowServer.createFromArguments(args);

        // we don't want to run a server when called with --version
        if (server == null) {
            return;
        }

        // we could set some additional properties here

        try {
            glowServer.run();
        } catch (BindException e) {
            e.printStackTrace();
        }
    }
}
```

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -97,24 +97,14 @@ public final class GlowServer implements Server {
      */
     public static void main(String[] args) {
         try {
-            // parse arguments and read config
-            final ServerConfig config = parseArguments(args);
-            if (config == null) {
+            final GlowServer server = createFromArguments(args);
+
+            // we don't want to run a server when called with --version
+            if (server == null) {
                 return;
             }
 
-            ConfigurationSerialization.registerClass(GlowOfflinePlayer.class);
-            GlowPotionEffect.register();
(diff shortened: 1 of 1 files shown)


jlirochon:
Please let me know if such a pull request would be acceptable, or if I can do anything to make it more pleasant :)

mastercoms:
I don't think this is a patch suitable for Glowstone++, as it is intended to be its own server software. This patch should be for projects that need it.

twizmwazin:
I don't see how this is a bad patch. This makes sense- make it easier for users to embed a server in their application, without changing normal functionality.

jlirochon:
It could be used as part of your test suite too. For exemple a test class could run a GlowServer to check that the server is listening on tcp port, and replying to some basic commands like "help". This is called smoke testing.

jlirochon:
Open source software often find many use cases that were not intended or expected. I don't see the point of intentionally making it difficult to use by third parties.

jimmikaelkael:
I think this could be useful.

mastercoms:
I guess since it doesn't affect us, we can merge it.

jlirochon:
Thank you @twizmwazin @jimmikaelkael @mastercoms :)

PR #115, opened 2015-12-14
Minecraft is on 1.8.9

Timbers007:
1.8.8 -> 1.8.9


MOMOTHEREAL:
Would also need the appropriate `pom.xml` changes, such as:

```xml
<version>1.8.8-SNAPSHOT</version>
```

mastercoms:
I don't think it matters, since 1.8.9 was a client change, and 1.8.9 clients are compatible with 1.8.8 servers. Also, Bukkit/Spigot is still on 1.8.8. However, I guess since this is a small and inconsequential change, we can add it.

PR #117, opened 2015-12-14
Can use bare hands (without holding an item) to interact with levers and buttons.

aramperes:
This hard fix made me realize that the following check was obsolete (BlockPlacementHandler.java@81):

``` java
if (!Objects.equals(holding, message.getHeldItem())) {
    // above handles cases where holding and/or message's item are null
    // todo: inform player their item is wrong
    return;
}
```

In fact, in the case of the interaction of a lever using bare hands, `holding` would represent an ItemStack of `Material.AIR * 0` while `message.getHeldItem()` would return `null`.

In order to resolve this, I placed an exceptional check to ensure this doesn't occur with bare hands (for this particular scenario).

### DIFF
--- a/src/main/java/net/glowstone/net/handler/play/player/BlockPlacementHandler.java
+++ b/src/main/java/net/glowstone/net/handler/play/player/BlockPlacementHandler.java
@@ -78,7 +78,10 @@ public final class BlockPlacementHandler implements MessageHandler<GlowSession, 
         ItemStack holding = player.getItemInHand();
 
         // check that held item matches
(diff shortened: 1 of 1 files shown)


TekkitCommando:
Thanks for contributing!
I have been working on other projects but may be a good idea to start looking at G++ again myself :P

mastercoms:
Thanks for the help. I guess this is a consequence of us changing to the weird Bukkit behavior of returning Material.AIR x 0 instead of null.

Issue #101, opened 2015-10-12
Stacktrace on survival, adventure, spectator inventory close


ghost:
Java version: 8
OS: Ubuntu 15.04
Glowstone++ version: 1.8.8-SNAPSHOT

**EDIT**: It seems that this only occurs when closing the survival inventory
**EDIT**: Now it occurs on survival, adventure and spectator

When I close my survival inventory the console shows this stack trace (I tried my best to capture it):

```
[SEVERE] Error while handling CloseWindowMessage(id=0) (handler: CloseWindowHandler)
java.lang.IllegalArgumentException: stack amount must be greater than zero
        at org.apache.commons.lang3.Validate.isTrue(Validate.java:155)
        at net.glowstone.entity.GlowHumanEntity.drop(GlowHumanEntity.java:456)
        at net.glowstone.entity.GlowPlayer.drop(GlowPlayer.java:1992)
        at net.glowstone.entity.GlowHumanEntity.closeInventory(GlowHumanEntity.java:395)
        at net.glowstone.net.handler.play.inv.CloseWindowHandler.handle(CloseWindowHandler.java:11)
        at net.glowstone.net.handler.play.inv.CloseWindowHandler.handle(CloseWindowHandler.java:7)
        at com.flowpowered.networking.session.BasicSession.handleMessage(BasicSession.java:80)
        at com.flowpowered.networking.session.BasicSession.messageReceived(BasicSession.java:139)
        at net.glowstone.net.GlowSession.pulse(GlowSession.java:430)
        at net.glowstone.net.SessionRegistry.pulse(SessionRegistry.java:23)
        at net.glowstone.scheduler.GlowScheduler.pulse(GlowScheduler.java:173)
        at net.glowstone.scheduler.GlowScheduler.access$100(GlowScheduler.java:24)
        at net.glowstone.scheduler.GlowScheduler$2.run(GlowScheduler.java:110)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
[...]

PR #136, opened 2015-12-28
Fixes crafting

### LINKED ISSUE #137: Crafting Bug
When I went into the inventory 2 by 2 crafting grid to make some planks, I placed the wood and it stayed for a fraction of a second, then disappeared. I exited the inventory and the logs flew out of me and I picked them up. This bug also happens in the regular crafting table.

bluekelp:
1.8.9 client (vanilla) glitched horribly for me when I tried to craft basic items (LOG -> WOOD -> STICKS).

Fixed the glitches, enhanced inventory code to not lose (some) unused inputs when closing crafting window (if using a crafting table), fixed an item dup bug, and refactored code a tiny bit.

---

EDIT by mastercoms:

fixes #137 and #101

### DIFF
--- a/src/main/java/net/glowstone/inventory/GlowCraftingInventory.java
+++ b/src/main/java/net/glowstone/inventory/GlowCraftingInventory.java
@@ -4,6 +4,7 @@ import net.glowstone.GlowServer;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.inventory.crafting.CraftingManager;
 import org.bukkit.Bukkit;
+import org.bukkit.Material;
 import org.bukkit.event.inventory.InventoryType;
 import org.bukkit.event.inventory.InventoryType.SlotType;
 import org.bukkit.inventory.*;
@@ -111,7 +112,22 @@ public class GlowCraftingInventory extends GlowInventory implements CraftingInve
 
     @Override
     public ItemStack[] getMatrix() {
-        return Arrays.copyOfRange(getContents(), MATRIX_START, getSize());
+        return replaceAirWithNull(Arrays.copyOfRange(getContents(), MATRIX_START, getSize()));
+    }
+
+    /**
+     * CraftingManager expects null for empty slots, not AIRx0. Convert our internal representation to something
+     * compatible to correct an error where after crafting the original inputs were dropped (because the craft()
+     * method failed to find a matching recipe and thus never decremented the input counts - leaving them "hidden"
+     * in the crafting squares but yielding output --> item duplication).
+     */
+    private ItemStack[] replaceAirWithNull(ItemStack[] items) {
+        for (int i = 0; i < items.length; i++) {
+            if (items[i].getType() == Material.AIR) {
+                items[i] = null;
+            }
+        }
+        return items;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


mastercoms:
I can't right click and drag to place items in the crafting box.

src/main/java/net/glowstone/net/handler/play/inv/WindowClickHandler.java
```diff
@@ -42,7 +42,7 @@ private boolean process(final GlowPlayer player, final WindowClickMessage messag
         final InventoryView view = player.getOpenInventory();
         final GlowInventory top = (GlowInventory) view.getTopInventory();
         final GlowInventory bottom = (GlowInventory) view.getBottomInventory();
-        final ItemStack slotItem = view.getItem(viewSlot);
+        final ItemStack slotItem = message.getItem();
```
mastercoms: Could you explain the reason for this change?

mastercoms:
I did a bit of regression testing, and it seems like https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/commit/6c7c822fda3653d96055467e7dac0a163412c2a4 was the last commit where everything worked. I don't remember changing anything related to the inventory after that.

src/main/java/net/glowstone/entity/GlowHumanEntity.java
```diff
@@ -390,14 +390,36 @@ public void closeInventory() {
...
+    private void dropUnusedInputs() {
+        for (int i = 0; i <= getTopInventory().getSlots().size(); i++) {
+            final ItemStack itemStack = getOpenInventory().getItem(i);
+            if (itemStack == null || itemStack.getAmount() == 0 || itemStack.getType() == Material.AIR) {
+                continue;
+            }
+
+            if (isCraftingInputSlot(i)) {
+                drop(itemStack);
+                getOpenInventory().setItem(i, null);
```
mastercoms: CraftBukkit's expected behavior is Material.AIRx0

mastercoms:
I have fixed the crafting issues in https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/commit/0be6bc0a5ec62dbf8aa967d831c7a1c47e6536fc

src/main/java/net/glowstone/net/handler/play/inv/WindowClickHandler.java
bluekelp: the view.getItem(viewSlot) was null, caused the action in WindowLogic something or other to return NOTHING - nerfing most crafting actions/clicks. maybe my change masks a more serious issue elsewhere?

mastercoms: Yeah, and the underlying issue was fixed in https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/commit/0be6bc0a5ec62dbf8aa967d831c7a1c47e6536fc

I had followed a plugin stacktrace incorrectly, as I was trying to fix a NullPointerException error with many plugins that deal with items. I discovered it was because these plugins were expecting air, not null, probably because of a Bukkit rule. So I "fixed" getItem, instead of fixing the true culprit, getItemInHand, which was fixed in above mentioned commit. The previous "fix" is what fixed a mostly stable crafting.

And your contribution probably made it perfect.

PR #141, opened 2016-01-02
[WIP] Add additional sounds to glowkit

twizmwazin:
This is just to support my added sounds from yesterday.


twizmwazin:
Added player sounds to this, uses Glowkit/#8

PR #143, opened 2016-01-06
Fixed wrong nbt key name Criteria -> CriteriaName

kaenganxt:
I tried to load an original minecraft world into Glowstone and it complained about the non-existent scoreboard objective nbt tag "Criteria". 

This PR changes the tag name to "CriteriaName" as it is used in Minecraft Vanilla (-> [Minecraft Wiki](http://minecraft.gamepedia.com/Scoreboard#NBT_format)) to allow scoreboard objectives to be loaded on a Glowstone server.


mastercoms:
Looks good. Thank you for your contribution.

PR #147, opened 2016-01-09
Remove spurious SetWindowContentsMessage packet sent in GlowPlayer constructor

deathcap:
The GlowPlayer() constructor incorrectly called updateInventory(), which
sent the SetWindowContentsMessage packet (aka 0x30 window_items
http://prismarinejs.github.io/minecraft-data/?d=protocol#toClient_window_items).
This occurred before the JoinGameMessage packet (aka 0x01 login
http://prismarinejs.github.io/minecraft-data/?d=protocol#toClient_login),
and was followed by another SetWindowContentsMessage packet:
1. SetWindowContentsMessage (GlowPlayer constructor)
2. JoinGameMessage (player.join())
3. SetWindowContentsMessage  (called by player.join())

The first packet is not sent by the official server and would crash some
clients e.g. mineflayer (see https://github.com/PrismarineJS/mineflayer/issues/353),
since it relies on the second packet to get the player's entity ID.

For this reason, this commit removes the first packet. The player
inventory is still updated on login via the third packet, as expected.

testing notes: https://gist.github.com/deathcap/bde8187261ac0b4cebe7

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -329,7 +329,6 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
 
         //creates InventoryMonitor to avoid NullPointerException
         invMonitor = new InventoryMonitor(getOpenInventory());
-        updateInventory(); // send inventory contents
     }
 
     public void join(GlowSession session, PlayerDataService.PlayerReader reader) {


deathcap:
@mastercoms can you review if my reasoning makes sense, merge if you agree? Originally I filed this as a mineflayer bug https://github.com/PrismarineJS/mineflayer/issues/353 but per the above description I now think it's Glowstone++

mastercoms:
Yeah, it seems that Glowstone++ is wrong. I will test to see if this patch creates problems with any other clients.

mastercoms:
Since the updateInventory was added in https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/commit/7bf6bf5e6c8e79b0750389b99c14c636298c3287#diff-5324b1a2be78a948158dcf3368bcdb10R179 which did not describe any fixes for anything, I think it is ok.

deathcap:
Cool thanks, I'll merge. If it causes issues (seems unlikely) I think they ought to be fixed by adding `updateInventory()` elsewhere since sending packets in object constructors just seems wrong imho.

PR #148, opened 2016-01-15
Added equals and hashCode methods

Paulomart:
Adding `equals` and `hashCode` methods.

Please sugest classes that should be added too.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1828,13 +1828,6 @@ public final class GlowWorld implements World {
 
     @Override
     public boolean equals(Object obj) {
-        if (obj == null) {
-            return false;
-        }
-        if (getClass() != obj.getClass()) {
-            return false;
-        }
-        GlowWorld other = (GlowWorld) obj;
-        return getUID() == other.getUID();
+        return obj == this;
     }
 }


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1820,4 +1820,21 @@ else if (tickMap.containsKey(target))
...
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        GlowWorld other = (GlowWorld) obj;
+        return getUID() == other.getUID();
```
kamcio96: should be equals

kamcio96:
And you can use lombok, it is simpler :)

src/main/java/net/glowstone/GlowWorld.java
Paulomart: CraftBukkit is using `==` to. I think we should do the same?

Paulomart:
Would be, but some classes require custom methods. Such as the Entity

src/main/java/net/glowstone/GlowWorld.java
mastercoms: Yes, use `==`.

kamcio96: Why? O.o

mastercoms: I agreed just because CraftBukkit used `==`, not that I think CraftBukkit is correct on this. 

I feel like it should use equals because the whole point of equals is to check for object equality, not if they are the same instance. So if the UUIDs have the same bit for bit value, they should be considered equal in `GlowWorld.equals()`. If the same world was loaded twice, it would not return true, as a new UUID would be constructed at https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/io/nbt/NbtWorldMetadataService.java#L41. On the other hand, if we had two of the same world, we wouldn't know for sure if they were equal, block for block.

So with that in mind, another thing to note is plugin compatibility. We must use `==` because that is what CraftBukkit does so that is what plugins will expect.

kamcio96: So 
```
return this == obj;
```
is enough there

Paulomart:
I'm not sure what classes are missing an equals methods? All classes that are exposed to the bukkit api?

src/main/java/net/glowstone/GlowWorld.java
Paulomart: I'm not sure if it is possible that one world can have multiple instances. @mastercoms would need to check this?

mastercoms: @Paulomart You can use a multi world manager and duplicaate the world folder. 

@kamcio96 Yes, I think that would be equivalent.

Paulomart: @mastercoms will then change to ==

mastercoms:
I'll take a look at what is missing when I have enough time.

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -604,12 +605,16 @@ public int getCounter() {
...
+    public int hashCode() {
+        return this.y << 24 ^ this.x ^ this.z ^ getWorld().hashCode();
+    }
+
+    public boolean equals(Object obj) {
+        if (obj == null || getClass() != obj.getClass()) {
             return false;
         }
+        GlowBlock other = (GlowBlock) obj;
+        return this.x == other.x && this.y == other.y && this.z == other.z && getWorld().equals(other.getWorld());
```
mastercoms: Could you tell me why you changed this?

Paulomart: CraftBukkit was only comparing x, y, z and the world, not the whole location.

Paulomart:
Marking this as ready. If you have any suggestion, tell me.

mastercoms:
Ok, but I might look for missing equals statements when I have time, tonight or tomorrow night.

mastercoms:
I feel like the PR looks good as is. Thank you for your contribution!

PR #151, opened 2016-01-16
Set InLove property for GlowAgeable only if contained in nbt data

kaenganxt:
The InLove tag may not be present as it is only added when needed in vanilla minecraft. This commit only sets the property of GlowAgeable if the NBT Tag is present. (It defaults to 0)


mastercoms:

PR #155, opened 2016-01-19
Add NBTReadLimiter

kamcio96:
Block NBT exploit with [Slot](http://wiki.vg/Slot_Data)

I tested it with correct items, but I don't have that crasher. If somebody has could you test it?

### DIFF
--- a/src/main/java/net/glowstone/util/nbt/NBTReadLimiter.java
+++ b/src/main/java/net/glowstone/util/nbt/NBTReadLimiter.java
@@ -18,7 +18,7 @@ public class NBTReadLimiter {
     public void read(int length) {
         this.read += length;
         if (this.read > this.limit) {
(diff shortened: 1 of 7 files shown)


src/main/java/net/glowstone/util/nbt/NBTReadLimiter.java
```diff
@@ -0,0 +1,24 @@
...
+    private long read = 0;
+
+    public NBTReadLimiter(long limit) {
+        this.limit = limit;
+    }
+
+    public void read(int length) {
+        this.read += length;
+        if (this.read > this.limit) {
+            throw new IllegalStateException("Readed more than " + this.limit + " bytes from NBT tag");
```
mastercoms: Read, not readed

mastercoms:
Tested with pyCraft and it successfully blocks the packet and kicks the player. Nice work!

PR #159, opened 2016-01-19
Implement iterator method in ServerListPingEvent

kamcio96:
Test code:

```
@Override
public void onEnable() {
    Bukkit.getPluginManager().registerEvents(this, this);
}

@EventHandler
public void onPing(ServerListPingEvent event) {
    System.out.println("Online:");
    Iterator<Player> i = event.iterator();
    int a=0;
    while (i.hasNext()) {
        System.out.println(++a + " " + i.next());
    }

    i = event.iterator();
    try {
        while (i.hasNext()) {
            i.remove();
        }
    } catch (Exception e) {
        System.out.println("Exception: " + e.getMessage());
    }
}
```

Log:

```
13:26:03 [INFO] Online:
13:26:03 [INFO] 1 GlowPlayer{name=kamcio96}
13:26:03 [INFO] Exception: null
```

### DIFF
--- a/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
+++ b/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
@@ -10,11 +10,11 @@ import net.glowstone.util.GlowServerIcon;
 import org.bukkit.entity.Player;
 import org.bukkit.event.server.ServerListPingEvent;
 import org.bukkit.util.CachedServerIcon;
+import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 
 import java.net.InetAddress;
-import java.util.Collection;
-import java.util.Iterator;
+import java.util.*;
 
 public final class StatusRequestHandler implements MessageHandler<GlowSession, StatusRequestMessage> {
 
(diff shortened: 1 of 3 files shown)


mastercoms:
is there a limit to how many players that can be in the iterator?

kamcio96:
no, should be?

deathcap:
I think it's supposed to be unlimited, since it returns all of the online players in order to allow plugins to hide specific online players: https://github.com/GlowstonePlusPlus/Glowkit/blob/bukkit%2Bglowkit/src/main/java/org/bukkit/event/server/ServerListPingEvent.java

```java
    /**
     * {@inheritDoc}
     * <p>
     * Calling the {@link Iterator#remove()} method will force that particular
     * player to not be displayed on the player list, decrease the size
     * returned by {@link #getNumPlayers()}, and will not be returned again by
     * any new iterator.
     *
     * @throws UnsupportedOperationException if the caller of this event does
     *     not support removing players
     */
    @Override
    public Iterator<Player> iterator() throws UnsupportedOperationException {
        throw new UnsupportedOperationException();
    }
```

@kamcio96 can you test, does this PR let plugins remove players from the iterator?

mastercoms:
Spigot has a config option to limit or randomize players, I think.

kamcio96:
So should it support remove() or not?

mastercoms:
It should support remove().

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
```diff
@@ -1,18 +1,21 @@
 package net.glowstone.net.handler.status;
 
 import com.flowpowered.networking.MessageHandler;
+import com.google.common.collect.Iterators;
```
mastercoms: Unused import :)

kamcio96:
Done

PR #163, opened 2016-01-22
Update merging filter for GlowItem. ItemMeta has to equal too.

kamcio96:


mastercoms:
Thank you for this fix!

PR #164, opened 2016-01-22
Implement chest animation and sound.

kamcio96:
Typo, where are you? :smile:

### DIFF
--- a/src/main/java/net/glowstone/block/entity/TEChest.java
+++ b/src/main/java/net/glowstone/block/entity/TEChest.java
@@ -3,21 +3,48 @@ package net.glowstone.block.entity;
 import net.glowstone.block.GlowBlock;
 import net.glowstone.block.GlowBlockState;
 import net.glowstone.block.state.GlowChest;
-import net.glowstone.inventory.GlowInventory;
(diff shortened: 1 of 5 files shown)


src/main/java/net/glowstone/block/entity/TEChest.java
```diff
@@ -20,4 +24,27 @@ public TEChest(GlowBlock block) {
...
+    public void addViewer() {
+        viewers++;
+        if (viewers == 1) {
+            updateInRange();
+            block.getWorld().playSound(block.getLocation(), Sound.CHEST_OPEN, 5f, 2f);
+        }
+    }
+
+    public void removeViewer() {
+        viewers--;
```
gabizou: You should probably check that the viewers is not already 0.

kamcio96: Player cannot close without open.
Or I don't know something.

mastercoms:
There a few bugs if you are viewing the chest's inventory and someone else breaks the chest block for a double chest, or adds a chest block to create a double chest.

kamcio96:
Know, I will fix this in the meantime

PR #165, opened 2016-01-22
Fix memory leak.

kamcio96:


mastercoms:
This is called by plugins, yes? Then what if many plugins register channels? Will the last plugins stop working?

MeltedLux:
Same way that this is fixed in Spigot. Plugins >128 will stop working but if you're using that many channels there's something wrong.

kamcio96:
No! It is called from payload packet handler, client can send REGISTER message

Issue #170, opened 2016-01-24
Entity destory packet is not send when disconnecting


Paulomart:
When a player gets disconnected or is kicked, the EntitiesDestoryMessage is not broadcasted to all clients. When interacting with the player the servers notices that there is no entity: `Player Paulomart tried to attack an entity that does not exist`

![2016-01-24_12 34 29](https://cloud.githubusercontent.com/assets/4148404/12535990/f985d70e-c296-11e5-9440-bcf03f96f81e.png)

Paulomart:
Is it possible that no entities are removed? I think it _should_ be done it `pulse()` but could not find anything?

mastercoms:
fixed in https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/commit/e326e50342560543fbbd5c42ebd03d9b19e54daa

PR #174, opened 2016-01-24
Send DestroyEntitiesMessage when entity is removed

Paulomart:
Fix for #170

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -508,7 +508,7 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
         List<Integer> destroyIds = new LinkedList<>();
         for (Iterator<GlowEntity> it = knownEntities.iterator(); it.hasNext(); ) {
             GlowEntity entity = it.next();
-            if (isWithinDistance(entity)) {
+            if (isWithinDistance(entity) && !entity.isDead()) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -695,6 +696,10 @@ public void remove() {
         active = false;
         world.getEntityManager().unregister(this);
         server.getEntityIdManager().deallocate(this);
+        DestroyEntitiesMessage message = new DestroyEntitiesMessage(Arrays.asList(id));
+        for (GlowPlayer player : getWorld().getRawPlayers()) {
+            player.getSession().send(message);
```
mastercoms: Check if player can see?

Paulomart: oh sure

kamcio96:
It isn't good solution

mastercoms:
Why not?

kamcio96:
Eg plugin can call remove few times.
It should be do here: https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/entity/GlowPlayer.java#L508

Paulomart:
Was searching for this, will update pr

kamcio96:
You should work around that method, you can override it or change https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/entity/GlowPlayer.java#L511

PR #177, opened 2016-01-24
no midair fire

akilan98:
Blocks dont get set on fire unless they are leaves or they have a block under them

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
@@ -47,22 +47,9 @@ public class ItemFlintAndSteel extends ItemTool {
         if (fireBlock.getType() != Material.AIR) {
             return true;
         }
-        player.sendMessage(clicked.getType().name());
-        if (!(clicked.getType() == Material.LEAVES
-                || clicked.getType() == Material.LONG_GRASS
-                || clicked.getType() == Material.YELLOW_FLOWER
-                || clicked.getType() == Material.RED_ROSE
-                || clicked.getType() == Material.DOUBLE_PLANT
-                || clicked.getType() == Material.DEAD_BUSH
-                || clicked.getType() == Material.VINE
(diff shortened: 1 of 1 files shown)


mastercoms:
why was glowkit changed?

akilan98:
You committed over it while I was working. It shouldn't change.

mastercoms:
Closing as fix is not needed.

src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
```diff
@@ -48,6 +48,10 @@ private boolean setBlockOnFire(GlowPlayer player, GlowBlock clicked, BlockFace f
             return true;
         }
 
+        if (clicked.getType() != Material.LEAVES && fireBlock.getRelative(BlockFace.DOWN).getType() == Material.AIR) {
```
mastercoms: Shouldn't this check if the type is not flammable? There are more blocks than leaves that allow for this.

akilan98:
Need to add other side burnable things. (Flowers don't have an enum constant :( )

akilan98:
Flowers don't get lit on the side

kamcio96:
https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/Material.java#L851 ?

akilan98:
see this is invalid

akilan98:
isFlammable checks if the block is flammable (dirt is) and then lights but it doesnt check if it's lighting on top of an object. It should not light in midair. It should however light on the side of some objects. If it's not an object that gets lit on the side there should be no fire

mastercoms:
I don't see dirt in the switch statement for isFlammable..

akilan98:
Actually the list I have should be replaced with isBurnable(). Those items can be lit on the side/bottom.

jimmikaelkael:
Why not to use a static collection for the material list instead of doing so much OR ?

akilan98:
@jimmikaelkael I want to use isBurnable() but it's not working when I tested...

PR #178, opened 2016-01-25
Add missing MetadataIndexes and MetadataTypes.

Paulomart:
Based on http://wiki.vg/Entities


Paulomart:
`Index ARMORSTAND_LEFT_LEG_POSITION(ArmorStand) conflicts with NO_AI(LivingEntity)` Ehrrr, blame Mojang for this one..

kamcio96:
Add here https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/master/src/test/java/net/glowstone/entity/MetadataIndexTest.java#L50

mastercoms:
Why can other metadata have the same index and not conflict?

Paulomart:
Because they are other Entity, it checks the hierarchy

mastercoms:
Ok, looks good to me.

Issue #190, opened 2016-02-08
Remapping dependencies


Paulomart:
When connecting to a glowstone server via bungee this happens and you cannot connect to the server.

```
15:17:19 [SEVERE] Error in network input
io.netty.handler.codec.DecoderException: java.lang.IndexOutOfBoundsException: readerIndex(3) + length(80) exceeds writerIndex(11): UnpooledUnsafeDirectByteBuf(ridx: 3, widx: 11, cap: 11)
at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:99)
at io.netty.handler.codec.MessageToMessageCodec.channelRead(MessageToMessageCodec.java:111)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:163)
at io.netty.handler.codec.ByteToMessageCodec.channelRead(ByteToMessageCodec.java:108)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:787)
at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:125)
at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:511)
at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:468)
at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:382)
at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:354)
at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:116)
at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137)
at java.lang.Thread.run(Thread.java:745)
[...]

PR #194, opened 2016-02-10
Add relocations for internal dependencies. Closes GH-190

deathcap:
Add relocations for internal dependencies. Closes GH-190

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,8 +152,16 @@
                         <configuration>
                             <relocations>
                                <relocation>
+                                   <pattern>org.objectweb.asm</pattern>
+                                   <shadedPattern>net.glowstone.libs.org.objectweb.asm</shadedPattern>
+                               </relocation>
+                               <relocation>
                                     <pattern>com.flowpowered</pattern>
                                     <shadedPattern>net.glowstone.libs.com.flowpowered</shadedPattern>
+                                </relocation>
+                               <relocation>
+                                    <pattern>jline</pattern>
+                                    <shadedPattern>net.glowstone.libs.jline</shadedPattern>
                                 </relocation>
                                 <relocation>
                                     <pattern>io.netty</pattern>


deathcap:
Are there other dependencies which should be relocated similarly? Not libraries intended for plugin use (part of Bukkit API). Maybe asm and jline too?

mastercoms:
I think asm and jline should be relocated as well.

kamcio96:
Nooo, nothing should be reloacated (or only flownetworking). If plugin use custom lib, it should include that lib.

Paulomart:
@kamcio96 but if a plugins the server to have netty 4, but it has netty 5?

kamcio96:
so plugin has to include netty 4

Paulomart:
Still it's an issue with custom injection, see my ticket.

kamcio96:
I wrote, here is bug: https://github.com/GlowstonePlusPlus/Glowkit/blob/bukkit%2Bglowkit/src/main/java/org/bukkit/plugin/java/PluginClassLoader.java#L97
:stuck_out_tongue_winking_eye:

Paulomart:
But if we relocate Netty and flow it wouldn't be an issue for anyone, the plugin needs to have it itself.

Paulomart:
I know but this way class lookups are faster because most of the classes are in the global class loader (I think?)

kamcio96:
It is really small difference

Paulomart:
Okey, but even then I don't see your issue with relocating

kamcio96:
There is no issue, ofc we can relocate it like CraftBukkit 1.7.10 (in 1.8 md_5 drop that idea and use raw package). It is only my opinion. I don't like upload plugin update's with more 2Mb.

But still, class in glowkit should be changed.

Paulomart:
You are saying that you don't like to upload big plugin files, but also saying that if a plugin is using netty, for example, it should be shaded in the plugin. This doesn't makes sense to me, sorry. The plugin would have to include it either way it you would want to make it right. Then there wore no problems with versions.

mastercoms:
Let's add a lib API? So plugins can request for a library to be downloaded.

Paulomart:
It's a very good, ideas, but then there must be support for custom sources, loading the plugin main without libs from that plugin and so on. I think this would be over engineered. Also there must be support for this in spigot too, otherwise people will just keep shading

PR #197, opened 2016-02-13
Entity Update

lukbukkit:
- Fixed bug when saving villagers and wolves
- Living entities have now a real death animation
- Many living entities now send a metadata package
- If its possible the metadatas of the entities, were saved in the metadata list of the entity class.
- Entities now sending their health to the player
- Entities have now a correct max health

PR #201, opened 2016-02-14
Implementation of ArmorStands

kaenganxt:
This pull requests implements armor stands and covers the following aspects:
- Correct placement of ArmorStands
- Sending of ArmorStand data to the client
- Destroying and interacting with the ArmorStand
- Saving and loading vanilla nbt data for the ArmorStand
- Implementation of Bukkit ArmorStand methods
- Implementation of the following events: PlayerInteractAtEntityEvent, PlayerArmorStandManipulateEvent

Currently the DisabledSlots nbt tag is not supported since bukkit does not provide api methods to manipulate it properly. So I would recommend adding it in a later PR after thinking about an implementation.

### DIFF
--- a/src/main/java/net/glowstone/entity/objects/GlowArmorStand.java
+++ b/src/main/java/net/glowstone/entity/objects/GlowArmorStand.java
@@ -8,6 +8,7 @@ import net.glowstone.GlowWorld;
 import net.glowstone.entity.GlowLivingEntity;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.entity.meta.MetadataIndex;
+import net.glowstone.net.message.play.entity.DestroyEntitiesMessage;
 import net.glowstone.net.message.play.entity.EntityEquipmentMessage;
 import net.glowstone.net.message.play.entity.EntityMetadataMessage;
 import net.glowstone.net.message.play.entity.SpawnObjectMessage;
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/objects/GlowArmorStand.java
```diff
@@ -0,0 +1,491 @@
...
+        for (int i = 0; i < 2; i++) {
+            pose[i] = EulerAngle.ZERO;
+        }
+        double ten = Math.toRadians(10);
+        double fifteen = Math.toRadians(15);
+        double one = Math.toRadians(1);
+        pose[2] = new EulerAngle(-ten, 0, -ten);
+        pose[3] = new EulerAngle(-fifteen, 0, ten);
+        pose[4] = new EulerAngle(-one, 0, -one);
+        pose[5] = new EulerAngle(one, 0, one);
```
Paulomart: Maybe make them a static final constant?

src/main/java/net/glowstone/entity/objects/GlowArmorStand.java
```diff
@@ -0,0 +1,491 @@
...
+    private boolean isVisible = true;
+    private boolean isSmall = false;
+    private boolean hasBasePlate = true;
+    private boolean hasGravity = true;
+    private boolean hasArms = false;
+
+    public GlowArmorStand(Location location) {
+        super(location);
+        setMaxHealth(2);
+        setHealth(2);
```
hibo98: @mastercoms When we pull #197 first then here can be used the new super constructor for GlowLivingEntity

mastercoms:
Waiting for #197 before reviewing this.

mastercoms:
#197 has been pulled.

mastercoms:

mastercoms:
I found that armor stands should only break when a player hits it twice quickly. This PR unfortunately does not do that.

kaenganxt:
It needs two hits if the player is in survival mode, but only one in creative mode. For me, it works fine in this PR.

hibo98:
It works for me too.

mastercoms:
Ah, I was not aware that creative mode took one hit. Never mind then.

PR #210, opened 2016-02-18
Fix horse spawn egg spawning variants

Sataniel98:
In Vanilla, horse spawn eggs only spawn the variants that spawn naturally; in Glowstone++, there's a chance they spawn zombie horses, too.


mastercoms:

PR #211, opened 2016-02-18
Fix getIp returning not the ip the server is bound to.

Paulomart:
The method is currently returning the ip in the config and not the ip
the server was bound too.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -417,6 +417,11 @@ public final class GlowServer implements Server {
     private int port;
 
     /**
+     * The server ip.
+     */
+    private String ip;
+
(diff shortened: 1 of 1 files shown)


gdude2002:
This looks intentional - What have you done to investigate whether it isn't?

Paulomart:
I start my servers using a wrapper panel, it passes the address and the port of the server via `--host <ip> --port <port>`. This works, but in my plugins I use Bukkit#getIP and Bukkit#getPort to pass data to the metrics. If the the ip is emtpy I skip sending metrics. The ip is not configured in the config.yml so the methods returns an empty String. However spigot is returing the IP the server has bound to, even if it is not configured in the server.properties

mastercoms:
Thank you for this fix!

PR #216, opened 2016-02-20
Implemented brodcast for BaseComponents

lukbukkit:
This PR should fix the build error!

### DIFF
--- a/src/test/java/net/glowstone/testutils/ServerShim.java
+++ b/src/test/java/net/glowstone/testutils/ServerShim.java
@@ -2,11 +2,9 @@ package net.glowstone.testutils;
 
 import com.avaje.ebean.config.ServerConfig;
 import net.glowstone.inventory.GlowItemFactory;
+import net.md_5.bungee.api.chat.BaseComponent;
 import org.bukkit.*;
-import org.bukkit.command.CommandException;
-import org.bukkit.command.CommandSender;
-import org.bukkit.command.ConsoleCommandSender;
-import org.bukkit.command.PluginCommand;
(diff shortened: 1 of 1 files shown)


hibo98:
There is an other in the test packages

LukBukkit:
Yes I know, I'll fix this error too

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1338,6 +1335,20 @@ public int broadcastMessage(String message) {
...
 
     @Override
+    public void broadcast(BaseComponent component) {
+        for (Player player : getOnlinePlayers()) {
+            player.spigot().sendMessage(component);
+        }
+    }
+
+    @Override
+    public void broadcast(BaseComponent... components) {
```
kamcio96: You did it wrong. You should use player loop. Do you know why?

LukBukkit: Okay, I'll change this. And no I don't know why.

kamcio96: BaseComponent[] is one message, it is serialized into one string. see BungeeCord code :wink:

Can you squash commits?

LukBukkit: The only reason I could think of is, that If I combine these BaseComponents and send it to every player, it could be faster.

kamcio96:
https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/commit/20bd86f1fbac0077f87ecf6ab08d725c1486b3d7

PR #219, opened 2016-02-24
Implemented Cows, Sheeps and Chickens

lukbukkit:
Hey,
with this PR I want to implement some of the entity functions:
- [x] Milking cows
- [x] Shearing and coloring sheeps
- [x] Chicken drop every 5 - 10 minutes eggs
- [x] Milk buckets are drinkable and remove all effects
- [x] Adding sounds

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemMilk.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemMilk.java
@@ -19,15 +19,14 @@ public class ItemMilk extends ItemFood {
         EventFactory.callEvent(event1);
         if (event1.isCancelled()) return false;
 
-        // todo implement playeritemconsumeevent set item
-
         player.setUsageItem(null);
         player.setUsageTime(0);
-        player.getInventory().getItemInHand().setType(Material.BUCKET);
 
-        for(PotionEffect potionEffect : player.getActivePotionEffects()) {
+        for (PotionEffect potionEffect : player.getActivePotionEffects()) {
             player.removePotionEffect(potionEffect.getType());
         }
+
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/block/itemtype/ItemMilk.java
```diff
@@ -0,0 +1,34 @@
...
+        EventFactory.callEvent(event1);
+        if (event1.isCancelled()) return false;
+
+        // todo implement playeritemconsumeevent set item
+
+        player.setUsageItem(null);
+        player.setUsageTime(0);
+        player.getInventory().getItemInHand().setType(Material.BUCKET);
+
+        for(PotionEffect potionEffect : player.getActivePotionEffects()) {
```
mastercoms: for (

kamcio96: @LukBukkit you should call super, see ItemGoldenApple, and if done set 
```
player.getInventory().getItemInHand().setType(Material.BUCKET);
```

LukBukkit: I don't call super, because:
1. The super method removes one item and I would add this item again, this makes no sense.
2. The FoodLevelChange isn't called, when I drink a milk bucket.

kamcio96: Milk bucket dosen't change food level?

LukBukkit: No it doesn't change food level. It just removes all effects.

mastercoms:

Issue #221, opened 2016-02-28
A few suggestions for the project


twizmwazin:
Hi, I wasn't terribly sure where this would best fit so I figured I'd just make an issue. I haven't really contributed much to the project (although I'd like to), but I have a few suggestions/complaints about the project.
- First, I feel that setup is too complicated. I'd really like a clone-and-go setup. This in my mind would mean splitting off the two submodules and using them as maven dependencies. I really think the current setup is annoying and clunky.
- Second, I feel that Glowkit is rather bloated. It seems to have an everything-and-the-kitchen-sink approach right now, with every possible patch included. For lack of a better word I think it is bloated. I feel that the API should be kept as close to upstream as possible, only having small patches that are absolutely necessary to retain functionality. Using the unmodified Bukkit API would be ideal. I feel efforts should go towards implementing the Bukkit APIs before anything else.
- I'm not really sure about how to go about fixing this, but I also feel that the project is way to closely tied with Bukkit. This could be somewhat detrimental in the long run, making it difficult to split the project from Bukkit.
- A bit petty, but the repo could also be cleaned up a little bit. There are some files in the root that I believe are totally unused.

These are just my suggestions and concerns, I'm curious to see what others agree or disagree on.

Another thing that just popped in to my mind would be a suggestion to move towards a scheduled, stable release model. Currently it is rolling, however a more structured cycle would arguably help the project come together and be closer to releasing a production-ready product.

mastercoms:
1. I mean, they are currently maven dependencies, but we don't have repo.glowstone.net set up to provide them.

1. Yes, Glowkit is bloated so much that there are duplicate methods.

1. kamcio and I are working on abstracting away from Bukkit.

1. I'll look into deleting some unused files

1. I was thinking about starting to version Glowstone++ since we've gotten a lot of bugs fixed and features added so that rolling release is not needed as much.

gdude2002:
@mastercoms Whose maven repo were we going to use again?

mastercoms:
Greatman's

Nicewuerfel:
Sorry if there is some kind of obvious answer to this question, but couldn't you use the Central Repository?

gdude2002:
No, only releases are allowed there, but we need to upload development artifacts.

Nicewuerfel:
Granted, you still need to specify a snapshot repo in the pom.xml, but you can deploy snapshots to oss.sonatype.org. I never understood why so many people feel the need to host their own repos...

[...]

PR #224, opened 2016-03-03
Use upstream flow-network

twizmwazin:
The fork seems kinds pointless, it provides a _slightly_ upgraded version of netty, but no significant improvement. From my perspective, fewer forks makes the project easier to manage. This sorta goes along with my ideas from #221.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,8 +46,8 @@
         </dependency>
         <dependency>
             <groupId>com.flowpowered</groupId>
-            <artifactId>flow-networking</artifactId>
-            <version>1.2.0-SNAPSHOT</version>
+            <artifactId>flow-network</artifactId>
+            <version>1.0.0</version>
         </dependency>
         <dependency>
             <groupId>jline</groupId>
--- a/setup.sh
+++ b/setup.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
 # Setup script for Glowstone++
 (cd Glowkit && mvn install)
-(cd FlowNetworking && mvn install)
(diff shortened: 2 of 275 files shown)


Paulomart:
I'm so sorry github mobile sucks. 

Anyway yes you are right, when we used 5.x it made sense but we are no longer so this is good.

MeltedLux:
CircleCI build fails because of this: https://github.com/GlowstoneMC/GlowstonePlusPlus/blob/master/setup.sh#L4

kamcio96:
(y)

twizmwazin:
Dammit I thought I removed that, one sec

mastercoms:
There are many significant improvements in Netty 4.1 and more are being added from Netty 5.0. I believe Netty 4.1 has more changes than 5.0 now. I would like more justification about downgrading than just because it is cleaner. I will pull request the other changes that we have to their repo.

twizmwazin:
I'll be honest, I'm not sure why you need Netty 4.1, Netty 4.1 features were not used in g++ at all.

mastercoms:
Some benefits are not new APIs and rather improvements to Netty itself. As with Java 8, I will port everything over to use newer features, but that will happen with time.

twizmwazin:
But are these improvements actually significant to the project? What exactly are these improvements?

mastercoms:
https://github.com/netty/netty/wiki/New-and-noteworthy-in-4.1

mastercoms:
I don't think this will be merged.

PR #227, opened 2016-03-05
Added 1.9 enchantments

aramperes:
- Added 2 new enchantments from 1.9: Mending and Frost Walker
- Added 'Rarity' enum instead of the 'weight' magic value
- Added 'treasure' property to enchantments


mastercoms:

PR #231, opened 2016-03-06
Add Baby spawning with spawn eggs.

aramperes:
Adds babies to Animals when clicking on a mob with the same type of Spawn Egg.

GlowCow, GlowPig and GlowSheep need to get their super#entityInteract method called in order to have babies.

### DIFF
--- a/src/main/java/net/glowstone/entity/passive/GlowSheep.java
+++ b/src/main/java/net/glowstone/entity/passive/GlowSheep.java
@@ -69,6 +69,9 @@ public class GlowSheep extends GlowAnimal implements Sheep {
     @Override
     public boolean entityInteract(GlowPlayer player, InteractEntityMessage message) {
         super.entityInteract(player, message);
+
+        if (!isAdult()) return false;
+
         if (player.getGameMode().equals(GameMode.SPECTATOR)) return false;
         if (player.getItemInHand() == null) return false;
         switch (player.getItemInHand().getType()) {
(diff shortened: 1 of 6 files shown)


Paulomart:
There is currently no setParent method, right? If not this would be either way out of scope.

mastercoms:
I think canceling spectator mode is outside of the scope of this PR. I think it would be better done inside our network handlers.

aramperes:
I've added an [API ](https://github.com/GlowstonePlusPlus/Glowkit/pull/15)to Glowkit for baby parenting, since Bukkit did not handle this. Also, this would be used when we get to the Baby AI, to get what entity babies should follow.

src/main/java/net/glowstone/entity/passive/GlowSheep.java
```diff
@@ -68,6 +68,7 @@ private byte getColorByte() {
 
     @Override
     public boolean entityInteract(GlowPlayer player, InteractEntityMessage message) {
+        super.entityInteract(player, message);
         if (player.getGameMode().equals(GameMode.SPECTATOR)) return false;
         if (player.getItemInHand() == null) return false;
         switch (player.getItemInHand().getType()) {
```
mastercoms, line 77: You can't use shears on a baby sheep.

mastercoms:
Merged in https://github.com/GlowstoneMC/GlowstonePlusPlus/commit/9a11c8e42eb5a37afce83e447099a724690a4132. Thank you.

PR #233, opened 2016-03-09
Trapped Chests and Chest Placement

ghost:
This pull request attempts to fix two issues. First of all, trapped chests are not registered as chests in the item table. This makes them not act like chests; notably, they don't have an inventory attached to them. Secondly, normal chests can be placed next to already attached chests (double chests). The same applies to trapped chests when added to the item table.


mastercoms:

PR #235, opened 2016-03-10
Use Class Sapling instead of Class Tree in Sapling

hibo98:


mastercoms:
Thank you!

jimmikaelkael:
I just don't understand why they added a Sapling api in bukkit if you can't get it's age. This sounds stupid as they had everything to add it...

It would have helped to get rid of those ordinal checks which is kind of dirty.

That said, nice addition!

mastercoms:
@jimmikaelkael the person who added the sapling api is @hibo98.

jimmikaelkael:
Ah, this is forgivable then. Sorry for the rude words.

PR #236, opened 2016-03-12
Fixes for Entity metadata not being saved properly, Zombie Villager profession implementation

aramperes:
Since 1.9, a lot of Metadata fields for entities were changed from bytes to booleans. This PR fixes the issues when spawning, saving and changing these entities:
- Zombies & Zombie Pigmen (Fixed spawn and saving)
- Creepers (Fixed saving)
- Ghasts (Fixed state changing)
- Witches (Fixed state changing)
- Pigs (Fixed saving and interacting)

I've also implemented a basic handler for VillagerProfessions for Zombies. Please take note that the ZOMBIE_IS_VILLAGER's value is 0 when not a villager, and any other value to this makes the Zombie a Zombie Villager with a profession of the given ID, with an offset of +1.

### DIFF
--- a/src/main/java/net/glowstone/entity/monster/GlowCreeper.java
+++ b/src/main/java/net/glowstone/entity/monster/GlowCreeper.java
@@ -26,12 +26,12 @@ public class GlowCreeper extends GlowMonster implements Creeper {
 
     @Override
     public boolean isPowered() {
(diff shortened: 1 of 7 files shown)


mastercoms:
Works! Thank you for this fix!

PR #237, opened 2016-03-12
[Moved] Mojangson utility for parsing and writing NBT Tags from/to JSON Text

aramperes:
This class is a utility for parsing and writing NBT compounds and lists from and to JSON text, using the JSONSimple library. There is currently no implemented use of this class in G++, but I will be working on the `summon` command when this utility is done.

"Mojangson" is the term Mojang uses to refer to NBT tags serialized as (almost) JSON, so it can be manipulated by the player using different commands, such as `/summon`, `/entitydata`, etc. The main difference between Mojangson over JSON is the relative flexibility: it does not need double-quotes for JSON element key names.

I am putting this here as I would like to have feedback on the class first before working on an implementation. Parsing and writing Mojangson/JSON using this utility _has_ been tested.

TODO list:
- [x] Transform NBT Tags to JSON
- [x] Transform JSON to NBT Tags
- [x] Transform JSON to Mojangson
- [x] Transform Mojangson to JSON
- [ ] Add suffixes to Mojangson tag values

### DIFF
--- a/src/main/java/net/glowstone/util/nbt/MojangsonUtil.java
+++ b/src/main/java/net/glowstone/util/nbt/MojangsonUtil.java
@@ -2,6 +2,8 @@ package net.glowstone.util.nbt;
 
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


mastercoms:
Would using `o.getClass() == X.class` be better than using instanceof in this case?

Also, would there be a benefit by using Gson instead of JSON-Simple?

aramperes:
@mastercoms 
What lines are you referring to?

To answer your question, I wouldn't see a real advantage on using either: JSON-Simple is 50% faster than GSON for large files (> 100MB), but GSON is faster by 5% for smaller files (<1KB). Which makes JSON-Simple a better solution for general cases, but since we are talking NBT serialization, there is pretty much no difference.

aramperes:
I will mark this as WIP, as there are a couple changes to make because Mojangson people have the brillant idea of accepting key names without double-quotes.

mastercoms:
For Gson vs JSON-Simple, I don't mean in terms of performance, necessarily. But rather, in terms of code simplicity.

aramperes:
@mastercoms 
I have never used Gson in a parsing context, but the writing process seems very similar to JSON-Simple, which I do not personally see an inconvenient in using. Plus, it seems that the whole project uses JSON-Simple, but I do see two Todos to replace J-S by Gson in the future by @kamcio96. In fact, Gson is imported because it is bundled with Vanilla MC and comes with Bukkit/Glowkit  (so does J-S), but Gson has no use in our server implementation currently.

All in all, I wouldn't say that it is an absolute need to use Gson instead of J-S _here_, but if we are going to make a transition, we should probably convert the whole repo at the same time, to make it a more logical decision.

mastercoms:
Yes, a big switch would be more reasonable. Is this still WIP?

aramperes:
I would put this on hold for a moment: Mojangson is actually very different (in terms of flexibility) than JSON, which means it uses more symbols that _may or may not_ be present depending on the type of value. I'll dig more into this before actually calling this any where near final.

aramperes:
@mastercoms 
I have created an [API for Mojangson](https://github.com/momothereal/Mojangson) in a separate repository. How should we implement this inside G++?

mastercoms:
put it in `util.mojangson`

aramperes:
Closing this because a JSON Parser is not appropriate for Mojangson. See new PR #238.

PR #238, opened 2016-03-13
Integration of Mojangson API for conversion from Mojangson to/from NBT.

aramperes:
The [Mojangson library](https://github.com/momothereal/Mojangson/) is a parsing and writing library for NBT Data tags. For more information on the goal of this PR, see #237. The previous PR was closed due to the larger aspect of Mojangson integration than using a JSON parser.

TODO List:
- [x] Integrate Mojangson API
- [x] Add conversion from G++'s NBT API to Mojangson
- [x] Add conversion from Mojangson to G++'s NBT API.
- [x] Add documentation for NBT-Mojangson conversion
- [x] Test all the things...!


Paulomart:
I was thinking, that it might be better to base this on the nbt component api? Instead of reimplementing all the value classes?

aramperes:
@Paulomart 
While both APIs are related to each other, they serve different purposes: for one, the need of an NBT API is for parsing the binary chunk format, which is not available to users without the use of external programs. On the other hand, Mojangson is only the syntax for _user_ input and output of "Data Tags", which are just the uncompressed NBT tags, so they can be used in commands. Since NBT tags and Mojangson have similar components (identical except for the ambiguous IntArrayTag, ByteArrayTag and ListTag for Mojanson, which can be grouped as `MojangsonArray`) and different use, one can be wrapped arounnd the other, which is included in this PR as well.

Edit: Whoops, wrong button

Paulomart:
From what I see, is Monjangson a way to allow users input nbt data. This is the only purpose for it, in glowstone there will be always this: String -> Mojangson -> NBT. Why not cut mojangson out? Do you see a real advantage in creating an extra representation of this? I think that it would make more sense to allow reading a string and give back a nbt tag.

mastercoms:
Well there still would have to be a parser for Mojangson. But I agree that some of the data representation can be used from NBT.

Paulomart:
> Well there still would have to be a parser for Mojangson. But I agree that some of the data representation can be used from NBT.

Yeah that was what meant

aramperes:
@mastercoms  @Paulomart 
How do you see the 'read(String)' method implemented with Tags that require a final value? Using the NBT API makes this less abstract than it was, but I do get the point of the redundancy of value classes.

Paulomart:
Yeah, it makes it less abstract, but I see no real use for it that would require it.

aramperes:
@mastercoms @Paulomart 
Do you think I should follow this style of structure of the API? ('Mojangson.java' is incomplete, will continue to write the parsing API when I get feedback on this)

aramperes:
@mastercoms @Paulomart 
I think this PR is feature-complete and ready to review, _I think_.

Paulomart:
Looks good from code side, the only think is the toMojangson method, I would put it into the parser. Not in the nbt stuff.

aramperes:
@Paulomart Hmm, I put it there because of

> I was thinking, that it might be better to base this on the nbt component api? Instead of reimplementing all the value classes?

Since each Tag has it's own NBT writer and value holder inside of its implementation, why not put NBT->Mojangson translation in the same place (which has value field access)?

Paulomart:
Yeah, It makes sense so too, but having the serializer to both directions makes sense too. You would also hot have a String::toNBTTag() if you could.

aramperes:
@Paulomart 
I am not quite sure what you mean. Should I rename toMojangson() to toNBTTag(), or should I add a method in `Mojangson.java`?

Paulomart:
I would personally move the to method into the util class. But I would like to hear @mastercoms opinion on that.

mastercoms:
I think Mojangson.java should have a method `fromNBT(Tag nbt)`, and `toNBT()`.

src/main/java/net/glowstone/util/mojangson/Mojangson.java
```diff
@@ -10,6 +10,12 @@
 
 public class Mojangson {
 
+    /*
+     * You shall not construct.
+     */
+    private Mojangson() {
+    }
```
mastercoms: Could the `}` be put on the same line as `{`?

github's new see only changes you missed feature bugged and commented this line. see https://github.com/GlowstoneMC/GlowstonePlusPlus/pull/238/files#r58314562

src/main/java/net/glowstone/util/mojangson/Mojangson.java
```diff
@@ -0,0 +1,332 @@
...
+
+import static net.glowstone.util.mojangson.MojangsonToken.*;
+
+public class Mojangson {
+
+    /*
+     * You shall not construct.
+     */
+    private Mojangson() {
+    }
```
mastercoms: Could the `}` be put on the same line as `{`?

mastercoms:
Ok, seems good. @Paulomart any thoughts?

mastercoms:
I think this would be messy, but would it be possible to move toMojangson to fromNBT?

aramperes:
@mastercoms If I can say so myself, that would only add to the mess. [Why?](http://i.imgur.com/TnQRX6v.gifv)

mastercoms:
To minimize changes, but really doesn't amount to much, which is why I said seems good before the suggestion.

Paulomart:
I still dont like the changes to /nbt/. Meaning the `toMojangson()`

mastercoms:
So you would support a unified fromNBT() method in Mojangson.java?

Paulomart:
Mojangson.java::toString(x : NBTCompund)

mastercoms:
That's what I mean by fromNBT :)

Paulomart:
Yeah okey then :+1:

aramperes:
Going to make some new tests for writing and parsing Mojangson.

Paulomart:
You changed some visibility levels, I guess this was needed?

aramperes:
@Paulomart 
It was needed because Mojangson was moved to a different package (`net.glowstone.util.mojangson`).

Paulomart:
Yeah, they wore package intern

src/main/java/net/glowstone/util/mojangson/Mojangson.java
```diff
@@ -319,13 +327,216 @@ public static Tag parseArray(String mojangson) throws MojangsonParseException {
...
 
     /**
-     * Creates a Mojangosn string from the given NBT tag.
+     * Creates a Mojangson string from the given NBT Tag. Convenience method for other fromX(Tag) methods.
+     *
+     * @param tag the NBT Tag to convert
+     * @return the converted Mojangson string
+     */
+    public static String fromTag(Tag tag) {
+        if (tag.getType() == TagType.BYTE) {
```
mastercoms: Would a switch statement be better here? Or see https://github.com/GlowstoneMC/GlowstonePlusPlus/pull/238/files#r59027599

src/main/java/net/glowstone/util/mojangson/Mojangson.java
```diff
@@ -319,13 +327,216 @@ public static Tag parseArray(String mojangson) throws MojangsonParseException {
...
+     * @return the converted Mojangson string
+     */
+    public static String fromByteArrayTag(ByteArrayTag tag) {
+        StringBuilder builder = new StringBuilder();
+        builder.append(ARRAY_START);
+        boolean start = true;
+
+        for (byte value : tag.getValue()) {
+            ByteTag b = new ByteTag(value);
+            if (start) {
```
mastercoms, line 389: Ternary operator?

src/main/java/net/glowstone/util/mojangson/Mojangson.java
```diff
@@ -0,0 +1,542 @@
...
+
+    /**
+     * Detects the Tag type of the Mojangson string, and parses it. Convenience method for other parse methods.
+     * This method will fall back to an IntTag if it could not find an appropriate Tag type, and to String if the value
+     * could not be parsed as an Integer either.
+     * @param mojangson The Mojangson string
+     * @return The parsed NBT Tag
+     * @throws MojangsonParseException if the given Mojangson string could not be parsed.
+     */
+    public static Tag parseTag(String mojangson) throws MojangsonParseException {
```
mastercoms, line 26: I feel like it would be better if these were named toNBT, toNBTInt, etc. I'm not sure on this though.

src/main/java/net/glowstone/util/mojangson/Mojangson.java
```diff
@@ -0,0 +1,542 @@
...
+        return String.valueOf(tag.getValue());
+    }
+
+    /**
+     * Creates a Mojangson string from the given Byte Tag.
+     *
+     * @param tag the Byte Tag to convert
+     * @return the converted Mojangson string
+     */
+    public static String fromByteTag(ByteTag tag) {
```
mastercoms: I think these should still be named fromTag, but have different parameter types, and then you can just remove the huge tree for the method above.

aramperes: How can generic Tags be parsed without knowing what type it is. (As in https://github.com/momothereal/GlowstonePlusPlus/blob/mojangson-api/src/main/java/net/glowstone/util/mojangson/Mojangson.java#L242). Would reflection be used in this case?

mastercoms: So you're asking if, for example, a `ByteTag` were to be casted to `Tag` and then passed into `fromTag`?

We should not force people to use the Tag -> Mojangson parser a certain way. If they want to cast it as a Tag, they probably want it to be treated as one.

As for Mojangson -> Tag, that should still have different method names, as it is required for proper parsing, since all of the parameters are of the same type, `String`.

aramperes: What I mean is, If I were to remove the generic `parseTag`, how would it know what kind to parse, if it is a generic Tag? If all the methods are named `parseTag(XTag)`, and it needs to parse a Tag depending on the type of tag, it cannot resolve it.

``` java
// Without a generic parseTag(Tag) method:
Tag t = something; // As in https://github.com/momothereal/GlowstonePlusPlus/blob/mojangson-api/src/main/java/net/glowstone/util/mojangson/Mojangson.java#L242
String mojangson = parseTag(t); // That wouldn't work without a method that has the generic type as the parameter.
```

src/main/java/net/glowstone/util/mojangson/Mojangson.java
aramperes, line 389: I don't see how to put a Ternary Op there...?

src/main/java/net/glowstone/util/mojangson/Mojangson.java
mastercoms: I am not talking about parseTag. I am talking about fromTag.

aramperes: Sorry, I misread that :s. What I was wanting to say is for cases such as this:
https://github.com/momothereal/GlowstonePlusPlus/blob/mojangson-api/src/main/java/net/glowstone/util/mojangson/Mojangson.java#L418

mastercoms: Either keep the type check or have usages of that method do the type check

PR #240, opened 2016-03-15
Change map with scoreboard entries

kamcio96:
I didn't test it. But should work :wink:


mastercoms:
What if someone changes their name?

kamcio96:
We don't care about it, addPlayer is deprecated :wink:

mastercoms:
So if someone changes their name it will show their previous name? Is that not a bug?

kamcio96:
idk, probably yes
Spigot has 2 string maps

mastercoms:
Closing due to the aforementioned bug.

mastercoms:
Merged, but git isn't detecting it :(. https://github.com/GlowstoneMC/Glowstone/commit/0069bce5c60c62a4ee2cc4f63ea4d43599c278f1

PR #243, opened 2016-03-20
Adds FallingBlock and Bug fix

GreenBeard:
Adds FallingBlock.
Fixes bug with flower pot not dropping items.

Known Bugs with FallingBlock:
~~1. For some reason BlockFalling.onNearBlockChanged() is being called twice instead of once causing two falling blocks to spawn for every one.
2. Red sand appears to be sand when falling. This is because I currently don't know how to add data to falling blocks. I haven't been able to find the documentation for how to do this.~~
Fixed

Also if anyone knows how to put a velocity into an EntityStore that would be very helpful.

### DIFF
--- a/src/main/java/net/glowstone/io/entity/FallingBlockStore.java
+++ b/src/main/java/net/glowstone/io/entity/FallingBlockStore.java
@@ -1,0 +1,41 @@
+package net.glowstone.io.entity;
+
+import net.glowstone.entity.objects.GlowFallingBlock;
+import net.glowstone.util.nbt.CompoundTag;
+
+import org.bukkit.Location;
+import org.bukkit.Material;
+
+class FallingBlockStore extends EntityStore<GlowFallingBlock> {
+
+    public FallingBlockStore() {
+        super(GlowFallingBlock.class, "FallingSand");
+    }
+
+
+    @Override
+    public GlowFallingBlock createEntity(Location location, CompoundTag compound) {
+        // Falling block will be set by loading code below
+        return new GlowFallingBlock(location, null, (byte) 0);
+    }
+
+    @Override
+    public void load(GlowFallingBlock entity, CompoundTag tag) {
+        super.load(entity, tag);
+
+        if (tag.isString("Material")) {
+            entity.setMaterial(Material.getMaterial(tag.getString("Material")));
+        }
+        /*if (tag.isString("Velocity")) {
+            entity.setVelocity(Vector.deserialize());
+        }*/
+    }
+
+    @Override
+    public void save(GlowFallingBlock entity, CompoundTag tag) {
+        super.save(entity, tag);
+
+        tag.putString("Material", entity.getMaterial().toString());
+        //tag.putString("Velocity", entity.getVelocity().serialize());
+    }
+}
--- a/src/main/java/net/glowstone/io/entity/FallingBlockStore.java
+++ b/src/main/java/net/glowstone/io/entity/FallingBlockStore.java
@@ -1,10 +1,10 @@
 package net.glowstone.io.entity;
 
+import net.glowstone.constants.ItemIds;
 import net.glowstone.entity.objects.GlowFallingBlock;
 import net.glowstone.util.nbt.CompoundTag;
 
 import org.bukkit.Location;
-import org.bukkit.Material;
 
 class FallingBlockStore extends EntityStore<GlowFallingBlock> {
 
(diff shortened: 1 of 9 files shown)


gdude2002:
PRs should be single-subject only. I'm not going to close it, but this really should have been two PRs - one for the flower pot fix, and one for `FallingBlock`.

mastercoms:
Seems to be just a one line fix, so I'll accept it for this PR.

GreenBeard:
The solution for the blocks duplicating wasn't very elegant. It was very unusual how the data for the blocks would update, but not the type unless I forced it to pull again from the world.

GreenBeard:
@gdude2002 @mastercoms is it ready to be pulled? Should I change anything?

kamcio96:
Can you merge first and second commit into one?

src/main/java/net/glowstone/io/entity/FallingBlockStore.java
```diff
@@ -0,0 +1,41 @@
...
+    public GlowFallingBlock createEntity(Location location, CompoundTag compound) {
+        // Falling block will be set by loading code below
+        return new GlowFallingBlock(location, null, (byte) 0);
+    }
+
+    @Override
+    public void load(GlowFallingBlock entity, CompoundTag tag) {
+        super.load(entity, tag);
+
+        if (tag.isString("Material")) {
```
mastercoms: Should this be `Block`?

src/main/java/net/glowstone/io/entity/FallingBlockStore.java
```diff
@@ -0,0 +1,41 @@
...
+            entity.setMaterial(Material.getMaterial(tag.getString("Material")));
```
mastercoms: Would this not be a Minecraft formatted block ID? ex. `minecraft:stone`

mastercoms:
It should not drop when falling on plants. Instead, it should remove the plant, without dropping the plant.

Also, anvils will remove any attachable it lands on, like torches.

And I'm fine if this PR does not include damaging with falling block, but please add a todo comment so we can keep track of it.

GreenBeard:
> Should this be Block?

Do you mean should "FallingSand" be "FallingBlock"? If that's what you mean the answer is no. Minecraft never renamed the entity from FallingSand to FallingBlock. If that is not what you meant than what did you mean @mastercoms?

---

**Edit by _gdude2002_**: Please edit your post instead of double-posting; even for a mention.

GreenBeard:
@mastercoms

mastercoms:
No I mean that it seems the tag data for choosing the block texture is called Block, not Material

GreenBeard:
@mastercoms Do you know how I would get the string like minecraft:stone ?

mastercoms:
It would still be tag.getString, but you would filter out minecraft: with a regex.

GreenBeard:
@mastercoms I would still have to set the string in the save function. Where would I get this string from the material?

mastercoms:
toString

GreenBeard:
@mastercoms If what you are saying is true I believe that the reverse lookup would work. Since the `Material.toString()` returns `Material.name()` and the list it tries to find the Material from in `Material.getMaterial(String name)` is the same list as in the `Material.name()` function. I probably should switch `Material.toString()` to `Material.name()` though.

I am not sure I understand the issue you are saying there is (if you are saying there is an issue).

mastercoms:
http://minecraft.gamepedia.com/Chunk_format#Dynamic_tiles

Check under FallingSand. "Block: The Block ID using the alphabetical ID format: minecraft:stone. Only in and after 1.8." But in your falling block store, you expect it to be named Material, instead of Block.

GreenBeard:
@mastercoms I don't understand how I would do what you want. Would I just do:
`tag.putString("Block", "minecraft:" + entity.getMaterial().name());` and `entity.setMaterial(Material.getMaterial(tag.getString("Block").replace("minecraft:", "")));`?

GreenBeard:
@mastercoms I got it, nevermind

GreenBeard:
@gdude2002 @mastercoms I found a really weird bug that I believe isn't due to my code (although it could be). When I change the block and data for a `DOUBLE_PLANT` the data doesn't update properly.

Steps to recreate
1. Place a any `DOUBLE_PLANT`
2. Place a red sand block above it (sand has default data of 0 so it won't work)
3. When the red sand block falls and replaces the `DOUBLE_PLANT` it looks fine but when you destroy the block below the red sand and it falls again the truth is weird. It becomes a sand block instead of a red sand block.

With `DEAD_BUSH` and `LONG_GRASS` the function is as expected and has no issues.

If I am not making sense please tell me and I will attempt to clarify.

mastercoms:
Works great! Thanks again!

PR #244, opened 2016-03-23
Add metrics

kamcio96:
http://mcstats.org/plugin/GlowstonePlusPlus


TheMolkaPL:
Should be optional in the configuration in my opinion.

kamcio96:
plugins\PluginMetrics\config.yml

mastercoms:
Isn't it the plugins job to report metrics info about the server?

kamcio96:
Plugins only report server's name. There we can see more info about it.

Spigot has http://mcstats.org/plugin/Spigot

mastercoms:
Ah, ok.

PR #249, opened 2016-03-24
Add support for more chunk generation options.

GreenBeard:
Specifically ChunkGenerator.generateChunkData(World world, Random random, int x, int z, ChunkGenerator.BiomeGrid biome)
Also fix pointless cast that didn't check `instanceof`


mastercoms:
Looks good, thanks for the contribution!

PR #251, opened 2016-03-25
Fix Grass Decay

ghost:
This pull request aims to fix grass decay as mentioned in https://github.com/GlowstoneMC/GlowstonePlusPlus/issues/84. After taking a peek at the underlying implementation of the same conditional in Spigot, this solution was acquired. Important is to notice that grass doesn't decay no matter what light level, as long as it isn't directly under a block with a certain opacity.


mastercoms:
Thanks

GreenBeard:
Does this fix the issue of grass decaying in some biomes for no reason?

ghost:
Yes, grass no longer decays randomly in vanilla-generated worlds (worlds with lighting data).

PR #252, opened 2016-03-25
Update Grass Spread

ghost:
This pull request includes two changes. The first one is that the main grass spread conditional is fixed. The second one is that a few unnecessary block allocations are removed; cached versions are used instead.


mastercoms:
Nice work! Thank you for your contribution.

PR #254, opened 2016-03-27
Update Mycelium

ghost:
The spreading and decaying of mycelium is broken in the same way they were for grass, since they are handled in the (nearly) exact same way for mycelium. This pull request attempts to fix this by applying the changes that were applied to grass to mycelium. Additionally, some unnecessary block allocations are removed by caching the blocks instead.


mastercoms:
Another welcome fix.

PR #258, opened 2016-03-27
Revert Cache Type

ghost:
As explained in https://github.com/GlowstoneMC/GlowstonePlusPlus/issues/250 the cached type in [GlowBlock](https://github.com/GlowstoneMC/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/block/GlowBlock.java#L57) behaves rather strangely. This causes all kinds of nasty problems. One of these is that an exception is thrown whenever a torch is placed:

```
[SEVERE] Error while handling BlockPlacementMessage(x=-4, y=3, z=26, direction=1, hand=0, cursorX=11, cursorY=16, cursorZ=8) (handler: BlockPlacementHandler)
java.lang.ClassCastException: org.bukkit.material.MaterialData cannot be cast to org.bukkit.material.Torch
    at net.glowstone.block.blocktype.BlockTorch.getAttachedFace(BlockTorch.java:37)
    at net.glowstone.block.blocktype.BlockNeedsAttached.updatePhysics(BlockNeedsAttached.java:21)
    at net.glowstone.block.blocktype.BlockTorch.afterPlace(BlockTorch.java:32)
    at net.glowstone.block.blocktype.BlockType.rightClickBlock(BlockType.java:341)
    at net.glowstone.net.handler.play.player.BlockPlacementHandler.handle(BlockPlacementHandler.java:143)
    at net.glowstone.net.handler.play.player.BlockPlacementHandler.handle(BlockPlacementHandler.java:21)
```

Another issue is that tall grass isn't updated on the client side properly (only the block data is changed) when it absorbs a placed block, as mentioned in https://github.com/GlowstoneMC/GlowstonePlusPlus/issues/245.

The goal of this pull request is simply to get rid of a source of unpredictable bugs. Removing the cached type seems the easiest and cleanest solution, so work can be done on other pieces of code without stumbling on weird behaviour. Of course the type cache can be reimplemented in the future.

### DIFF
--- a/src/main/java/net/glowstone/block/GlowBlock.java
+++ b/src/main/java/net/glowstone/block/GlowBlock.java
@@ -54,7 +54,6 @@ public final class GlowBlock implements Block {
     private final int y;
     private final int z;
     private GlowWorld world;
-    private Material type;
 
     public GlowBlock(GlowChunk chunk, int x, int y, int z) {
         world = chunk.getWorld();
@@ -176,11 +175,7 @@ public final class GlowBlock implements Block {
     // Type and typeid getters/setters
     @Override
     public Material getType() {
-        if (type != null) {
-            return type;
-        }
-        type = Material.getMaterial(getTypeIdNoCache());
-        return type;
+        return Material.getMaterial(getTypeId());
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


mastercoms:
Block type cache is fixed in https://github.com/GlowstoneMC/GlowstonePlusPlus/commit/85e98f311f766f18debf17c0ab649b964b60518e, so we don't need to revert it.

PR #259, opened 2016-03-27
[WIP] Igloo generation & new structure format

aramperes:
This PR adds the generation of Igloos. This generation process is different from other temple generators, because Mojang introduced structure templates (currently only for Igloos and End Cities). Since we want to avoid asset distribution, I transformed these templates into JSON format instead of the NBT format.

Igloos are split in 3 parts (templates): top, middle and bottom. The middle and bottom part have a 50% chance of generating.

Unfortunately, I added another change to CompoundTag in this same PR, because it was necessary for the template TE/Entity parser. Same with EntityRegistry.

To-Do list:
- Igloos:
  - [x] Top generator
  - [x] Middle generator
  - [x] Bottom generator
  - [ ] Random loot
- Structure templates
  - [x] JSON Reproduction
  - [x] Parser
  - [x] Block loading
  - [x] TileEntity loading
  - Entities
    - [x] Entity loading
    - [x] Entity spawning
    - [x] Entity compound appropriation
- Code
  - [ ] Documentation
  - [ ] JUnit Tests


src/main/java/net/glowstone/generator/structures/GlowTemple.java
```diff
@@ -21,6 +21,8 @@ public GlowTemple(World world, int chunkX, int chunkZ) {
         types.put(Biome.JUNGLE, TempleType.JUNGLE_TEMPLE);
         types.put(Biome.JUNGLE_HILLS, TempleType.JUNGLE_TEMPLE);
         types.put(Biome.SWAMPLAND, TempleType.WITCH_HUT);
+        types.put(Biome.FOREST, TempleType.IGLOO);        //TODO: Temporary, change to ICE_FLAT
+        types.put(Biome.FOREST_HILLS, TempleType.IGLOO);  //TODO: Temporary, change to TAIGA_COLD
```
mastercoms, line 25: when can this be changed?

mastercoms:
Could the JSON be more readable in formatting and naming?

src/main/java/net/glowstone/generator/structures/GlowTemple.java
aramperes, line 25: I put it like that because I couldn't find an Ice Plains biome, since there were more forest biomes... I'll change it back when I'm done testing the generation of the structure.

aramperes:
@mastercoms 
I can format it properly, yes, but the size of the files gets considerably larger (2.5 times the current size).
For the naming, it's necessary to leave a trace of the original Tag Type for the NBT structure to be reproduced internally for TE/Entities.

mastercoms:
Could there be a `.json` and `.min.json`?

aramperes:
@mastercoms 
Sure, but the `.json` would still be included inside the JAR? I can understand putting it on here, but it seems like a waste of space to pack it too imo

mastercoms:
We can exclude in the build process.

mastercoms:
Could the structure JSON data be committed as another, dummy user? I don't really know how this stuff works legally, but I think it would be safer.

jimmikaelkael:
For random loot, there's class that may be reusable:
net.glowstone.generator.objects.RandomItemsContent.java

Not sure it fits your needs, but that's what I used to fill dungeon/temple chests

aramperes:
We will be implementing a more modular structure system - new PR soon

PR #261, opened 2016-03-28
Fix Some Placement

ghost:
There are quite some incorrect placement checks. This pull request intends to fix some of them; specifically, for carpets, dead bushes, nether warts, saplings and tall grass.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockCarpet.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockCarpet.java
@@ -1,11 +1,12 @@
 package net.glowstone.block.blocktype;
 
 import net.glowstone.block.GlowBlock;
+import org.bukkit.Material;
 import org.bukkit.block.BlockFace;
 
 public class BlockCarpet extends BlockNeedsAttached {
     @Override
     public boolean canPlaceAt(GlowBlock block, BlockFace against) {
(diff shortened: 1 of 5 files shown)


src/main/java/net/glowstone/block/blocktype/BlockCarpet.java
```diff
@@ -6,6 +6,6 @@
 public class BlockCarpet extends BlockNeedsAttached {
     @Override
     public boolean canPlaceAt(GlowBlock block, BlockFace against) {
-        return super.canPlaceAt(block, against) && !block.getRelative(BlockFace.DOWN).isEmpty();
+        return !block.getRelative(BlockFace.DOWN).isEmpty();
```
kamcio96:
```
world.getTypeId(x, y, z) != 0 
```

I know it is annoying but creating new objects is unnecessary

ghost: Might as well do it for all type checks and block data checks then?

kamcio96: Yes, it should be changed everywhere, especially in BlockLiquid, [eg](https://github.com/GlowstoneMC/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/block/blocktype/BlockLiquid.java#L133)

Issue #239, opened 2016-03-14
Unknown biome id from vanilla generated world


mastercoms:
```
[SEVERE] Error while generating chunk (15,-14)
java.lang.ArrayIndexOutOfBoundsException: 1002
        at net.glowstone.constants.GlowBiome.getBiome(GlowBiome.java:106)
        at net.glowstone.generator.OverworldGenerator.generateTerrainDensity(OverworldGenerator.java:269)
        at net.glowstone.generator.OverworldGenerator.generateRawTerrain(OverworldGenerator.java:172)
        at net.glowstone.generator.OverworldGenerator.generateChunkData(OverworldGenerator.java:121)
        at net.glowstone.ChunkManager.generateChunk(ChunkManager.java:265)
        at net.glowstone.ChunkManager.loadChunk(ChunkManager.java:163)
        at net.glowstone.GlowChunk.load(GlowChunk.java:187)
        at net.glowstone.generator.populators.StructurePopulator.populate(StructurePopulator.java:36)
        at net.glowstone.ChunkManager.populateChunk(ChunkManager.java:230)
        at net.glowstone.ChunkManager.forcePopulation(ChunkManager.java:245)
        at net.glowstone.entity.GlowPlayer.streamBlocks(GlowPlayer.java:685)
        at net.glowstone.entity.GlowPlayer.pulse(GlowPlayer.java:550)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
        at java.util.LinkedList$LLSpliterator.forEachRemaining(LinkedList.java:1235)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:418)
        at net.glowstone.GlowWorld.pulsePlayers(GlowWorld.java:492)
        at net.glowstone.GlowWorld.pulse(GlowWorld.java:373)
[...]

PR #263, opened 2016-04-02
Overworld generation fix

jimmikaelkael:
Fixes incomplete 1.9 migration (#239).

Next time please ask me instead of modifying the biomegrid, this is a sensible mechanic.


mastercoms:
Hm, sorry. I tried to revert everything I removed after I figured out the problem, but I missed these. Thanks for the fix.

PR #265, opened 2016-04-14
[WIP] Structure template API

aramperes:
This PR will add a completely new Structure API to Glowstone++. The aim of these changes is for custom structures and customizable generation by server owners and plugin developers to be possible. This will also mean that all current structures (Temples, Igloos, End City) will be converted into a JSON-based file structure, 'templates'.

This PR consists of:
- Converting hard-coded structures into loadable JSON file templates
- Runtime generation manipulation by plugins
- Geometric and Schematic structure support
  - Geometry can be used to create simpler structures with shapes
  - Schematics can be used by server owners (because it is a universal format)

Current development progress:
- [x] API for Structure templates
  - [x] Geometric structure builder
  - [ ] Schematic structure builder
- [ ] JSON file loader (templates)
- [ ] Add templates to world generation
- [ ] Document
- [ ] Testing

Sample Igloo template (WIP): https://gist.github.com/momothereal/afdc9f533e0559dc447b754b6c793a53

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,10 @@ There are a few basic ground-rules for contributors:
 
 Declaring formal releases remains the prerogative of the project maintainer.
 
+## How?
+
+For more information on how to start contributing to Glowstone, please view our [pull request how-to](https://github.com/GlowstoneMC/GlowstonePlusPlus/wiki/PR-Handling).
+
 ## Changes to this arrangement
 
 This is an experiment and feedback is welcome! This document may also be subject to pull-requests or changes by contributors where you believe you have something valuable to add or change.


Paulomart:
Will this also allow loading in mojang format?

aramperes:
No, we decided to move away from Mojang stuff and unite structures into one format. Igloo and End City will be re-made using Geometry.

Paulomart:
What about allowing to load mojang format? For faster updating, etc.

aramperes:
We dont want to distribute the assets, though.

> mastercoms: You don't have to hard code them, but don't import the file from 1.9. I think that we should look at the information in them, and determine the best way to organize that information ourselves. If we want to use JSON, we shouldn't be changing formats just because we're copying code. We should make our own original structure for the JSON that can be better than the Mojang structure, and use the file format best suited for our purposes, not because we want to avoid being similar to Mojang.

Paulomart:
ok then

mastercoms:
What happened? :(

aramperes:
@mastercoms Long story short, I tried squashing two commits and accidentally merged two branches together, so I'll need to re-commit the changes to this PR.

PR #268, opened 2016-04-20
Add Elytra flying/gliding.

aramperes:
This PR adds the functionality of the Elytra, including gliding and flying.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -63,6 +63,7 @@ import org.bukkit.entity.Villager;
 import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
 import org.bukkit.event.entity.EntityRegainHealthEvent;
 import org.bukkit.event.entity.EntityRegainHealthEvent.RegainReason;
+import org.bukkit.event.entity.EntityToggleGlideEvent;
 import org.bukkit.event.entity.FoodLevelChangeEvent;
 import org.bukkit.event.player.*;
 import org.bukkit.event.player.PlayerTeleportEvent.TeleportCause;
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1175,12 +1176,16 @@ public double getEyeHeight(boolean ignoreSneaking) {
...
     }
 
     @Override
-    public void setGliding(boolean b) {
+    public void setGliding(boolean gliding) {
+        if (EventFactory.callEvent(new EntityToggleGlideEvent(this, gliding)).isCancelled()) {
+            return;
+        }
 
+        metadata.setBit(MetadataIndex.STATUS, StatusFlags.GLIDING, gliding);
```
mastercoms, line 1188: Use `event.isGliding()` instead of `gliding`. I know that isGliding is final in the event, but I plan to change that.

src/main/java/net/glowstone/net/handler/play/player/PlayerUpdateHandler.java
```diff
@@ -83,6 +85,13 @@ confusion about where its actual location is (e.g. during login)
             player.setOnGround(message.isOnGround());
         }
 
+        // Checks if the player is still wearing the Elytra
+        ItemStack chestplate = player.getInventory().getChestplate();
+        boolean hasElytra = chestplate != null && chestplate.getType() == Material.ELYTRA && chestplate.getDurability() > 1;
+        if (player.isGliding() && (player.isOnGround() || !hasElytra)) {
```
mastercoms, line 91: Does this require an `isInWater()` check?

mastercoms:
Is it possible to add this to living entity and then add a special case for clientside handling for player? So plugins can add elytra to other living entities besides players. It's up to you to decide if that's in the scope of this PR.

aramperes:
@mastercoms Might be out of scope; other mobs can glide too, although mob gliding is done server-side (meaning this would have to re-implement elytra phyics too), while player gliding is just a 'Start Gliding' action and controls flight physics by the client.

mastercoms:
We can leave the physics as a TODO :)

PR #273, opened 2016-05-15
Update to 1.9.4, compatibility with 1.9, 1.9.2, 1.9.3

aramperes:
Adds compatibility with 1.9.X clients (1.9-1.9.4), as well as the new `/stopsound` command for 1.9.4 clients.

Changes made:
- Added new protocol: Compatible_Play
  - 1.9 protocol is `PLAY_107`, can be turned off in glowstone.yml
  - 1.9.2-3 protocol is `PLAY_109`
  - 1.9.4 protocol is `PLAY`
  - Protocol changes [here](http://wiki.vg/Protocol_History#1.9.3-pre2)
- `/stopsound` command, only works on 1.9.4 clients
- `allow-legacy-clients` glowstone.yml entry to enable/disable 1.9 clients to connect

Glowkit PR for version change: [18](https://github.com/GlowstonePlusPlus/Glowkit/pull/18)

### DIFF
--- a/src/main/java/net/glowstone/net/GlowSession.java
+++ b/src/main/java/net/glowstone/net/GlowSession.java
@@ -375,12 +375,14 @@ public final class GlowSession extends BasicSession {
     }
 
     public ProtocolType getProtocolType() {
-        if (version == GlowServer.PROTOCOL_VERSION)
-            return ProtocolType.PLAY;
-        if (version == GlowServer.LEGACY_PROTOCOL_VERSION)
-            return ProtocolType.PLAY_107;
-        if (version == GlowServer.COMPATIBLE_PROTOCOL_VERSION)
-            return ProtocolType.PLAY_109;
+        switch (version) {
+            case GlowServer.PROTOCOL_VERSION:
+                return ProtocolType.PLAY;
+            case GlowServer.LEGACY_PROTOCOL_VERSION:
+                return ProtocolType.PLAY_107;
+            case GlowServer.COMPATIBLE_PROTOCOL_VERSION:
+                return ProtocolType.PLAY_109;
+        }
         return null;
     }
 
--- a/src/main/java/net/glowstone/net/GlowSession.java
+++ b/src/main/java/net/glowstone/net/GlowSession.java
@@ -378,9 +378,9 @@ public final class GlowSession extends BasicSession {
         switch (version) {
             case GlowServer.PROTOCOL_VERSION:
                 return ProtocolType.PLAY;
-            case GlowServer.LEGACY_PROTOCOL_VERSION:
+            case GlowServer.LEGACY_PROTOCOL_1_9:
                 return ProtocolType.PLAY_107;
-            case GlowServer.COMPATIBLE_PROTOCOL_VERSION:
+            case GlowServer.LEGACY_PROTOCOL_1_9_2:
                 return ProtocolType.PLAY_109;
         }
         return null;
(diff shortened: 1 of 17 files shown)


kamcio96:
What about PLAY, PLAY_107, PLAY_109?

Paulomart:
Yeah I don't like the names currently too. Including protocol number seems like a reasonable way.

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -374,6 +374,16 @@ public void setPlayer(PlayerProfile profile) {
         send(new UserListItemMessage(Action.ADD_PLAYER, entries));
     }
 
+    public ProtocolType getProtocolType() {
+        if (version == GlowServer.PROTOCOL_VERSION)
```
mastercoms: can this be a switch statement?

src/main/java/net/glowstone/net/GlowSession.java
```diff
@@ -375,12 +375,14 @@ public void setPlayer(PlayerProfile profile) {
     }
 
     public ProtocolType getProtocolType() {
-        if (version == GlowServer.PROTOCOL_VERSION)
-            return ProtocolType.PLAY;
-        if (version == GlowServer.LEGACY_PROTOCOL_VERSION)
-            return ProtocolType.PLAY_107;
-        if (version == GlowServer.COMPATIBLE_PROTOCOL_VERSION)
-            return ProtocolType.PLAY_109;
+        switch (version) {
```
Paulomart, line 378: Why not have some kind of map here to help this out?

mastercoms, line 378: I feel like that would be a waste of memory for just some convenience.

Paulomart, line 378: LEGACY and COMPATIBLE are in my view not the perfect names for the use case. Would go something with more like Bungee does. https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/ProtocolConstants.java

aramperes, line 378: @Paulomart So would you prefer `LEGACY_PROTOCOL_1_9` and `LEGACY_PROTOCOL_1_9_2` as a replacement for `LEGACY_PROTOCOL_VERSION` and `COMPATIBLE_PROTOCOL_VERSION`?

Paulomart, line 378: Yes! It clearly states which MC versions protocol number that is.

Paulomart:
Also what about 1.10? Should we do a full update or a protocol hack? Or shall we wait ?

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
```diff
@@ -42,7 +42,8 @@ public void handle(GlowSession session, StatusRequestMessage message) {
         JSONObject version = new JSONObject();
         version.put("name", "Glowstone++ " + GlowServer.GAME_VERSION);
         int protocolVersion = session.getVersion();
-        if (protocolVersion > 109 || protocolVersion < 107) protocolVersion = GlowServer.PROTOCOL_VERSION;
+        boolean compatible = protocolVersion == GlowServer.PROTOCOL_VERSION || protocolVersion == GlowServer.COMPATIBLE_PROTOCOL_VERSION || (protocolVersion == GlowServer.LEGACY_PROTOCOL_VERSION && server.canSupportLegacyClients());
```
Paulomart: This line is crowded, can we make this more readable

aramperes:
@Paulomart I've been working on a 1.10 branch for a while now (https://github.com/momothereal/GlowstonePlusPlus/tree/1.10-dev). We discussed on Gitter that we should probably focus on finishing 1.8-1.9 features before updating G++ to 1.10.

mastercoms:
Well protocol should be updated.

Issue #275, opened 2016-05-30
Achievements are not being saved


ghost:
When you get a achievement, it's not being saved. 

```
30/05/2016 - 18:50:45 [INFORMATION] Shad0wCore joined the game
30/05/2016 - 18:51:49 [INFORMATION] Shad0wCore has just earned the achievement [Taking Inventory]
30/05/2016 - 18:51:53 [INFORMATION] Shad0wCore issued command: /save-all
30/05/2016 - 18:51:53 [INFORMATION] Shad0wCore: Forcing save..
30/05/2016 - 18:51:56 [INFORMATION] Shad0wCore: Save complete.
30/05/2016 - 18:51:57 [INFORMATION] Shad0wCore [/127.0.0.1:64444] lost connection
30/05/2016 - 18:51:57 [INFORMATION] Shad0wCore left the game
30/05/2016 - 18:52:04 [INFORMATION] Scanning plugins...
30/05/2016 - 18:52:04 [INFORMATION] PluginTypeDetector: found 1 Bukkit, 0 Sponge, 0 Forge, 0 Canary, 0 unknown plugins (total 1)
30/05/2016 - 18:52:04 [INFORMATION] Set PluginClassLoader as parallel capable
30/05/2016 - 18:52:04 [INFORMATION] Preparing spawn for world...
30/05/2016 - 18:52:05 [INFORMATION] Preparing spawn for world: done
30/05/2016 - 18:52:05 [INFORMATION] Preparing spawn for world_nether...
30/05/2016 - 18:52:06 [INFORMATION] Preparing spawn for world_nether: done
30/05/2016 - 18:52:06 [INFORMATION] Preparing spawn for world_the_end...
30/05/2016 - 18:52:06 [INFORMATION] Preparing spawn for world_the_end: done
30/05/2016 - 18:52:06 [INFORMATION] [Essentials] Enabling Essentials vTeamCity
30/05/2016 - 18:52:06 [INFORMATION] Bukkit version format changed. Version not checked.
30/05/2016 - 18:52:06 [INFORMATION] 1.9.2-SNAPSHOT-6bbf762-MC1.9.2
30/05/2016 - 18:52:06 [INFORMATION] 1.9.2-R0.1-SNAPSHOT
30/05/2016 - 18:52:06 [INFORMATION] [Essentials] Starting Essentials ban format conversion
30/05/2016 - 18:52:06 [INFORMATION] [Essentials] Found 1 files to convert...
[...]

PR #277, opened 2016-05-30
Save player achievements and stats in stats/ directory.

aramperes:
Resolves issue #275.

StatisticMaps (from GlowPlayer) will now be saved when needed to the appropriate stats/ directory, as well as read when the player is loaded.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -22,6 +22,7 @@ import net.glowstone.inventory.GlowInventory;
 import net.glowstone.inventory.GlowItemFactory;
 import net.glowstone.inventory.crafting.CraftingManager;
 import net.glowstone.io.PlayerDataService;
+import net.glowstone.io.PlayerStatisticIoService;
 import net.glowstone.io.ScoreboardIoService;
 import net.glowstone.map.GlowMapView;
 import net.glowstone.net.GlowNetworkServer;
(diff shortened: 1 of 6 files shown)


src/main/java/net/glowstone/GlowServer.java
```diff
@@ -972,6 +973,15 @@ public ScoreboardIoService getScoreboardIoService() {
     }
 
     /**
+     * Returns the player statitics I/O service attached to the first world.
+     *
+     * @return the server's statistics I/O service
+     */
+    public PlayerStatisticIoService getPlayerStatisticIoService() {
+        return worlds.getWorlds().get(0).getStorage().getPlayerStatisticIoService();
```
mastercoms: This won't be compatible with multiworld plugins.

src/main/java/net/glowstone/io/PlayerStatisticIoService.java
```diff
@@ -0,0 +1,80 @@
...
+    public void readStats(GlowPlayer player) {
+        File statsFile = getPlayerFile(player.getUniqueId());
+        player.getStatisticMap().getValues().clear();
+        if (statsFile.exists()) {
+            try {
+                JSONParser parser = new JSONParser();
+                JSONObject json = (JSONObject) parser.parse(new FileReader(statsFile));
+                for (Object o : json.keySet()) {
+                    String key = (String) o;
+                    int value = (int) (long) json.get(o);
```
mastercoms: Why are there two casts here?

aramperes: @mastercoms eh, iirc JSONSimple stores numbers as long, but I don't know if you can skip that cast right to int. Give me a sec.

Edit: Yep, can confirm, double-cast is necessary: `java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer`

mastercoms: Oh I did not realize it was `Long`. Try using `intValue()` instead of casting. https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#intValue--

aramperes: @mastercoms Well I mean, not like if it changes that much :P

``` java
//Long.class
public int intValue() {
    return (int)this.value;
}
```

mastercoms: I feel like using a function is better because what if something in Java changes in the future? And it can be identified by the JVM if there are any optimizations available for this.

Issue #284, opened 2016-06-05
GlowWorld.requestPulse is used for non-repeating update functions


GreenBeard:
GlowWorld.requestPulse (which is repeating) is being used in instances where it could be instead a single pulse after a delay.

GreenBeard:
Should be fixed in #285 so as soon as the is pull is accepted then this bug can be closed.

PR #285, opened 2016-06-05
[WIP] Added new requestSinglePulse in GlowWorld

GreenBeard:
Added new requestSinglePulse in GlowWorld.java and changed liquid blocks (water, and lava) to use this method instead of the forever looping requestPulse method.

This fixes the issue with #284 and stops the memory from forever spiking after new chunks are spawned.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -137,7 +137,6 @@ public final class GlowWorld implements World {
     /**
      * Contains how regular blocks should be pulsed.
      */
-    private final ConcurrentHashMap tickMap = new ConcurrentHashMap<>(0);
     private final ConcurrentHashMap singleTickMap = new ConcurrentHashMap<>(0);
     private final Spigot spigot = new Spigot() {
         @Override
@@ -354,7 +353,6 @@ public final class GlowWorld implements World {
         activeChunksSet.clear();
 
         // We should pulse our tickmap, so blocks get updated.
-        pulseTickMap();
         pulseSingleTickMap();
 
         // pulse players last so they actually see that other entities have
@@ -1787,16 +1785,6 @@ public final class GlowWorld implements World {
         return result;
     }
 
-    private void pulseTickMap() {
-        ItemTable itemTable = ItemTable.instance();
-        getTickMap().entrySet().stream().filter(entry -> worldAge % entry.getValue() == 0).forEach(entry -> {
-            GlowBlock block = getBlockAt(entry.getKey());
-            BlockType notifyType = itemTable.getBlock(block.getTypeId());
-            if (notifyType != null)
-                notifyType.receivePulse(block);
-        });
-    }
-
     private void pulseSingleTickMap() {
         ItemTable itemTable = ItemTable.instance();
         getSingleTickMap().entrySet().stream().filter(entry -> entry.getValue() - worldAge <= 0).forEach(entry -> {
@@ -1808,28 +1796,8 @@ public final class GlowWorld implements World {
         });
     }
 
-    private ConcurrentHashMap<Location, Long> getTickMap() {
-        return tickMap;
-    }
-
     private ConcurrentHashMap<Location, Long> getSingleTickMap() {
         return singleTickMap;
-    }
-
-    /**
-     * Calling this method will request that the block is ticked on the next iteration
-     * that applies to the specified tick rate.
-     *
-     * @param block The block to tick.
-     * @param tickRate The tick rate to tick the block at.
-     */
-    public void requestPulse(GlowBlock block, long tickRate) {
-        Location target = block.getLocation();
-
-        if (tickRate > 0)
-            tickMap.put(target, tickRate);
-        else if (tickMap.containsKey(target))
-            tickMap.remove(target);
     }
 
     /**
@@ -1837,7 +1805,7 @@ public final class GlowWorld implements World {
      * tick delay
      *
      * @param block The block to tick.
-     * @param tickRate The tick delay after which to pulse the block.
+     * @param tickDelay The tick delay after which to pulse the block.
      */
     public void requestSinglePulse(GlowBlock block, long tickDelay) {
         Location target = block.getLocation();
@@ -1845,10 +1813,6 @@ public final class GlowWorld implements World {
         if (tickDelay < 0)
             tickDelay = 0;
         singleTickMap.put(target, worldAge + tickDelay);
-    }
-
-    public void cancelPulse(GlowBlock block) {
-        requestPulse(block, 0);
     }
 
     @Override
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -138,6 +138,7 @@ public final class GlowWorld implements World {
      * Contains how regular blocks should be pulsed.
      */
     private final ConcurrentHashMap tickMap = new ConcurrentHashMap<>(0);
+    private final ConcurrentHashMap singleTickMap = new ConcurrentHashMap<>(0);
     private final Spigot spigot = new Spigot() {
         @Override
         public void playEffect(Location location, Effect effect) {
@@ -354,6 +355,7 @@ public final class GlowWorld implements World {
 
         // We should pulse our tickmap, so blocks get updated.
         pulseTickMap();
+        pulseSingleTickMap();
 
         // pulse players last so they actually see that other entities have
         // moved. unfortunately pretty hacky. not a problem for players b/c
@@ -1795,8 +1797,23 @@ public final class GlowWorld implements World {
         });
     }
 
+    private void pulseSingleTickMap() {
+        ItemTable itemTable = ItemTable.instance();
+        getSingleTickMap().entrySet().stream().filter(entry -> entry.getValue() - worldAge <= 0).forEach(entry -> {
+            GlowBlock block = getBlockAt(entry.getKey());
+            BlockType notifyType = itemTable.getBlock(block.getTypeId());
+            if (notifyType != null)
+                notifyType.receivePulse(block);
+            getSingleTickMap().remove(entry.getKey(), entry.getValue());
+        });
+    }
+
     private ConcurrentHashMap<Location, Long> getTickMap() {
         return tickMap;
+    }
+
+    private ConcurrentHashMap<Location, Long> getSingleTickMap() {
+        return singleTickMap;
     }
 
     /**
@@ -1809,11 +1826,25 @@ public final class GlowWorld implements World {
     public void requestPulse(GlowBlock block, long tickRate) {
         Location target = block.getLocation();
 
-
         if (tickRate > 0)
             tickMap.put(target, tickRate);
         else if (tickMap.containsKey(target))
             tickMap.remove(target);
+    }
+
+    /**
+     * Calling this method will request that the block is ticked after the specified
+     * tick delay
+     *
+     * @param block The block to tick.
+     * @param tickRate The tick delay after which to pulse the block.
+     */
+    public void requestSinglePulse(GlowBlock block, long tickDelay) {
+        Location target = block.getLocation();
+
+        if (tickDelay < 0)
+            tickDelay = 0;
+        singleTickMap.put(target, worldAge + tickDelay);
     }
 
     public void cancelPulse(GlowBlock block) {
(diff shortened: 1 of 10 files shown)


GreenBeard:
Well I realized that I accidently left in the isLiquid function which I never used. Although it could be helpful. Should I leave it or just get rid of it?

Paulomart:
The isLiquid method has nothing to do with the pr so it would be the best if it got removed.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1795,10 +1797,30 @@ private void pulseTickMap() {
...
+        HashMap<Location, Long> pulsesToRemove = new HashMap<Location, Long>();
+        getSingleTickMap().entrySet().stream().filter(entry -> entry.getValue() - worldAge <= 0).forEach(entry -> {
+            GlowBlock block = getBlockAt(entry.getKey());
+            BlockType notifyType = itemTable.getBlock(block.getTypeId());
+            if (notifyType != null)
+                notifyType.receivePulse(block);
+            pulsesToRemove.put(entry.getKey(), entry.getValue());
+        });
+
+        for (Map.Entry<Location, Long> entry : pulsesToRemove.entrySet()) {
```
mastercoms: Why not remove the entries in the first stream?

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -92,15 +104,15 @@ public void placeBlock(GlowPlayer player, GlowBlockState state, BlockFace face,
         // 0 = Full liquid block
         state.setType(getMaterial());
         state.setRawData((byte) 0);
-        state.getBlock().getWorld().requestPulse(state.getBlock(), isWater(state.getBlock().getType()) || state.getBlock().getBiome() == Biome.HELL ? TICK_RATE_WATER : TICK_RATE_LAVA);
+        state.getBlock().getWorld().requestSinglePulse(state.getBlock(), isWater(state.getBlock().getType()) || state.getBlock().getBiome() == Biome.HELL ? TICK_RATE_WATER : TICK_RATE_LAVA);
```
mastercoms: Make the tick delay 0.

mastercoms:
I appreciate this PR because it is a good CPU optimization, but this does not have an effect on memory.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1809,13 +1826,25 @@ private void pulseTickMap() {
...
+     * Calling this method will request that the block is ticked after the specified
+     * tix delay
+     *
+     * @param block The block to tick.
+     * @param tickRate The tick delay after which to pulse the block.
+     */
+    public void requestSinglePulse(GlowBlock block, long tickDelay) {
+        Location target = block.getLocation();
+
+        singleTickMap.put(target, worldAge + tickDelay);
```
mastercoms, line 1815: Do we want to allow negative tickDelays?

src/main/java/net/glowstone/generator/decorators/nether/LavaDecorator.java
```diff
@@ -53,7 +53,7 @@ public void decorate(World world, Random random, Chunk source) {
                 state.setType(Material.LAVA);
                 state.update(true);
                 // 10 instead of the default 5 because lava should flow faster in the nether
-                ((GlowWorld) block.getWorld()).requestPulse((GlowBlock) block, 10);
+                ((GlowWorld) block.getWorld()).requestSinglePulse((GlowBlock) block, 10);
```
mastercoms: Set delay to 0.

src/main/java/net/glowstone/generator/decorators/overworld/FlowingLiquidDecorator.java
```diff
@@ -51,7 +51,7 @@ public void decorate(World world, Random random, Chunk source) {
                     BlockState state = block.getState();
                     state.setType(type);
                     state.update(true);
-                    ((GlowWorld) block.getWorld()).requestPulse((GlowBlock) block, 20);
+                    ((GlowWorld) block.getWorld()).requestSinglePulse((GlowBlock) block, 20);
```
mastercoms: Set delay to 0.

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -246,6 +247,9 @@ public void updatePhysics(GlowBlock block) {
         if (!(Byte.compare(block.getState().getRawData(), isWater(block.getType()) || block.getBiome() == Biome.HELL ? STRENGTH_MIN_WATER : STRENGTH_MIN_LAVA) == 0) || block.getRelative(DOWN).getType() == Material.AIR) {
             calculateFlow(block);
         }
+        if (initalData != block.getState().getRawData()) {
```
mastercoms: I can't quite understand what this is doing.

mastercoms: I can't quite understand what this is doing.

GreenBeard: That is supposed to check if there is a change to a liquid block and if so then it will tick again (after waiting the proper amount of time). Does it look like I didn't implement it correctly? If the data didn't change then the block clearly doesn't require ticking again until a block changes around it.

src/main/java/net/glowstone/GlowWorld.java
GreenBeard, line 1815: I don't see why we shouldn't. I honestly think negative tick doing delays would be a bad idea to do but I'm not sure there would be harm. Should I throw an exception if it's negative or just treat is as 1? It is sort of already being treated as 1 because it will just tick the next available time effectively being 1.

(If you think I shouldn't allow negative tickDelays should I allow 0 tick delays?)

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
mastercoms: Could you give an example of what change it is looking for?

src/main/java/net/glowstone/GlowWorld.java
mastercoms, line 1815: Treat <= 0 as 0.

GreenBeard, line 1815: Should I change tickDelay to 0 if it's less than 0? I think if you want anything less than zero to be changed treated as zero it wouldn't really have an effect because it will just be run next tick and it is stored in the hashmap as `worldAge + tickDelay`. If you want I will change it though.

mastercoms, line 1815: It might cause some bugs that we're not aware of, or if we change something about ticks in the future, having a negative tick delay might cause something weird to happen.

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
GreenBeard: It was meant to stop times where a blocks physics wasn't being run again when blocks where changing around it so it would be a leftover water block (that shouldn't be there) but after some testing I have realized that the check was unnecessary as the physics was already updating when blocks around it changed.

GreenBeard:
@mastercoms Does everything look good to you now?

mastercoms:
A cool thing was brought up in gitter. Calling super.receivePulse() after you requestPulse() makes it run once, so we can use that instead since it is wasteful to have two collections and two loops.

aramperes:
@mastercoms for reference: https://github.com/GlowstoneMC/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/block/blocktype/BlockType.java#L220

GreenBeard:
@mastercoms I don't understand how having 2 different lists is wasteful if they aren't storing the same values. They would be just as big as one list together (although if concurrent hashmaps have some overhead that I am not aware of then I'm wrong). Also it isn't really two loops. Yes both lists are looped through but each item would need to be looped through anyway and they don't have separate threads or anything. 

Also if I'm wrong about what I said above and it improves the efficiency of the code then the way that requestPulse works wouldn't be the same as the requestSinglePulse because it uses % which is getting the remainder and checking if it is 0 so it is less accurate on the first tick's delay. The first tick is much more critical if you are only doing one tick. Though if there is still benefit to changing I would be happy to do so.

mastercoms:
I can check the effects if you change it. If they're not what we want, then we can revert the changes. 

All collections have overhead.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1795,10 +1797,25 @@ private void pulseTickMap() {
...
 
+    private void pulseSingleTickMap() {
+        ItemTable itemTable = ItemTable.instance();
+        getSingleTickMap().entrySet().stream().filter(entry -> entry.getValue() - worldAge <= 0).forEach(entry -> {
+            GlowBlock block = getBlockAt(entry.getKey());
+            BlockType notifyType = itemTable.getBlock(block.getTypeId());
+            if (notifyType != null)
+                notifyType.receivePulse(block);
+            getSingleTickMap().remove(entry.getKey(), entry.getValue());
+        });
```
kamcio96, line 1796: What about `singleTickMap.clear()` instead `remove` ?

GreenBeard:
@kamcio96 Not all the tasks are run on the same tick so I don't think I can clear it. (Some have 10 tick delays, some have 1, etc.) I apologize if I am misunderstanding what you are saying.

kamcio96:
Ahh, I thought "pulse in next tick"

GreenBeard:
@mastercoms I did what you suggested. I am however very against your suggestion because it makes the code **harder to read in the future** and it also hardly saves any memory at all like (1kb or 2?). Finally I can't do the function I should which is have the blocks tick after 0 (when placed and added to the world) because then it cancels the pulse so I had to change it to 1.

@gdude2002 @kamcio96 I would like to know if you agree with what @mastercoms said. (Please also read why I am against it in my [previous post](https://github.com/GlowstoneMC/GlowstonePlusPlus/pull/285#issuecomment-225452455)).

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -138,6 +138,7 @@
      * Contains how regular blocks should be pulsed.
      */
     private final ConcurrentHashMap tickMap = new ConcurrentHashMap<>(0);
+    private final ConcurrentHashMap singleTickMap = new ConcurrentHashMap<>(0);
```
mastercoms, line 140: Forgot to remove this

mastercoms:
There is an if statement in requestPulse that checks if the tick rate is above 0.  It is not removed from the tick, but it is never added.

You also forgot to cancel the tick.

I would recommend adding an else statement in the if tree in requestPulse to add the block to the tick with a tick rate of 1, and then cancel pulse. There will be no extra delay if you use 1 because the tick map will never tick instantaneously, and will always at minimum instead do it on the next tick.

The code is a lot simpler because of this change, but I understand your concern about people becoming confused so add a JavaDoc describing this behavior with using 0 as a tick rate to prevent any confusion.

GreenBeard:
@mastercoms
To respond to what you said about it not being removed if it is less than 0 it clearly is. And also sorry I forgot to remove the map I'll fix that.

```
        if (tickRate > 0)
            tickMap.put(target, tickRate);
        else if (tickMap.containsKey(target))
            tickMap.remove(target);
```

Also I did cancel the tick for the BlockLiquid.

```
        updatePhysics(block);
        // If you set the pulse rate to 0 it removes the pulse from the list
        block.getWorld().requestPulse(block, 0);
```

And the code is more complicated with this change not "simpler". Why are you so against 2 lists? I honestly don't understand.

mastercoms:
It is only removed if it is in the tick map to begin with. But since it is added with a tick rate of 0, then it is just ignored. I was responding to this part of your comment:

> Finally I can't do the function I should which is have the blocks tick after 0 (when placed and added to the world) because then it cancels the pulse so I had to change it to 1.

Where is the added complexity in my proposal? It adds new functionality to requestPulse for single tick operations, when you put 0 for the tickrate and the block isn't in the tick map already.

GreenBeard:
@mastercoms I think everything should be good now. (Except for maybe fire? I tested it out and it seemed to work but it didn't quite seem to burn fast enough. I'm not sure and would love feedback)

mastercoms:
I merged your PR, with a few changes to fix bugs and performance issues and to fit my suggestion. It seemed like I wasn't clear enough in my explanation so I just did it myself.

PR #303, opened 2016-06-22
Add Kotlin Maven dependencies (Kotlin support)

aramperes:
Adds support for Kotlin in Glowstone. ([More info about Kotlin](https://en.wikipedia.org/wiki/Kotlin_%28programming_language%29))

Since Kotlin is a JVM-based language, it doesn't require any special loading. However, some basic functions like inheritance and reflection-related-stuff require the Kotlin runtime/reflection libraries.

For anyone curious, this is a sample Kotlin Bukkit plugin:

``` kotlin
import org.bukkit.plugin.java.JavaPlugin

class MyPlugin : JavaPlugin() {
    override fun onEnable() {
        logger.info("This is a Kotlin plugin... wew lad")
        getCommand("mycommand").executor = MyCommand()
        server.pluginManager.registerEvents(MyListener(), this)
    }
}
```

``` kotlin
import org.bukkit.command.Command
import org.bukkit.command.CommandExecutor
import org.bukkit.command.CommandSender

class MyCommand : CommandExecutor {
    override fun onCommand(sender: CommandSender, command: Command, label: String, args: Array<out String>): Boolean {
        var name = sender.name
        sender.sendMessage("Your name is " + name)
        return false
    }
}
```

``` kotlin
import org.bukkit.event.EventHandler
import org.bukkit.event.Listener
import org.bukkit.event.player.PlayerJoinEvent

class MyListener : Listener {
    @EventHandler
    fun onJoin(event: PlayerJoinEvent) {
        event.player.sendMessage("o shit waddup")
    }
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,6 +77,16 @@
             <version>1.14.8</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-runtime</artifactId>
+            <version>1.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
+            <version>1.1-SNAPSHOT</version>
+        </dependency>
         <!-- Automatically downloaded by Glowstone++'s library manager
         <dependency>
             <groupId>org.xerial</groupId>


gdude2002:
wew lad.

gdude2002:
I'm cool with this, but I want @mastercoms to sign off on it.

aramperes:
Quick note: SpigotAPI/Glowkit does not appear to work individually with Kotlin, because the compiler cannot resolve the BungeeCord API (e.g. BaseComponent) directly from them. Instead, use the Glowstone server jar as library.

mastercoms:
kotlin gives me a funny feeling every time I look at it. but why not?

ghost:
I think you missed kotlin-stdlib and kotlin maven plugin

aramperes:
@ensirius kotlin-stdlib is automatically resolved because it is a dependency for kotlin-runtime. kotlin-maven is for compilation, and since we probably won't use Kotlin in Glowstone directly, it is not needed.

phase:
Are we going to get support for any other JVM languages? Scala is used a lot in Minecraft modding.

aramperes:
@phase I wouldn't see why not. I just tested a Scala plugin and it loads/enables fine without adding any library!

``` scala
class ScalaPlugin extends JavaPlugin {
  override def onEnable(): Unit = {
    getLogger.info("This is a scala plugin")
  }
}
```

I'll try some other languages as well. If you have suggestions, please let me know :)

e: Tested Groovy, would work fine with the Groovy maven dependency shaded in: https://mvnrepository.com/artifact/org.codehaus.groovy/groovy-all

Issue #186, opened 2016-01-31
Legacy ping support (fe01 / fe packet)


deathcap:
Glowstone++ supports supports the 'modern' status handshake ping (Netty-based protocol), but not the "legacy" (0xfe packet) ping. According to http://wiki.vg/Protocol#Legacy_Server_List_Ping:

> While not technically part of the current protocol, legacy clients may send this packet to initiate Server List Ping, and modern servers should handle it correctly.

All vanilla Minecraft servers 1.4.4+ support this ping, including 1.8.9 and 1.9 snapshots. Client software such as server status pages may still use it (being much simpler than the modern ping), so I think it would be a good idea for Glowstone++ to support it.

References:
- https://github.com/GlowstonePlusPlus/GlowstonePlusPlus/blob/master/src/main/java/net/glowstone/net/protocol/HandshakeProtocol.java - add here or nearby? but note 0xfe does not have a varint length or packet id field, so special-casing it may be the most feasible approach
- https://github.com/deathcap/node-minecraft-ping - tool for sending various ping types for testing; can also use older vanilla client versions, check if Glowstone++ appears in the server list
- https://github.com/PrismarineJS/node-minecraft-protocol/pull/334 - adding support to another server implementation, could be useful for comparison

Attempts to send a 0xfe01 ping currently will timeout, causing this message in the server log:

```
11:03:41 [INFO] [/127.0.0.1:57037] kicked: Timed out
```
##

mastercoms:
:+1:

deathcap:
Another tool to help test legacy pings: https://github.com/WaterfallMC/LegacyPinger

GreenBeard:
Does anyone know where the length of packets is now read? There are two functions in [GlowProtocol.java](https://github.com/GlowstoneMC/GlowstonePlusPlus/blob/f84d3f08644f2fd43d2828468855bedd87ac9060/src/main/java/net/glowstone/net/protocol/GlowProtocol.java) and I don't understand where the length of the packet is read in newReadHeader but I do understand the old readHeader. The reason I want to know is because the old ping packet didn't have the length prepended to it.

gdude2002:
This ticket needs categorization on the [waffle board](https://waffle.io/GlowstoneMC/Glowstone).

Has this ticket seen any movement or has there been any action expected since the previous comment?

GreenBeard:
@gdude2002 I had this finished for a week or so now just forgot to make the pull request. I'll do that now.

[...]

PR #304, opened 2016-06-23
Added support for legacy pings/queries.

GreenBeard:
This pull request is intended to add support for legacy pings/queries. These come mostly from automated scripts that use the simpler form of pinging compared to the more modern complicated method.

The [WIP] tag is because I was not sure if there was a better method to get the instance of GlowServer in LegacyPingHandler.channelRead than to add another method to GlowNetworkServer. Any comments on these would be appreciated.

~~Also I would like to warn that the normal notchian servers have less checks on legacy pings than this. Specifically in the `case 1` switch block (line 49 of LegacyPingHandler), they don't check if the second byte is 0x01 (they just check if it has a second byte) while this implementation does. If others believe that I should do exactly what the notchian server does I will revise it to do so.~~

This pull request should address #186.

### DIFF
--- a/src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
+++ b/src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
@@ -1,0 +1,100 @@
+package net.glowstone.net.handler.legacyping;
+
+import java.io.UnsupportedEncodingException;
+import java.net.InetSocketAddress;
+
+import org.bukkit.event.server.ServerListPingEvent;
+
+import com.google.common.base.Charsets;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import net.glowstone.EventFactory;
+import net.glowstone.GlowServer;
+import net.glowstone.net.GlowNetworkServer;
+
+public class LegacyPingHandler extends ChannelInboundHandlerAdapter {
+
+    GlowNetworkServer networkServer;
+
+    public LegacyPingHandler(GlowNetworkServer networkServer) {
+        this.networkServer = networkServer;
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext channelHandlerContext, Object object) throws Exception {
+        ByteBuf bytebuf = (ByteBuf) object;
+
+        bytebuf.markReaderIndex();
+        boolean legacyPingProtocol = false;
+
+        try {
+            if (bytebuf.readByte() == (byte) 0xFE) {
+                GlowServer server = networkServer.getServer();
+                int readableBytes = bytebuf.readableBytes();
+
+                InetSocketAddress inetsocketaddress = (InetSocketAddress) channelHandlerContext.channel().remoteAddress();
+
+                ServerListPingEvent legacyPingEvent = new ServerListPingEvent(inetsocketaddress.getAddress(), server.getMotd(), server.getOnlinePlayers().size(), server.getMaxPlayers());
+                EventFactory.callEvent(legacyPingEvent);
+
+                switch (readableBytes) {
+                    case 0:
+                        sendByteBuf(channelHandlerContext, responseToByteBuf(channelHandlerContext, String.format("%s\u00a7%d\u00a7%d", legacyPingEvent.getMotd(), legacyPingEvent.getNumPlayers(), legacyPingEvent.getMaxPlayers())));
+                        legacyPingProtocol = true;
+                        break;
+                    case 1:
+                        if (bytebuf.readByte() == (byte) 0x01) {
+                            sendByteBuf(channelHandlerContext, responseToByteBuf(channelHandlerContext, String.format("\u00a71\u0000%d\u0000%s\u0000%s\u0000%d\u0000%d", GlowServer.PROTOCOL_VERSION, GlowServer.GAME_VERSION, legacyPingEvent.getMotd(), legacyPingEvent.getNumPlayers(), legacyPingEvent.getMaxPlayers())));
+                            legacyPingProtocol = true;
+                        }
+                        break;
+                    default:
+                        if (bytebuf.readByte() == (byte) 0x01 && bytebuf.readByte() == (byte) 0xFA && "MC|PingHost".equals(new String(bytebuf.readBytes(bytebuf.readShort() * 2).array(), Charsets.UTF_16BE))) {
+                            int dataLength = bytebuf.readUnsignedShort();
+                            short clientVersion = bytebuf.readUnsignedByte();
+                            String hostname = bytebuf.readBytes(bytebuf.readShort() * 2).toString(Charsets.UTF_16BE);
+                            @SuppressWarnings("unused")
+                            int port = bytebuf.readInt();
+
+                            if (clientVersion >= 73 && 7 + (hostname.length() * 2) == dataLength && bytebuf.readableBytes() == 0) {
+                                sendByteBuf(channelHandlerContext, responseToByteBuf(channelHandlerContext, String.format("\u00a71\u0000%d\u0000%s\u0000%s\u0000%d\u0000%d", GlowServer.PROTOCOL_VERSION, GlowServer.GAME_VERSION, legacyPingEvent.getMotd(), legacyPingEvent.getNumPlayers(), legacyPingEvent.getMaxPlayers())));
+                                legacyPingProtocol = true;
+                            }
+                        }
+                        break;
+                }
+            }
+        } catch (RuntimeException e) {
+            // Silently catch the exception
+            ;
+        } finally {
+            // check if not successful, otherwise the connection has already been closed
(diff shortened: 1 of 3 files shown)


gdude2002:
You'd be surprised how many tools and utils rely on bugs in the Mojang server.

I think it would probably be best to match it as best we can.

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
```diff
@@ -0,0 +1,101 @@
...
+        boolean legacyPingProtocol = false;
+
+        try {
+            if (bytebuf.readByte() == (byte) 0xFE) {
+                GlowServer server = networkServer.getServer();
+                int readableBytes = bytebuf.readableBytes();
+
+                InetSocketAddress inetsocketaddress = (InetSocketAddress) channelHandlerContext.channel().remoteAddress();
+
+                ServerListPingEvent legacyPingEvent = new ServerListPingEvent(inetsocketaddress.getAddress(), server.getMotd(), server.getOnlinePlayers().size(), server.getMaxPlayers());
```
kamcio96, line 40: ServerListPingEvent is sync event and should be handled in main thread. Now "netty" thread will wait for main loop tick. It will cause some lag. Add LegacyStatusRequest packet here :wink:

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+    }
+
+    private void sendByteBuf(ChannelHandlerContext channelhandlercontext, ByteBuf bytebuf) {
+        channelhandlercontext.writeAndFlush(bytebuf).addListener(ChannelFutureListener.CLOSE);
+    }
+
+    private ByteBuf responseToByteBuf(String string) {
+        ByteBuf bytebuf = Unpooled.buffer();
```
kamcio96: You know length `ctx.alloc().buffer(length + 3);`

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
```diff
@@ -0,0 +1,101 @@
...
+
+        bytebuf.writeByte(0xFF);
+        bytebuf.writeShort(string.length());
+        try {
+            bytebuf.writeBytes(string.getBytes("UTF-16BE"));
```
kamcio96, line 93: I'm not sure `string.length()` and `string.getBytes(...).length` is always equal.

GreenBeard, line 93: They aren't equal you're right but that's not what the specification does. It sends the number of characters not the length of the data.

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
GreenBeard, line 40: I am confused on what you are saying I should change :confused: . I understand that it will wait for the main thread (which will introduce minor lag) but that allows for plugins to modify the response. Isn't that what the event is for? If I'm completely misunderstanding you could you restate what you said.

Edit: If I am not misunderstanding you (which I probably am misunderstanding you), then doesn't the current 'modern' status query/ping have the "netty" thread wait for the main loop? I am looking at [Line 32 to 37 of StatusRequestHandler.java](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java#L32)

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
GreenBeard: Thanks updated.

GreenBeard:
@gdude2002 After double checking what the normal notchian server does I believe that I was initially incorrect (about the second byte not being checked, it actually is checked). However ~~spigot~~ bungeecord servers don't appear to care what junk I send them as long as it starts with 0xFE which is what confused my earlier.

Should I mimic the notchian server and not respond to faulty pings? Or should I just allow any garbage like the ~~spigot~~ bungeecord servers? **Note the craftbukkit servers act like the normal notchian servers and ignore faulty/junk legacy pings**

Edit: Servers I had assumed were using spigot turned out to be using bungeecord and I must apologize and say that the above information is incorrect, spigot servers ignore bad ping requests. However bungeecord (which the servers were actually running) accepts them. After realizing this I recommend that we don't allow improper pings because only bungeecord responds to them so anyone who makes a mistake will realize it isn't working properly and fix it if they do basic testing on their scripts (I am assuming only scripts will use legacy pings/queries).

gdude2002:
Well, shit. xD

I.. don't know, honestly. What do you think would be best?

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
Cybermaxke: Why do you not use the `ByteBufAllocator`? It can be retrieved through `channelhandlercontext#alloc`.

GreenBeard:
@gdude2002 I think that the pull request is ready. If you want me to squash the two commits into one I can do that.

gdude2002:
If you feel up to it, please do that. I'm in bed right now, and won't be
able to merge this for 8 or 9 hours, so ping another contributor on gitter
or IRC if you want a quick review!

On 29 Jun 2016 00:35, "GreenBeard" notifications@github.com wrote:

@gdude2002 https://github.com/gdude2002 I think that the pull request is
ready. If you want me to squash the two commits into one I can do that.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
https://github.com/GlowstoneMC/Glowstone/pull/304#issuecomment-229215873,
or mute the thread
https://github.com/notifications/unsubscribe/AAMdeVq3Xw9Uh3NIQqgKylVrmh6abuXbks5qQa_agaJpZM4I9QIP
.

kamcio96:
@greenbeard you shouldn't call it from netty thread(I know bukkit does it). You can add boolean or other field to StatusRequest packet, named legacy. Decode it in your channel handler and then handle like other packets. 

and writing string length it's really really bad idea :/

GreenBeard:
@kamcio96 

> you shouldn't call it from netty thread(I know bukkit does it). You can add boolean or other field to StatusRequest packet, named legacy. Decode it in your channel handler and then handle like other packets. 

I honestly don't understand the benefits. It can't respond the same way (with the same regular packet) so that would be difficult and all the benefit is lost because the event can just be used instead. Also to respond the fact that the netty thread might freeze when it is waiting for the 'main loop tick', that doesn't matter at all because each client/connection **has it's own thread** (I'm pretty sure but someone correct me if I am completely wrong) so only the legacy ping response will lag which doesn't really matter anyway because it will be less than a 1ms or 2.

> and writing string length it's really really bad idea :/

So what do you want me to do? **That's how the legacy packet works!** I agree it's stupid that it uses the string length and not the data's length but we are trying to actually have a functioning response here. We can't change the legacy protocol otherwise it isn't compatible :neutral_face:

GreenBeard:
@gdude2002 I squashed the two commits into one and it should be ready to be pulled.

mastercoms:
In the future, please don't squash PRs.

GreenBeard:
@mastercoms I asked if I should and @gdude2002 said that I should

> If you want me to squash the two commits into one I can do that.

me

> If you feel up to it, please do that.

gdude2002

Also the second commit was literally just changing `ByteBuf bytebuf = Unpooled.buffer();` to `ByteBuf bytebuf = ctx.alloc().buffer(3 + string.length());` and the function and calls to have ctx as an argument.

src/main/java/net/glowstone/net/handler/legacyping/LegacyPingHandler.java
```diff
@@ -0,0 +1,100 @@
...
+                            if (clientVersion >= 73 && 7 + (hostname.length() * 2) == dataLength && bytebuf.readableBytes() == 0) {
+                                sendByteBuf(channelHandlerContext, responseToByteBuf(channelHandlerContext, String.format("\u00a71\u0000%d\u0000%s\u0000%s\u0000%d\u0000%d", GlowServer.PROTOCOL_VERSION, GlowServer.GAME_VERSION, legacyPingEvent.getMotd(), legacyPingEvent.getNumPlayers(), legacyPingEvent.getMaxPlayers())));
+                                legacyPingProtocol = true;
+                            }
+                        }
+                        break;
+                }
+            }
+        } catch (RuntimeException e) {
+            // Silently catch the exception
```
mastercoms, line 71: Log the exception

aramperes:
@GreenBeard 
Is this Pull Request in-progress, or is it still ready for review? Have not heard of this for a few weeks.

GreenBeard:
@momothereal I think it is ready to be reviewed.

aramperes:
@GreenBeard alright, I'll do some benchmarks tomorrow and hopefully merge it.

aramperes:
Tested and looks good, thanks for your contribution! Will be merged ASAP.

Issue #297, opened 2016-06-14
[$15] Internationalization [$15]


BlackTrojan:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=35136041)](https://www.bountysource.com/issues/35136041-internationalization?utm_source=35136041&utm_medium=shield&utm_campaign=ISSUE_BADGE)

I wanted see customization messages, eg when the command does not exist or when the version player does not match the version of the server.

mastercoms:
Glowstone uses hard coded messages all throughout the code. It would be smart to add localization files, as you propose.

gdude2002:
This ticket needs categorization on the [waffle board](https://waffle.io/GlowstoneMC/Glowstone).

Has this ticket seen any movement or has there been any action expected since the previous comment?

tsengia:
I will start working on this now, idk if someone could assign me to this, but I'm working on it now.

gdude2002:
Thanks - I've updated Waffle accordingly.

tsengia:
Should the server handle different languages for each player, or should the server have one set language?

aramperes:
Note: this issue (and therefore the bounty set on it) is about the task of putting all the strings around Glowstone in a file. The task of creating the framework/API to make this possible is in issue #873.

Pr0methean:
Who owns the translations on Mojang's Crowdin? If we could import them and match up the string IDs, we'd have instant support for all the locales that Minecraft has.

aramperes:
We contacted Mojang about the licensing of the translation files in the past, which they deflected. To be on the safe side, we decided to exclude the files as they might be considered assets owned by Mojang.

PR #315, opened 2016-07-03
[WIP] Internationalization Support

tsengia:
This PR will solve issue #297 by adding in internationalization support to Glowstone.
A `LanguageManager` class has been added to handle the retrieval of internationalized Strings.
It also provides a method to retrieve the internationalized String according to the server's settings.
The `default-locale` setting will set the language/locale to be used on the console and by default if `local-locale` is set to true. If `lock-locale` is set to false, the player's locale will be used to relay messages to them.
Currently there is only a few hard-coded messages that have been replaced with internationalized support, and more languages other than English needs to be added.


Paulomart:
Good, but we need arguments that can be put into translated strings. Using the formatter would make sense. Eg: String::format

tsengia:
@Paulomart noted, and will add in

mastercoms:
Looks nice so far.

Since you're changing logger usage already, could you also make sure that Glowstone always uses logger.loglevel (like logger.info) for normal messages, and logger.log(Level, message, throwable) for error messages. Also, please search for usages of System.out.print and other logging functions if any and change them to use logger.

Also, sorry for the formatting messages (I hate dealing with formatting, too), but I have to wait until this PR is not WIP to comment on actual code.

Finally, it seems like you changed the permissions for the files that you modified (644 to 755). Please change them back if you can.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -295,7 +294,7 @@ public void playEffect(Location location, Effect effect, int id, int data, float
      * The one block the player is currently digging.
      */
     private GlowBlock digging;
-
+
```
mastercoms, line 297: Excess diff

src/main/java/net/glowstone/util/ServerConfig.java
```diff
@@ -377,7 +379,7 @@ private boolean migrate() {
         PERMISSIONS_FILE("files.permissions", "permissions.yml", Migrate.BUKKIT, "settings.permissions-file"),
         COMMANDS_FILE("files.commands", "commands.yml"),
         HELP_FILE("files.help", "help.yml"),
-
+
```
mastercoms, line 382: Excess diff

src/main/java/net/glowstone/util/lang/LanguageManager.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /*
+     * Returns the internationalized message according to the effective locale of the player.
+     */
+    public String getString(String key, GlowPlayer p) {
+        String locale = this.getEffectiveLocale(p);
+        ResourceBundle rb = ResourceBundle.getBundle(baseName, Locale.forLanguageTag(locale));
+        return rb.getString(key);
+    }
+    /**
```
mastercoms: Add a new line

src/main/java/net/glowstone/util/lang/LanguageManager.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.ResourceBundle;
+
+import net.glowstone.GlowServer;
+import net.glowstone.entity.GlowPlayer;
+
+
+public class LanguageManager {
+    private String baseName = "lang/messages";
+    private ResourceBundle defaultBundle = ResourceBundle.getBundle(baseName, Locale.forLanguageTag(GlowServer.defaultLocale));
+    /*
```
mastercoms: Add a new line

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -216,10 +219,24 @@
...
+    public static String defaultLocale;
+    /**
+     * Boolean if the default locale in the config should always be used.
+     */
+    public static boolean lockLocale;
+    /**
+     * The LanguageManager for the server. Used for internationalization.
+     */
+    public static LanguageManager languageManager;
     /**
```
mastercoms, line 237: Add a new line

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -216,10 +219,24 @@
      * A view of all online players.
      */
     private final Set<GlowPlayer> onlineView = Collections.unmodifiableSet(onlinePlayers);
+    
+   /**
+    * The default locale to be used in the console and with the locked locale; read from the config.
+    */
+    public static String defaultLocale;
+    /**
```
mastercoms, line 228: Add a new line

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -216,10 +219,24 @@
...
+     * Boolean if the default locale in the config should always be used.
+     */
+    public static boolean lockLocale;
+    /**
```
mastercoms, line 232: Add a new line

Paulomart:
Just an idea: On my server I have a Database with all my strings and the translations for them. It would be cool if plugins could do two things:
1. Hook into this to get and format messages etc.
2. Provide a source of translations from a database, a file etc. This would be checked first if null a fallback would be used.

tsengia:
@Paulomart Or instead the user could install/edit their messages-yourLang.properties file to have the translated Strings for them (or download a translated messages-yourLang.properties file). Additionally, looking at the structure of the code, I think a plugin could easily extend the LanguageManager class and replace the default at start up, so only log messages before it would not be translated, but after it loads, everything would be translated by the plugin.

Paulomart:
@ExpandingDev but theres no api for that. Even after you exteded the manager you need to set it. This would require to find all instances etc. Better to keep one instance and make it not extendable.

tsengia:
@Paulomart The current LanguageManager is stored in GlowServer. There is only 1 GlowServer, and that is also statically typed, so the plugin could access the GlowServer, and use a setter to set the new, extended LanguageManager as the current LanguageManager. There is no need to keep track of every instance of LanguageManager classes. 
Here, some skeleton/protype code to explain:
https://gist.github.com/ExpandingDev/f0c5ce47835290e8d54967c183b44068

tsengia:
@Paulomart Actually, the `lang` field is public, so it does not need a getter or setter, just a basic asignment with `=`

Or can the plugin not access the GlowServer? I thought it could, but if it cannot, then that's a problem.

Paulomart:
If a plugin keeps it's own instance, like that: `LanguageManager lang = GlowServer.lang`.
And you set the GlowServer.lang field the plugin will still keep the old instance. This is not what we want. Just make an Interface LanguageProvider with getText(String key, String... values) and so on. The part that this is not in the API mhh asking @mastercoms

tsengia:
@Paulomart ah, ok, didn't think of that. I will make the interface now

Paulomart:
@ExpandingDev please see Mystalion/GlowstonePlusPlus@aec8f09f9e46e6126aef4d61cdb744cd4e08b1a3

tsengia:
@Paulomart Oh, yeah that seems to be a better idea. Can we/I merge your commit into the PR somehow? Or should I merge your fork into my fork? Or make an entirely different PR for Glowstone?

tsengia:
@Paulomart Nvm, got it

Paulomart:
Ok then :)

PR #322, opened 2016-07-14
Fixed bug for GlowstoneServer#getName, getVersion and getBukkitVersion

VaiTon:
Specified methods could return "null". That could causes freezes for
client at "Downloading terrain...".

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -70,7 +70,6 @@ import org.bukkit.util.permissions.DefaultPermissions;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
 import org.json.simple.parser.ParseException;
-import net.glowstone.util.Metrics;
 
 import java.awt.image.BufferedImage;
 import java.io.File;
@@ -926,6 +925,7 @@ public final class GlowServer implements Server {
 
     /**
      * Returns the list of whitelisted players on this server.
+     *
      * @return A file containing a list of UUIDs for this server's whitelisted players.
      */
     public UuidListFile getWhitelist() {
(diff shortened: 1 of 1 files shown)


mastercoms:
In what cases could these return null?

VaiTon:
When it's not packaged by maven

gdude2002:
Can you elaborate on that? Is this something we should care about?

mastercoms:
Ok. I was just seeing if it was a bug in the code that we could fix directly.

Paulomart:
We build using maven, if somebody uses another build method, it's unsupported. Would not recommend pulling.

VaiTon:
Simply added a validation: when not builded with _maven_ it'll show `UKNOWN`.
Nothing to be concerned about.

Paulomart:
I know that your code does. Its use less in 99.99% of any use cases. If you need this consider making your own patch. But in my opinion this is just unneeded.

mastercoms:
But what's the harm in pulling this?

zml2008:
This could be useful when running GS from an IDE -- the full build setup isn't present, but it's still useful to attach a debugger to the server sometimes, so as long as it doesn't break things (my guess here is plugins expecting version to be in a certain format -- which they shouldn't because it's an implementation detail but whatever) it's a net benefit.

Paulomart:
It won't harm, it's just code that is not needed and complicates things.

gdude2002:
@zml2008 Pretty much every Java IDE I've used can build and run using Maven, and attach a debugger to that. This seems like a non-issue.

Just to put my vote in, I'd vote against pulling this. I did suggest creating the PR in the first place, but I thought it was an actual bug, not something caused by using an unsupported build system or method of building. You should be using Maven, and if you have an IDE that doesn't support it, then you need an IDE that's actually capable of handling Java, in my opinion.

mastercoms:
Rejecting this pull as it encourages unsupported compilation, and may lead to more work in the future due to commiting work to supporting non-maven builds.

PR #330, opened 2016-07-30
Fix containers not dropping contents in creative.

### LINKED ISSUE #328: Destroying an inventory-holding block in creative doesn't drop contents.
When in creative mode, destroying a inventory block such as chests, hoppers, droppers and dispensers will not drop the contents of the inventory on the ground.

Note that this does not occur in survival mode (works as intended).

aramperes:
Fixes #328.

No extra information needed.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockContainer.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockContainer.java
@@ -49,7 +49,7 @@ public class BlockContainer extends BlockType {
         return drops;
     }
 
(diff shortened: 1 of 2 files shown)


mastercoms:
please always use braces :)

aramperes:
Could've use a ternary there instead. Don't blame me for making it clearer than it could've been ;)

mastercoms:
oh yes use ternary it is much more clear

aramperes:
you dropped this: /s

PR #332, opened 2016-07-31
GlowBlock cache v2

kamcio96:


mastercoms:
World gen is taking a long time. Still a lot of performance issues, too.

PR #334, opened 2016-08-07
Hoppers and tile tick format

aramperes:
The goal of this PR is to add the functionality of hoppers. Parts of the TileTick Format (#60) are also implemented since they need to continue pulsing after server restart.

Hoppers To-Do:
- [x] Transfer items to another container
- [x] "Swallow" items above the hopper and from containers above
- [x] Locking hoppers with redstone signal


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -403,7 +403,33 @@ public void onEntityStep(GlowBlock block, LivingEntity entity) {
...
-    public void requestPulse(GlowBlockState state) {
-        // do nothing
+    /**
+     * Request pulsing to the block
+     *
+     * @param block the block to pulse
+     */
+    public void requestPulse(GlowBlock block) {
+        if (getPulseTickSpeed() != null) {
+            block.getWorld().requestPulse(block, getPulseTickSpeed(), isPulseOnce());
```
mastercoms: Should we just access the block type methods when we pulse a block?

aramperes: Hmm maybe, but I don't know if that's necessary. The pulse map would have to access the ItemTable a lot more than it currently does.

mastercoms: Doesn't it already get block type?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1790,6 +1790,15 @@ private void pulseTickMap() {
         return tickMap;
     }
 
+    public void requestPulse(GlowBlock block) {
+        ItemTable itemTable = ItemTable.instance();
+        BlockType type = itemTable.getBlock(block.getType());
+        if (type == null || type.getPulseTickSpeed() == null) {
```
mastercoms: I meant in pulseTickMap(), so that we don't have to store those values any more. So less memory.

aramperes:
Looks like this PR is complete feature-wise, and it even (unintentionally) fixed some bugs related to water!

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1,5 +1,6 @@
 package net.glowstone;
 
+import io.netty.util.internal.ConcurrentSet;
```
Paulomart: Please consider using another implementation. Netty uses this API only internal and they are not part of the API.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -419,4 +419,23 @@ public void onRedstoneUpdate(GlowBlock block) {
     public void onEntityStep(GlowBlock block, LivingEntity entity) {
         // do nothing
     }
+
+    /**
+     * The rate at which the block should be pulsed.
+     *
+     * @return null if the block should not pulse, or a number of ticks between pulses.
+     */
+    public Integer getPulseTickSpeed(GlowBlock block) {
```
mastercoms: Why Integer and not int?

src/main/java/net/glowstone/generator/decorators/nether/LavaDecorator.java
```diff
@@ -52,7 +53,7 @@ public void decorate(World world, Random random, Chunk source) {
                 BlockState state = block.getState();
                 state.setType(Material.LAVA);
                 state.update(true);
-                ((GlowWorld) block.getWorld()).requestPulse((GlowBlock) block, 1, true);
+                Bukkit.getScheduler().scheduleSyncDelayedTask(null, new PulseTask((GlowBlock) block), 1);
```
mastercoms: Have you tried doing this task async?

src/main/java/net/glowstone/GlowWorld.java
mastercoms: It works, does it not?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1774,59 +1774,45 @@ public void sendPluginMessage(Plugin source, String channel, byte[] message) {
 
     private void pulseTickMap() {
```
mastercoms, line 1775: Remove this method and move it to requestPulse(). Then in requestPulse(), schedule a PulseTask with different scheduler methods depending on whether it is single or not. And perhaps add an isPulseAsync getter to BlockType.

src/main/java/net/glowstone/GlowWorld.java
Paulomart: "It works" should not be the quality we aim for. Look at package-info.java in Netty.

src/main/java/net/glowstone/GlowWorld.java
aramperes, line 1775: What BlockType would require an async/sync pulse?

src/main/java/net/glowstone/block/blocktype/BlockType.java
aramperes: I chose Integer because you can return null. It's a lazy way of doing it; it checks if getPulseTickSpeed is null, then it doesn't require a pulse at all. Overriding and returning a value will give the tick delay.

src/main/java/net/glowstone/GlowWorld.java
aramperes, line 1775: Also, that would break with block that need to change their tick speed without resetting pulse; like command blocks for instance. I would just keep the current system.

src/main/java/net/glowstone/block/blocktype/BlockType.java
mastercoms: How about -1 or 0?

mastercoms: I just don't like using `null` as a value most of the time, especially when there is a good alternative.

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -190,7 +189,7 @@ private void flow(GlowBlock source, BlockFace direction) {
         }
         // flow to the target
         ((GlowBlock) fromToEvent.getToBlock()).setType(fromToEvent.getBlock().getType(), strength, true);
-        ((GlowWorld) fromToEvent.getToBlock().getWorld()).requestPulse(((GlowBlock) fromToEvent.getToBlock()), isWater(fromToEvent.getToBlock().getType()) || fromToEvent.getToBlock().getBiome() == Biome.HELL ? TICK_RATE_WATER : TICK_RATE_LAVA, true);
+        //((GlowWorld) fromToEvent.getToBlock().getWorld()).requestPulse(((GlowBlock) fromToEvent.getToBlock()));
```
mastercoms: Why was this line commented? It breaks water.

aramperes: I don't actually remember; let me check...

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -255,4 +254,13 @@ public boolean canTickRandomly() {
         return true;
     }
 
+    @Override
+    public boolean isPulseOnce(GlowBlock block) {
+        return false;
```
mastercoms: Water is a pulse once block, since each flow from one block to the other only happens once.

src/main/java/net/glowstone/io/anvil/AnvilChunkIoService.java
```diff
@@ -138,6 +143,28 @@ public boolean read(GlowChunk chunk) throws IOException {
...
         }
 
+        if (levelTag.isList("TileTicks", TagType.COMPOUND)) {
+            List<CompoundTag> tileTicks = levelTag.getCompoundList("TileTicks");
+            for (CompoundTag tileTick : tileTicks) {
+                int tileX = tileTick.getInt("x");
+                int tileY = tileTick.getInt("y");
+                int tileZ = tileTick.getInt("z");
+                String id = tileTick.getString("i");
+                Material material = Material.getMaterial(id);
```
mastercoms: IDs are formatted as `provider:material`, e.g `minecraft:flowing_lava`. Also, there is no Bukkit material called `FLOWING_LAVA`. It's just `LAVA`. Stationary lava in vanilla is called `minecraft:lava`, while in Bukkit it's called `STATIONARY_LAVA`.

Don't worry about fixing it, I have already fixed it in my merge conflicts commit.

aramperes: Hmm, yeah I haven't thought about that! Good thing it is fixed then

src/main/java/net/glowstone/block/blocktype/BlockLiquid.java
```diff
@@ -188,8 +188,9 @@ private void flow(GlowBlock source, BlockFace direction) {
             strength = STRENGTH_MAX;
         }
         // flow to the target
-        ((GlowBlock) fromToEvent.getToBlock()).setType(fromToEvent.getBlock().getType(), strength, true);
-        //((GlowWorld) fromToEvent.getToBlock().getWorld()).requestPulse(((GlowBlock) fromToEvent.getToBlock()));
+        GlowBlock toBlock = (GlowBlock) fromToEvent.getToBlock();
+        toBlock.setType(fromToEvent.getBlock().getType(), strength, true);
```
mastercoms: Updating physics flows water. So this would be a duplicate action.

aramperes: So should I change `applyPhysics` to false or remove the pulse?

src/main/java/net/glowstone/block/blocktype/BlockRedstoneTorch.java
```diff
@@ -55,7 +55,7 @@ public void placeBlock(GlowPlayer player, GlowBlockState state, BlockFace face,
     @Override
     public void updatePhysics(GlowBlock me) {
```
mastercoms, line 56: Redstone torches are faulty. Many standard redstone structures were working before, however, they are not working now. I am not sure if this is a result of this PR, but this PR should fix the issue, providing that we just start the pulse incorrectly.

An error also occurs while using timers (which did not happen before):

```
java.lang.ArrayIndexOutOfBoundsException
```

aramperes: @mastercoms I can't really reproduce what you are saying, for me redstone mechanisms work pretty fine.. can you post a screenshot or something?

edit: loop works too for me
edit2: redstone repeaters need to be saved, they are not using the tick map at the moment and freeze when saving the world.

mastercoms: [redstone timer](https://postimg.org/image/c9670bhzj/)

doesn't work after about 10 seconds

aramperes: Seems to be a problem with concurrency; this is the full stacktrace:

```
17:56:11 [SEVERE] Error occurred while pulsing world world
java.lang.ArrayIndexOutOfBoundsException: 35
        at java.util.LinkedList.toArray(Unknown Source)
        at java.util.ArrayList.<init>(Unknown Source)
        at net.glowstone.entity.GlowPlayer.processBlockChanges(GlowPlayer.java:619)
        at net.glowstone.entity.GlowPlayer.pulse(GlowPlayer.java:568)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(Unknown Source)
        at java.util.stream.ReferencePipeline$2$1.accept(Unknown Source)
        at java.util.LinkedList$LLSpliterator.forEachRemaining(Unknown Source)
        at java.util.stream.AbstractPipeline.copyInto(Unknown Source)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
        at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(Unknown Source)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(Unknown Source)
        at java.util.stream.AbstractPipeline.evaluate(Unknown Source)
        at java.util.stream.ReferencePipeline.forEach(Unknown Source)
        at net.glowstone.GlowWorld.pulsePlayers(GlowWorld.java:483)
        at net.glowstone.GlowWorld.pulse(GlowWorld.java:364)
        at net.glowstone.scheduler.WorldScheduler$WorldThread.run(WorldScheduler.java:139)
        at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
        at java.util.concurrent.FutureTask.run(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.lang.Thread.run(Unknown Source)
```

Something tells me that `blockChanges` list is out of sync of the tick map loop.

aramperes: @mastercoms Can you test it with the latest commit? I have yet to encounter the bug with that change

src/main/java/net/glowstone/io/anvil/AnvilChunkIoService.java
```diff
@@ -150,7 +150,15 @@ public boolean read(GlowChunk chunk) throws IOException {
                 int tileY = tileTick.getInt("y");
                 int tileZ = tileTick.getInt("z");
                 String id = tileTick.getString("i");
-                Material material = Material.getMaterial(id);
+                if (id.startsWith("minecraft:")) {
```
mastercoms, line 153: wrap the whole thing in this if statement, and else, say there was an unsupported tile tick from a mod

src/main/java/net/glowstone/io/anvil/AnvilChunkIoService.java
```diff
@@ -150,7 +150,15 @@ public boolean read(GlowChunk chunk) throws IOException {
...
+                    id = id.replace("minecraft:", "");
+                    if (id.startsWith("flowing_")) {
+                        id = id.replace("flowing_", "");
+                    } else if ("id".equals("water") || id.equals("lava")) {
```
mastercoms, line 157: you accidentally wrote a constant `"id"`. :)

mastercoms:
Hoppers pull before they push.

push before pull example

This should fill the bottom chest first and then when the bottom hopper fills, it should go to the middle chest.

src/main/java/net/glowstone/block/blocktype/BlockHopper.java
```diff
@@ -50,10 +60,127 @@ public TileEntity createTileEntity(GlowChunk chunk, int cx, int cy, int cz) {
...
+    }
+
+    @Override
+    public void onRedstoneUpdate(GlowBlock block) {
+        ((Hopper) block.getState().getData()).setActive(!block.isBlockPowered());
+    }
+
+    private void pullItems(GlowBlock block, TEHopper hopper) {
+        GlowBlock source = block.getRelative(BlockFace.UP);
+        if (source.getType() == null || source.getType() == Material.AIR) {
```
picknchew: Hoppers with slabs or signs above will not work. getType() will never return null, only air.

PickNChew:
Items from the hopper should not drop when the player breaks the block in creative mode.

mastercoms:
~~Bottom hopper does not transfer sideways.~~

EDIT: due to the hopper not being connected, ok

PR #339, opened 2016-08-19
Ameliorate sounds internally

aramperes:
This PR has the goal to improve the usage of sounds internally, by introducing utilities for different sound variations, as well as add new sound triggers for blocks, entities, etc.

Currently, the PR adds a new utility, `SoundUtil`, as well as changes to these sound triggers:
- Add Button release sound
- Add lever power switch sound
- Fix a duplication of sound when placing a block. This is because the client placing the block plays the sound without the server having to send it. The server should only send placement sounds to other players around the block.
- Add hurt/death sounds for all mobs

**I am open to suggestions on what to add for this (anything related to sounds)!**


src/main/java/net/glowstone/util/SoundUtil.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+public class SoundUtil {
+
+    private static final ThreadLocalRandom rand = ThreadLocalRandom.current();
+
+    public static void playSoundAtLocationExcept(Location location, Sound sound, float volume, float pitch, GlowPlayer... exclude) {
+        if (location == null || sound == null) return;
+        GlowWorld world = (GlowWorld) location.getWorld();
+        double radiusSquared = Math.pow(volume * 16, 2);
```
mastercoms: I think `volume * volume * 256` would be more optimized.

src/main/java/net/glowstone/util/SoundUtil.java
```diff
@@ -0,0 +1,40 @@
...
+        world.getRawPlayers().stream().filter(player -> player.getLocation().distanceSquared(location) <= radiusSquared).filter(player -> !Arrays.asList(exclude).contains(player)).forEach(player -> player.playSound(location, sound, volume, pitch));
```
mastercoms: I don't think using `parallelStream()` will result in any concurrency issues.

src/main/java/net/glowstone/util/SoundUtil.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public static void playSoundPitchRange(Location location, Sound sound, float volume, float pitchBase, float pitchRange, GlowPlayer... exclude) {
+        playSoundPitchRange(location, sound, volume, pitchBase, pitchRange, true, exclude);
+    }
+
+    public static float randomReal(float range) {
+        return (rand.nextFloat() - rand.nextFloat()) * range;
+    }
+
```
mastercoms: Remove this extra line, please

src/main/java/net/glowstone/util/SoundUtil.java
```diff
@@ -0,0 +1,40 @@
...
+import net.glowstone.entity.GlowPlayer;
+import org.bukkit.Location;
+import org.bukkit.Sound;
+
+import java.util.Arrays;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class SoundUtil {
+
+    private static final ThreadLocalRandom rand = ThreadLocalRandom.current();
```
mastercoms: Wouldn't this result in the same ThreadLocalRandom being used across all threads?

src/main/java/net/glowstone/util/SoundUtil.java
aramperes: I'm not sure what you mean, what do you want changed there?

mastercoms: world.getRawPlayers().stream() could use [`parallelStream()`](http://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#parallelStream--) instead, since I believe it's safe to use it here as previous elements don't affect further elements in the list and the methods being used are thread safe, I think.

mastercoms (approved):
Apparently I can't view that you've resolved my change requests until I mark them as resolved. And, some changes are not shown until I review them. GitHub's logic is great.

src/main/java/net/glowstone/util/SoundUtil.java
mastercoms: omg lol it still doesn't mark these as resolved even after I submitted a new review stating that

PR #364, opened 2016-08-31
Animals drop items

valleydev:
Some animals drop items now, but not xp. No grilled flesh is dropped. Wool is not same color as sheep.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -458,8 +458,7 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
         }
     }
 
-    public void kill()
-    {
+    public void kill() {
         active = false;
         Sound deathSound = getDeathSound();
         if (deathSound != null) {
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -454,20 +454,24 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
         }
 
         if (health <= 0) {
-            active = false;
-            Sound deathSound = getDeathSound();
-            if (deathSound != null) {
-                world.playSound(location, deathSound, 1.0f, 1.0f);
-            }
-            playEffect(EntityEffect.DEATH);
-            if (this instanceof GlowPlayer) {
-                PlayerDeathEvent event = new PlayerDeathEvent((GlowPlayer) this, new ArrayList<>(), 0, this.getName() + " died.");
-                EventFactory.callEvent(event);
-                server.broadcastMessage(event.getDeathMessage());
-            } else {
-                EventFactory.callEvent(new EntityDeathEvent(this, new ArrayList<>()));
-            }
-            // todo: drop items
+            kill();
+        }
+    }
+
+    public void kill()
+    {
+        active = false;
+        Sound deathSound = getDeathSound();
+        if (deathSound != null) {
+            world.playSound(location, deathSound, 1.0f, 1.0f);
+        }
+        playEffect(EntityEffect.DEATH);
+        if (this instanceof GlowPlayer) {
+            PlayerDeathEvent event = new PlayerDeathEvent((GlowPlayer) this, new ArrayList<>(), 0, this.getName() + " died.");
+            EventFactory.callEvent(event);
+            server.broadcastMessage(event.getDeathMessage());
+        } else {
+            EventFactory.callEvent(new EntityDeathEvent(this, new ArrayList<>()));
         }
     }
 
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -378,7 +378,11 @@ public final class GlowPlayer extends GlowHumanEntity implements Player {
         setGameModeDefaults();
 
         // send server brand and supported plugin channels
-        session.send(PluginMessage.fromString("MC|Brand", server.getName()));
+        String serverImplName = server.getName();
+        if (serverImplName == null) // blame windows 10 + oracle java
+            serverImplName = "";
+
+        session.send(PluginMessage.fromString("MC|Brand", serverImplName));
         sendSupportedChannels();
         joinTime = System.currentTimeMillis();
         reader.readData(this);
--- a/src/main/java/net/glowstone/entity/passive/GlowChicken.java
+++ b/src/main/java/net/glowstone/entity/passive/GlowChicken.java
@@ -54,8 +54,7 @@ public class GlowChicken extends GlowAnimal implements Chicken {
     }
 
     @Override
-    public void kill()
-    {
+    public void kill() {
         super.kill();
 
         Random r = new Random();
--- a/src/main/java/net/glowstone/entity/passive/GlowChicken.java
+++ b/src/main/java/net/glowstone/entity/passive/GlowChicken.java
@@ -52,4 +52,16 @@ public class GlowChicken extends GlowAnimal implements Chicken {
             generateEggLayDelay();
         }
     }
+
+    @Override
+    public void kill()
+    {
+        super.kill();
+
+        Random r = new Random();
+
+        getWorld().dropItemNaturally(getLocation(), new ItemStack(Material.FEATHER, r.nextInt(2)));
+        getWorld().dropItemNaturally(getLocation(), new ItemStack(Material.RAW_CHICKEN, r.nextInt(1)));
+    }
+
 }
--- a/src/main/java/net/glowstone/entity/passive/GlowCow.java
+++ b/src/main/java/net/glowstone/entity/passive/GlowCow.java
@@ -21,8 +21,7 @@ public class GlowCow extends GlowAnimal implements Cow {
     }
 
     @Override
-    public void kill()
-    {
+    public void kill() {
         super.kill();
 
         Random r = new Random();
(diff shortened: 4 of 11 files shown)


mastercoms:
You should always make changes on a separate branch, instead of master, so that you can create new pull requests later. So make a branch like `animal-drops`, and put your changes in there. Also, make sure that your [email](https://help.github.com/articles/setting-your-email-in-git/) and [username](https://help.github.com/articles/setting-your-username-in-git/) are properly set in git and match your GitHub information.

Please create a new pull request with these changes.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -454,20 +454,23 @@ public void setHealth(double health) {
...
+    }
+
+    public void kill() {
+        active = false;
+        Sound deathSound = getDeathSound();
+        if (deathSound != null) {
+            world.playSound(location, deathSound, 1.0f, 1.0f);
+        }
+        playEffect(EntityEffect.DEATH);
+        if (this instanceof GlowPlayer) {
```
mastercoms: Since you're making a new kill() method for each entity, it would be a good idea to move this to GlowPlayer.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -378,7 +378,11 @@ public void join(GlowSession session, PlayerReader reader) {
         setGameModeDefaults();
 
         // send server brand and supported plugin channels
-        session.send(PluginMessage.fromString("MC|Brand", server.getName()));
+        String serverImplName = server.getName();
```
mastercoms: Please explain this change

valleydev:
where can i set git username and mail in itellij ?

aramperes:
@valleydev In your console/bash:

`git config --global user.name "valleydev"`
`git config --global user.email "someone@somewhere.com"`

valleydev:
Will this request be approved?
Feedback appreciated

src/main/java/net/glowstone/entity/passive/GlowChicken.java
```diff
@@ -52,4 +52,15 @@ public void pulse() {
             generateEggLayDelay();
         }
     }
+
+    @Override
+    public void kill() {
+        super.kill();
+
+        Random r = new Random();
```
aramperes, line 60: Please cache Random in a util thread, and use `ThreadLocalRandom.current()`

src/main/java/net/glowstone/entity/passive/GlowOcelot.java
```diff
@@ -31,4 +31,6 @@ public void setOwner(AnimalTamer animalTamer) {
         // TODO
         super.setOwner(animalTamer);
     }
+
+
```
aramperes, line 35: Extra spaces

Paulomart:
I don't like the new kill method. If a superclass doesn't call kill in super it may cause problems. Better introduce a new method that only handle's the item dropping. There was a discussion how to implement loot tables, but I don't know what's the state on that.

valleydev:
Agreed on Paulomart 
new virtual method: 
`public void dropItemsAfterDeath( deathreason )`

Please help: i need the death reason as an argument e.g. burned in lava drops grilled chicken instead of raw chicken. Can you please make a suggestion on a type that can be used for death?

Instead of 
`Random r = new Random();`

i should use
`Random r = ThreadLocalRandom.current();`

correct?

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -378,7 +378,11 @@ public void join(GlowSession session, PlayerReader reader) {
...
+        if (serverImplName == null) // blame windows 10 + oracle java
+            serverImplName = "";
```
mastercoms, line 383: Maybe use a better default value here?

src/main/java/net/glowstone/entity/passive/GlowCow.java
```diff
@@ -19,6 +21,19 @@ public GlowCow(Location location) {
...
     @Override
+    public void kill() {
+        super.kill();
+
+        Random r = new Random();
+
+        int beefCnt = r.nextInt(3);
+        int leatherCnt = r.nextInt(2);
+
+        getWorld().dropItemNaturally(getLocation(), new ItemStack(Material.RAW_BEEF, beefCnt));
```
mastercoms, line 32: I would like the design to be more like the block drops system:

https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/blocktype/BlockType.java#L89

https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/blocktype/BlockGrass.java#L23

src/main/java/net/glowstone/entity/GlowPlayer.java
valleydev: this exception was just annoying every time ...
Which value do you recommend?

src/main/java/net/glowstone/entity/passive/GlowCow.java
valleydev: Your suggestion seems to me more complicated. Can you please provide further explanation why there should be a list?

Or do you recommend this

`public List<ItemsStack> getItemsForDropAfterDeath( deathreason )`

mastercoms:
The damage method contains DamageReason, but does not pass this value to setHealth. https://github.com/valleydev/Glowstone/blob/a7dc7eeab23922856e5d2c067cae2dec298ae4c6/src/main/java/net/glowstone/entity/GlowLivingEntity.java#L493

Perhaps make a new method called `setHealth(double health, DamageCause cause)`.

And yes, that is the correct usage of ThreadLocalRandom.

src/main/java/net/glowstone/entity/passive/GlowCow.java
mastercoms: Yes, something like that would be good.

src/main/java/net/glowstone/entity/GlowPlayer.java
mastercoms: `Glowstone`

http://wiki.vg/Plugin_channels#MC.7CBrand

src/main/java/net/glowstone/entity/passive/GlowCow.java
valleydev: Can you please clarify the differences between a block and an entity?

For both classes some items are dropped after death/destruction.
Maybe we can do an Interface e.g. IDropsItemsAfterDead (.net Interfaces start with capital i, same in java?)

mastercoms: Well, entities die due to damage due to a DamageCause, but blocks get destroyed due to a block break packet with a certain held item.

I guess it would be possible to unify them but it would take a lot of changing of the code and I don't think it will make it cleaner.

valleydev:
Will have a look at this tomorrow. Bukkit-classes are giving me a hard time.

`setHealth` is part of `org.bukkit.entity.damageable` interface and cannot simply be changed.

mastercoms:
I meant add a new method. Make `setHealth(double health)` use `setHealth(double health, DamageCause cause)` and make the cause unknown.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
aramperes, line 468: Add the content of this if-statement in the GlowPlayer class `kill()` implementation.

mastercoms: But this implementation calls super.kill(), which would call `EventFactory.callEvent(new EntityDeathEvent(this, new ArrayList<>()))`

src/main/java/net/glowstone/entity/GlowPlayer.java
mastercoms: Please remove this now, as a fix has been provided in another PR.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
aramperes: Maybe `EntityDeathEvent` also normally called for players? I don't know.

mastercoms: But it's an if-else statement, so it wasn't called before.

olliestanley: EntityDeathEvent should definitely also be called on player death imo. Not doing so would create weird inconsistencies i.e EntityDamageEvent called when a player is damaged but EntityDeathEvent not called when a player dies.

mastercoms: `PlayerDeathEvent` extends `EntityDeathEvent`.

olliestanley: Ah I see. Perhaps a second kill method of protected visibility which has a `silent` parameter so that it can be called by subclasses only when the subclass wishes to fire its own event?

mastercoms: I feel like that isn't very elegant.

olliestanley: Maybe not, but it seems more important to keep references to subclasses out as much as possible than to keep it 100% elegant. It also wouldn't affect anyone using the API.

mastercoms: I guess you're right, but we should still investigate alternatives to that.

mastercoms:
@valleydev any news on this?

mastercoms:
Closing due to inactivity.

PR #367, opened 2016-09-02
Particles

picknchew:
Implement all of the spawnParticle methods in World.
Add support for Particle enum in GlowParticle.
Remove hacky workaround for playEffect method.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1726,8 +1726,7 @@ public final class GlowWorld implements World {
             throw new IllegalArgumentException("wrong data type " + data.getClass() + " should be " + particle.getDataType());
         }
 
-        for (Player player : getPlayers()) {
-            GlowPlayer glowPlayer = (GlowPlayer) player;
+        for (GlowPlayer player : getRawPlayers()) {
             double distance = player.getLocation().distanceSquared(location);
             boolean isLongDistance = GlowParticle.isLongDistance(particle);
 
@@ -1735,7 +1734,7 @@ public final class GlowWorld implements World {
             int[] particleData = GlowParticle.getData(particle, data);
 
             if (distance <= 1024.0D || isLongDistance && distance <= 262144.0D) {
-                glowPlayer.getSession().send(new PlayParticleMessage(particleId, isLongDistance, (float) location.getX(), (float) location.getY(), (float) location.getZ(), (float) offsetX, (float) offsetY, (float) offsetZ, (float) extra, count, particleData));
+                player.getSession().send(new PlayParticleMessage(particleId, isLongDistance, (float) location.getX(), (float) location.getY(), (float) location.getZ(), (float) offsetX, (float) offsetY, (float) offsetZ, (float) extra, count, particleData));
             }
         }
     }
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1731,7 +1731,7 @@ public final class GlowWorld implements World {
             boolean isLongDistance = GlowParticle.isLongDistance(particle);
 
             int particleId = GlowParticle.getId(particle);
-            int[] particleData = GlowParticle.getData(particle, data);
+            int[] particleData = GlowParticle.getExtData(particle, data);
 
             if (distance <= 1024.0D || isLongDistance && distance <= 262144.0D) {
                 player.getSession().send(new PlayParticleMessage(particleId, isLongDistance, (float) location.getX(), (float) location.getY(), (float) location.getZ(), (float) offsetX, (float) offsetY, (float) offsetZ, (float) extra, count, particleData));
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -18,6 +18,7 @@ import net.glowstone.io.WorldMetadataService.WorldFinalValues;
 import net.glowstone.io.WorldStorageProvider;
 import net.glowstone.io.anvil.AnvilWorldStorageProvider;
 import net.glowstone.net.message.play.entity.EntityStatusMessage;
+import net.glowstone.net.message.play.game.PlayParticleMessage;
 import net.glowstone.net.message.play.player.ServerDifficultyMessage;
 import net.glowstone.util.BlockStateDelegate;
 import net.glowstone.util.GameRuleManager;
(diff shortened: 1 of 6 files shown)


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1670,63 +1667,82 @@ public WorldBorder getWorldBorder() {
...
+    public <T> void spawnParticle(Particle particle, Location location, int count, double offsetX, double offsetY, double offsetZ, double extra, T data) {
+        if (particle == null) {
+            throw new IllegalArgumentException("particle cannot be null!");
+        }
+
+        if (data != null && !particle.getDataType().isInstance(data)) {
+            throw new IllegalArgumentException("wrong data type " + data.getClass() + " should be " + particle.getDataType());
+        }
 
+        for (Player player : getPlayers()) {
```
mastercoms: Use `getRawPlayers()` to skip the cast.

src/main/java/net/glowstone/constants/GlowParticle.java
```diff
@@ -71,58 +76,139 @@ private GlowParticle() {
...
+            case MOB_APPEARANCE:
+                return true;
+        }
+
+        return false;
     }
 
-    private static void set(Effect particle, int id) {
-        ids[particle.ordinal()] = id;
+    private static void set(Particle particle, Effect effect, int id) {
```
mastercoms, line 210: Use `Hashtable`, instead of `HashMap`, and catch for `NullPointerException`.

src/main/java/net/glowstone/net/message/play/game/PlayParticleMessage.java
```diff
@@ -10,9 +10,9 @@
     private final boolean longDistance;
     private final float x, y, z;
     private final float ofsX, ofsY, ofsZ;
-    private final float data;
+    private final float extData;
```
mastercoms: Why were these switched? It [causes confusion](https://github.com/GlowstoneMC/Glowstone/pull/367/files#diff-250587d7c6abd10595cec0302287551bR30).

picknchew: Spigot API says that it's the other way around and I agree.

src/main/java/net/glowstone/net/codec/play/game/PlayParticleCodec.java
```diff
@@ -24,11 +24,13 @@ public ByteBuf encode(ByteBuf buf, PlayParticleMessage message) throws IOExcepti
         buf.writeFloat(message.getOfsX());
         buf.writeFloat(message.getOfsY());
         buf.writeFloat(message.getOfsZ());
-        buf.writeFloat(message.getData());
+        buf.writeFloat(message.getExtData());
         buf.writeInt(message.getCount());
-        for (int extData : message.getExtData()) {
+
+        for (int extData : message.getData()) {
```
mastercoms: Change to `int data`. It's confusing right now due to mismatched names.

src/main/java/net/glowstone/constants/GlowParticle.java
picknchew: Why `HashTable`? This will only slow it down. Checking for null is not particularly useful either as these are set values.

src/main/java/net/glowstone/constants/GlowParticle.java
```diff
@@ -13,53 +16,55 @@
 public final class GlowParticle {
 
     private static final int[] EMPTY = new int[0];
-    private static final int[] ids = new int[Effect.values().length];
+    private static final int[] ONE_EMPTY = new int[]{0};
+    private static final int[] TWO_EMPTY = new int[]{0, 0};
+
+    private static final Map<Object, Integer> ids = new HashMap<>();
```
mastercoms, line 22: Actually, why was `ids` changed to `HashMap` in the first place?

src/main/java/net/glowstone/constants/GlowParticle.java
mastercoms: I thought since you made it a `HashMap`, you wanted to support mods adding new particle effects.

src/main/java/net/glowstone/constants/GlowParticle.java
picknchew: Maintainability, readability and also to support the Particle enum without having to create another array.

mastercoms:
Why was rename reverted? It was fine except for the variable in the for each statement.

picknchew:
I realized I dun goofed. Spigot actually says that speed is extra data and that was what the previous rename was supposed to be. I think it's best we stick with what we have now.

mastercoms:
Ah ok. Perhaps the warning messages should be reverted as well?

src/test/java/net/glowstone/constants/ParticleTest.java
```diff
@@ -44,16 +44,19 @@ public void testGetData() {
         switch (particle) {
             case ITEM_BREAK:
                 assertEquals("Wrong getData for " + particle, true, particle.getData() != null);
-                assertArrayEquals("Wrong extData for " + particle, new int[]{Material.STONE.getId(), 1}, GlowParticle.getData(particle, STONE));
+                assertArrayEquals("Wrong data for " + particle, new int[]{Material.STONE.getId(), 1}, GlowParticle.getData(particle, STONE));
```
mastercoms: An artifact from the rename?

src/test/java/net/glowstone/constants/ParticleTest.java
```diff
@@ -44,16 +44,19 @@ public void testGetData() {
...
                 break;
             case TILE_BREAK:
+                assertEquals("Wrong getData for " + particle, true, particle.getData() != null);
```
mastercoms: These are still switched. And I think the warning for `data` should be `Wrong data for`, not `Wrong getData for`.

src/test/java/net/glowstone/constants/ParticleTest.java
```diff
@@ -43,17 +43,20 @@ public void testGetData() {
 
         switch (particle) {
             case ITEM_BREAK:
-                assertEquals("Wrong getData for " + particle, true, particle.getData() != null);
-                assertArrayEquals("Wrong extData for " + particle, new int[]{Material.STONE.getId(), 1}, GlowParticle.getData(particle, STONE));
+                assertEquals("Wrong data for " + particle, true, particle.getData() != null);
+                assertArrayEquals("Wrong extData for " + particle, new int[]{Material.STONE.getId(), 1}, GlowParticle.getExtData(particle, STONE));
```
mastercoms: Do you think that `extra data` would be more readable vs `extData`? I'm not sure because it deviates from the actual variable name, but is a human friendly name.

mastercoms:
Confirmed working with TrailGUI.

PR #368, opened 2016-09-04
Fixed a glitch when running Glowstone without Maven

DaMatrix:
When Glowstone is started inside eclispe (haven't tested any other IDEs yet), GlowServer.getName() is null.  When this happens, the player is not added to the tab list and commands that need the player name don't work ("player not found").

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -260,6 +260,8 @@ public final class GlowServer implements Server {
      * The {@link MaterialValueManager} of this server.
      */
     private MaterialValueManager materialValueManager;
+    
+    private boolean isNameNull;
 
     /**
      * Creates a new server.
@@ -618,6 +620,9 @@ public final class GlowServer implements Server {
         if (!channel.isActive()) {
             logger.warning("Failed to bind rcon. Address already in use?");
         }
+        //Decide this after everything else has been started
+        this.isNameNull = getClass().getPackage().getImplementationTitle() == null;
+
     }
 
     /**
@@ -1090,7 +1095,7 @@ public final class GlowServer implements Server {
 
     @Override
     public String getName() {
-        if (getClass().getPackage().getImplementationTitle() == null)   {
+        if (isNameNull)   {
             return "Glowstone";
         } else {
             return getClass().getPackage().getImplementationTitle();
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -1090,7 +1090,11 @@ public final class GlowServer implements Server {
 
     @Override
     public String getName() {
-        return getClass().getPackage().getImplementationTitle();
+        if (getClass().getPackage().getImplementationTitle() == null)   {
+            return "Glowstone";
+        } else {
+            return getClass().getPackage().getImplementationTitle();
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -378,7 +378,12 @@ public void join(GlowSession session, PlayerReader reader) {
         setGameModeDefaults();
 
         // send server brand and supported plugin channels
-        session.send(PluginMessage.fromString("MC|Brand", server.getName()));
+        //System.out.println("server name: " + server.getName());
+        try {
```
mastercoms: Move this fix to the actual method, please.

src/main/java/net/glowstone/net/message/play/game/PluginMessage.java
```diff
@@ -19,7 +19,11 @@
     public static PluginMessage fromString(String channel, String text) {
         ByteBuf buf = Unpooled.buffer(5 + text.length());
         try {
-            ByteBufUtils.writeUTF8(buf, text);
+            try {
```
mastercoms: I mean in `GlowServer.getName()`. Sorry for not being clear.

DaMatrix: ok, will do

mastercoms:
I'm not sure if this is a won't fix, since it only happens with an unsupported build method. I believe fix was requested and rejected before, but perhaps I'll pull it this time due to the added convenience of being able to run it in your IDE.

src/main/java/net/glowstone/net/message/play/game/PluginMessage.java
```diff
@@ -29,4 +29,3 @@ public static PluginMessage fromString(String channel, String text) {
     }
 
 }
-
```
mastercoms, line 32: You accidentally removed the newline at the end of this file.

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1090,7 +1090,11 @@ public void setPlayerOnline(GlowPlayer player, boolean online) {
 
     @Override
     public String getName() {
-        return getClass().getPackage().getImplementationTitle();
+        if (getClass().getPackage().getImplementationTitle() == null)   {
```
mastercoms: Store it in a variable.

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1090,7 +1095,11 @@ public void setPlayerOnline(GlowPlayer player, boolean online) {
 
     @Override
     public String getName() {
-        return getClass().getPackage().getImplementationTitle();
+        if (isNameNull)   {
```
mastercoms, line 1098: Again, I'm so sorry for not being specific. I meant store the name within the method. Though, what you did right now might actually be better. I'll test and see.

EDIT: Going off your idea, maybe we can store a global name variable that uses a ternary that checks implementation title for null to assign itself.

VaiTon:
I requested it in #322 and it was rejected

mastercoms:
Yes, I acknowledged that it was rejected before but that I was conflicted about this again, since it seems more people are having problems with this.

https://github.com/GlowstoneMC/Glowstone/pull/368#issuecomment-244608991

Anyway, I'm closing this PR due to inactivity.

PR #372, opened 2016-09-11
Fix mistake in AnvilChunkIoService

olliestanley:
Instead of comparing the variable 'id', the String constant "id" was compared instead.


mastercoms:
Oh yeah, I forget to check if @momothereal actually fixed that: https://github.com/GlowstoneMC/Glowstone/pull/334#discussion_r77514062

Issue #282, opened 2016-06-03
You can place blocks at the same location as other entities


aramperes:
Self-explanatory; players can place blocks at the same location as an entity, including yourself. A fix for this issue would have to take into account the size/box of the entity.

mastercoms:
Partially fixed in https://github.com/GlowstoneMC/Glowstone/commit/5104c727e93bc6a40df690e655581736ce139783. Still needs to account for bounding box.

aramperes:
Done in #373

PR #373, opened 2016-09-16
Use entity bounding boxes for block placement

aramperes:
This adds a check for nearby entities that have bounding boxes before placing a block.

Addresses issue #282.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -11,7 +11,6 @@ import net.glowstone.block.itemtype.ItemType;
 import net.glowstone.entity.GlowEntity;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.entity.physics.BlockBoundingBox;
-import net.glowstone.entity.physics.BoundingBox;
 import net.glowstone.util.SoundInfo;
 import net.glowstone.util.SoundUtil;
 import org.bukkit.*;
@@ -308,14 +307,9 @@ public class BlockType extends ItemType {
 
         if (getMaterial().isSolid()) {
             BlockBoundingBox box = new BlockBoundingBox(target);
-            BoundingBox searchBox = BoundingBox.copyOf(box);
-            Vector vec = new Vector(32, 32, 32);
-            searchBox.minCorner.subtract(vec);
-            searchBox.maxCorner.add(vec);
-            List<Entity> entities = target.getWorld().getEntityManager().getEntitiesInside(searchBox, null);
+            List<Entity> entities = target.getWorld().getEntityManager().getEntitiesInside(box, null);
             for (Entity e : entities) {
-                GlowEntity entity = (GlowEntity) e;
-                if (entity instanceof LivingEntity && entity.intersects(box)) {
+                if (e instanceof LivingEntity) {
                     return;
                 }
             }
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -11,6 +11,7 @@ import net.glowstone.block.itemtype.ItemType;
 import net.glowstone.entity.GlowEntity;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.entity.physics.BlockBoundingBox;
+import net.glowstone.entity.physics.BoundingBox;
 import net.glowstone.util.SoundInfo;
 import net.glowstone.util.SoundUtil;
 import org.bukkit.*;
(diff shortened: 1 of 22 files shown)


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -303,11 +305,28 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
             return;
         }
 
-        // TODO: check bounding box
         if (getMaterial().isSolid()) {
-            for (Entity entity : against.getChunk().getEntities()) {
-                if (entity instanceof LivingEntity && entity.getLocation().getBlockX() == target.getX() && entity.getLocation().getBlockY() == target.getY() && entity.getLocation().getBlockZ() == target.getZ()) {
-                    return;
+            GlowChunk[] chunks = new GlowChunk[]{
```
mastercoms: What if the entity is really big? Copy `getNearbyEntities()` in `GlowEntity` to `BoundingBox`, and then use it here instead so the entity search accounts for bounding box, rather than location.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -303,11 +305,28 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
...
+                    target.getWorld().getChunkAt(target.getChunk().getX() - 1, target.getChunk().getZ() - 1),
+                    target.getWorld().getChunkAt(target.getChunk().getX() - 1, target.getChunk().getZ() + 1),
+                    target.getWorld().getChunkAt(target.getChunk().getX(), target.getChunk().getZ() - 1),
+                    target.getWorld().getChunkAt(target.getChunk().getX() + 1, target.getChunk().getZ() - 1),
+            };
+            BlockBoundingBox box = new BlockBoundingBox(target);
+            for (GlowChunk chunk : chunks) {
+                for (Entity e : chunk.getEntities()) {
+                    GlowEntity entity = (GlowEntity) e;
+                    if (!(entity instanceof LivingEntity)) {
```
mastercoms: Could you merge the conditions together? `entity instanceof LivingEntity && entity.intersects(box)`

src/main/java/net/glowstone/block/blocktype/BlockType.java
aramperes: It would really surprise me if an entity takes 3 chunks in a single direction. The largest entity (x/z) is  the Ender dragon (16). In any case, an entity will never have a colliding box that goes further than this.

src/main/java/net/glowstone/block/blocktype/BlockType.java
aramperes: Not really because the `return;` statement says that the block can't be placed (if an entity intersects the block), and you can place blocks on non-living entities (which is why there's a `continue;` statement.)

mastercoms: But if you merge the conditions together then it will still not return on non LivingEntites.

src/main/java/net/glowstone/block/blocktype/BlockType.java
mastercoms: `getNearbyEntities()` simplifies the process as well, as you only need to loop through one for-each, and you only need to check if the entities in the list are an instance of `LivingEntity`.

src/main/java/net/glowstone/block/blocktype/BlockType.java
aramperes: Oh right, I didn't get what you mean. My bad

mastercoms (changes requested):
idk what to put here but there are things that i requested and i am telling github that i did this

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -306,27 +307,16 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
...
-                        return;
-                    }
+            BoundingBox searchBox = BoundingBox.copyOf(box);
+            Vector vec = new Vector(32, 32, 32);
+            searchBox.minCorner.subtract(vec);
+            searchBox.maxCorner.add(vec);
+            List<Entity> entities = target.getWorld().getEntityManager().getEntitiesInside(searchBox, null);
+            for (Entity e : entities) {
+                GlowEntity entity = (GlowEntity) e;
+                if (entity instanceof LivingEntity && entity.intersects(box)) {
```
mastercoms: `entity.intersects(box)` is repetitive because `getEntitiesInside()` already checks this.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -303,10 +306,16 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
             return;
         }
 
-        // TODO: check bounding box
         if (getMaterial().isSolid()) {
-            for (Entity entity : against.getChunk().getEntities()) {
-                if (entity instanceof LivingEntity && entity.getLocation().getBlockX() == target.getX() && entity.getLocation().getBlockY() == target.getY() && entity.getLocation().getBlockZ() == target.getZ()) {
+            BlockBoundingBox box = new BlockBoundingBox(target);
+            BoundingBox searchBox = BoundingBox.copyOf(box);
+            Vector vec = new Vector(32, 32, 32);
```
mastercoms: Wait so here are you expanding the bounding box? Why not just keep it the same size?

mastercoms (approved):

mastercoms:
I hope you get better! Thanks for your contribution!

But before I merge,

```
[ERROR] src/main/java/net/glowstone/block/blocktype/BlockType.java[11:8] (imports) UnusedImports: Unused import - net.glowstone.entity.GlowEntity.
```

PR #374, opened 2016-09-26
Implement entity features

aramperes:
Among other things, this aims to implement some missing entity features.
- Shulkers
  - Make them spawnable
  - Save/Load NBT data
  - Send entity metadata values
  - Implement `POSITION` and `OPTPOSITION` data types

---
- Iron Golem & Snowman
  - Make them spawnable via their statue

---
- Wither
  - Make them spawnable via `/summon WitherBoss`
  - Invulnerability state + spawning effect, health regeneration
  - Spawning statue

---
- Guardian
  - Make the Elder guardian state visible via metadata.

---
- Minecarts
  - Implement Rideable, Storage, Furnace, Spawner, Hopper and Explosive minecarts
  - Store inventory for storage & hopper minecarts

---
- Misc
  - Fix `/summon` command from recent changes to spawning API
  - Add tab-completion to `/summon` command
  - Smoothen numerical type compatibility for NBT (e.g. `{Elder:1}` can be accessed as `tag.getByte("Elder");`
  - New _BlockPattern_ "API" - ease-of-use for matching block patterns like entity statues

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1225,6 +1225,8 @@ public final class GlowWorld implements World {
                     if (!spawnEvent.isCancelled()) {
                         List<Message> spawnMessage = entity.createSpawnMessage();
                         getRawPlayers().stream().filter(player -> player.canSeeEntity(impl)).forEach(player -> player.getSession().sendAll(spawnMessage.toArray(new Message[spawnMessage.size()])));
+                    } else {
+                        entity.remove();
                     }
                 }
             } catch (NoSuchMethodException e) {
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -30,6 +30,7 @@ import org.bukkit.block.BlockState;
 import org.bukkit.entity.*;
 import org.bukkit.event.entity.CreatureSpawnEvent;
 import org.bukkit.event.entity.CreatureSpawnEvent.SpawnReason;
+import org.bukkit.event.entity.EntitySpawnEvent;
 import org.bukkit.event.weather.LightningStrikeEvent;
 import org.bukkit.event.weather.ThunderChangeEvent;
 import org.bukkit.event.weather.WeatherChangeEvent;
@@ -1209,7 +1210,7 @@ public final class GlowWorld implements World {
                 Constructor<? extends GlowEntity> constructor = clazz.getConstructor(Location.class);
                 entity = constructor.newInstance(location);
                 if (entity instanceof LivingEntity) {
-                    CreatureSpawnEvent spawnEvent = new CreatureSpawnEvent((LivingEntity) entity, reason);
+                    CreatureSpawnEvent spawnEvent = EventFactory.callEvent(new CreatureSpawnEvent((LivingEntity) entity, reason);
                     if (!spawnEvent.isCancelled()) {
                         entity.createSpawnMessage();
                     } else {
@@ -1217,7 +1218,9 @@ public final class GlowWorld implements World {
                         entity.remove();
                     }
                 } else {
-                    entity.createSpawnMessage();
+                    EntitySpawnEvent spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
+                    if (!spawnEvent.isCancelled())
+                        entity.createSpawnMessage();
                 }
             } catch (NoSuchMethodException e) {
                 GlowServer.logger.log(Level.WARNING, "Invalid entity spawn: ", e);
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1211,23 +1211,18 @@ public final class GlowWorld implements World {
                 Constructor<? extends GlowEntity> constructor = clazz.getConstructor(Location.class);
                 entity = constructor.newInstance(location);
                 GlowEntity impl = entity;
+                EntitySpawnEvent spawnEvent;
                 if (entity instanceof LivingEntity) {
-                    CreatureSpawnEvent spawnEvent = EventFactory.callEvent(new CreatureSpawnEvent((LivingEntity) entity, reason));
-                    if (!spawnEvent.isCancelled()) {
-                        List<Message> spawnMessage = entity.createSpawnMessage();
-                        getRawPlayers().stream().filter(player -> player.canSeeEntity(impl)).forEach(player -> player.getSession().sendAll(spawnMessage.toArray(new Message[spawnMessage.size()])));
-                    } else {
-                        // TODO: separate spawning and construction for better event cancellation
-                        entity.remove();
-                    }
+                    spawnEvent = EventFactory.callEvent(new CreatureSpawnEvent((LivingEntity) entity, reason));
                 } else {
-                    EntitySpawnEvent spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
-                    if (!spawnEvent.isCancelled()) {
-                        List<Message> spawnMessage = entity.createSpawnMessage();
-                        getRawPlayers().stream().filter(player -> player.canSeeEntity(impl)).forEach(player -> player.getSession().sendAll(spawnMessage.toArray(new Message[spawnMessage.size()])));
-                    } else {
-                        entity.remove();
-                    }
+                    spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
+                }
+                if (!spawnEvent.isCancelled()) {
+                    List<Message> spawnMessage = entity.createSpawnMessage();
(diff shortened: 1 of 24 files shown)


aramperes:
I'm putting this under review because I think it reached the point which I aimed for. If anyone would like to see something more added, please let me know.

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1208,12 +1208,16 @@ public GlowEntity spawn(Location location, Class<? extends GlowEntity> clazz, Sp
             try {
                 Constructor<? extends GlowEntity> constructor = clazz.getConstructor(Location.class);
                 entity = constructor.newInstance(location);
-                CreatureSpawnEvent spawnEvent = new CreatureSpawnEvent((LivingEntity) entity, reason);
-                if (!spawnEvent.isCancelled()) {
-                    entity.createSpawnMessage();
+                if (entity instanceof LivingEntity) {
+                    CreatureSpawnEvent spawnEvent = new CreatureSpawnEvent((LivingEntity) entity, reason);
+                    if (!spawnEvent.isCancelled()) {
+                        entity.createSpawnMessage();
```
mastercoms: I know this was here before, but I'm pretty sure that the spawn message needs to be sent to players that can see the chunk the entity is in.

src/main/java/net/glowstone/entity/meta/MetadataType.java
```diff
@@ -19,8 +19,8 @@
     ITEM(ItemStack.class),
     BOOLEAN(Boolean.class),
     VECTOR(EulerAngle.class),
-    POSITION(Position.class),
```
mastercoms, line 22: What is the reason for this change?

aramperes: Well I'm not really sure why `Position.class` was used here, that class is a utility and is not initializable. Location on the other hand can hold position data, so I think this is the most appropriate type.

[This](https://github.com/GlowstoneMC/Glowstone/pull/374/files/0444a35caece2595b33296ab2faa0d8c55e2a8a7#diff-270aed41c5e1297cb3ddafcfdf4efeebR150) is where the Location data is actually transformed into the correct [format](http://wiki.vg/Protocol#Position).

src/main/java/net/glowstone/GlowWorld.java
aramperes: Hmm yes I will look into that, I think it would be similar to [this](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/entity/TileEntity.java#L43-L46) for TEs?

src/main/java/net/glowstone/entity/meta/MetadataType.java
mastercoms: Would it be appropriate to move the transformation to the Position class?

src/main/java/net/glowstone/GlowWorld.java
mastercoms: Yes

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1208,12 +1210,22 @@ public GlowEntity spawn(Location location, Class<? extends GlowEntity> clazz, Sp
...
                 Constructor<? extends GlowEntity> constructor = clazz.getConstructor(Location.class);
                 entity = constructor.newInstance(location);
-                CreatureSpawnEvent spawnEvent = new CreatureSpawnEvent((LivingEntity) entity, reason);
-                if (!spawnEvent.isCancelled()) {
-                    entity.createSpawnMessage();
+                GlowEntity impl = entity;
+                if (entity instanceof LivingEntity) {
+                    CreatureSpawnEvent spawnEvent = EventFactory.callEvent(new CreatureSpawnEvent((LivingEntity) entity, reason));
+                    if (!spawnEvent.isCancelled()) {
+                        List<Message> spawnMessage = entity.createSpawnMessage();
```
mastercoms: Why use `List`, instead of an implementing class like `ArrayList`?

src/main/java/net/glowstone/command/SummonCommand.java
```diff
@@ -83,40 +89,43 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
+    @Override
+    public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
+        Validate.notNull(sender, "Sender cannot be null", new Object[0]);
+        Validate.notNull(args, "Arguments cannot be null", new Object[0]);
+        Validate.notNull(alias, "Alias cannot be null", new Object[0]);
+        if (args.length == 1) {
+            String arg = args[0];
+            ArrayList<String> completion = new ArrayList<>();
+            for (EntityType type : EntityType.values()) {
+                if (!checkSummon(null, type)) {
```
mastercoms: Combine if statement?

src/main/java/net/glowstone/block/blocktype/BlockPumpkin.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private boolean spawnIronGolem(Location location) {
+        if (IRONGOLEM_PATTERN.matches(location, true, 1, 0)) {
+            location.getWorld().spawnEntity(location.clone().subtract(-0.5, 2, -0.5), EntityType.IRON_GOLEM);
+            return true;
+        }
+        return false;
+    }
+
+    private boolean spawnSnowman(Location location) {
```
mastercoms: What is the boolean here used for?

aramperes: I copied the `spawnIronGolem` method and left it there, so it had no usage for snowmen (now removed)

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1208,12 +1210,22 @@ public GlowEntity spawn(Location location, Class<? extends GlowEntity> clazz, Sp
...
+                        entity.remove();
+                    }
                 } else {
-                    // TODO: separate spawning and construction for better event cancellation
-                    entity.remove();
+                    EntitySpawnEvent spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
+                    if (!spawnEvent.isCancelled()) {
+                        List<Message> spawnMessage = entity.createSpawnMessage();
+                        getRawPlayers().stream().filter(player -> player.canSeeEntity(impl)).forEach(player -> player.getSession().sendAll(spawnMessage.toArray(new Message[spawnMessage.size()])));
+                    }
```
mastercoms: Why is there not an `entity.remove()` else branch?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1208,12 +1210,24 @@ public GlowEntity spawn(Location location, Class<? extends GlowEntity> clazz, Sp
...
+                        getRawPlayers().stream().filter(player -> player.canSeeEntity(impl)).forEach(player -> player.getSession().sendAll(spawnMessage.toArray(new Message[spawnMessage.size()])));
+                    } else {
+                        // TODO: separate spawning and construction for better event cancellation
+                        entity.remove();
+                    }
                 } else {
-                    // TODO: separate spawning and construction for better event cancellation
-                    entity.remove();
+                    EntitySpawnEvent spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
+                    if (!spawnEvent.isCancelled()) {
```
mastercoms: Can the cancel check be merged from both events?

aramperes: Would you like a ternary,

``` java
EntitySpawnEvent spawnEvent = EventFactory.callEvent(entity instanceof LivingEntity ? new CreatureSpawnEvent((LivingEntity) entity, reason) : new EntitySpawnEvent(entity));
```

Or an if-else block,

``` java
EntitySpawnEvent spawnEvent;
if (entity instanceof LivingEntity) {
    spawnEvent = EventFactory.callEvent(new CreatureSpawnEvent((LivingEntity) entity, reason));
} else {
    spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
}
```

picknchew: An if else block looks much more readable in my opinion.

src/main/java/net/glowstone/net/GlowBufUtils.java
```diff
@@ -144,11 +142,15 @@ public static void writeMetadata(ByteBuf buf, List<Entry> entries) throws IOExce
                     buf.writeFloat((float) Math.toDegrees(angle.getZ()));
                     break;
                 case POSITION:
+                    {
```
mastercoms: What is the purpose of these braces?

aramperes: These two variable names clash together because the case block does not isolate the variable.

``` java
case POSITION:
    BlockVector vector = (BlockVector) value; // Clashes
    buf.writeLong(Position.getPosition(vector));
    break;
case OPTPOSITION:
    if (value != null) {
        BlockVector vector = (BlockVector) value; // Clashes
        buf.writeLong(Position.getPosition(vector));
    }
    break;
```

An alternative solution would be to give a different name... I'm not sure what's the best solution.

PR #376, opened 2016-10-24
Fixed dungeons spawning process

VaiTon:
Fixed bugs:
- Chest spawning next to the spawner or next to each other
- Cobblestone ceiling not created

### DIFF
--- a/src/main/java/net/glowstone/generator/structures/GlowDungeon.java
+++ b/src/main/java/net/glowstone/generator/structures/GlowDungeon.java
@@ -141,19 +141,19 @@ public class GlowDungeon extends GlowStructurePiece {
                 if (builder.getBlockState(new Vector(x, 1, z)).getType() == Material.AIR) {
                     BlockFace face = null;
                     int solidBlocksCount = 0;
-                    if (builder.getBlockState(new Vector(x - 1, 1, z)).getType().equals(Material.COBBLESTONE)) {
+                    if (builder.getBlockState(new Vector(x - 1, 1, z)).getType() == Material.COBBLESTONE) {
                         solidBlocksCount++;
                         face = BlockFace.EAST;
                     }
-                    if (builder.getBlockState(new Vector(x + 1, 1, z)).getType().equals(Material.COBBLESTONE)) {
+                    if (builder.getBlockState(new Vector(x + 1, 1, z)).getType() == Material.COBBLESTONE) {
                         solidBlocksCount++;
                         face = BlockFace.WEST;
                     }
-                    if (builder.getBlockState(new Vector(x, 1, z - 1)).getType().equals(Material.COBBLESTONE)) {
+                    if (builder.getBlockState(new Vector(x, 1, z - 1)).getType() == Material.COBBLESTONE) {
                         solidBlocksCount++;
                         face = BlockFace.SOUTH;
                     }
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/generator/structures/GlowDungeon.java
```diff
@@ -141,19 +141,19 @@ public boolean generate(World world, Random random, StructureBoundingBox genBoun
                 if (builder.getBlockState(new Vector(x, 1, z)).getType() == Material.AIR) {
                     BlockFace face = null;
                     int solidBlocksCount = 0;
-                    if (builder.getBlockState(new Vector(x - 1, 1, z)).getType().isSolid()) {
+                    if (builder.getBlockState(new Vector(x - 1, 1, z)).getType().equals(Material.COBBLESTONE)) {
```
mastercoms: does it not also use mossy cobblestone?

src/main/java/net/glowstone/generator/structures/GlowDungeon.java
```diff
@@ -96,9 +96,9 @@ public boolean generate(World world, Random random, StructureBoundingBox genBoun
 
         for (int x = 0; x < sizeX; x++) {
             for (int z = 0; z < sizeZ; z++) {
-                for (int y = HEIGHT - 2; y >= 0; y--) {
+                for (int y = HEIGHT - 1; y >= 0; y--) {
                     BlockState state = builder.getBlockState(new Vector(x, y, z));
-                    if (y > 0 && x > 0 && z > 0 && x < sizeX - 1 && z < sizeZ - 1) {
+                    if (y > 0 && x > 0 && z > 0 && x < sizeX - 1 && y < HEIGHT - 2 && z < sizeZ - 1) {
```
mastercoms: since this is putting air in the room, shouldn't the bounds for y be **1** to HEIGHT - 1?

mastercoms:
Could this PR also get rid of the dungeon generation debug message:

``` java
GlowServer.logger.info("dungeon generated: " + loc.getBlockX() + "," + loc.getBlockY() + "," + loc.getBlockZ());
```

src/main/java/net/glowstone/generator/structures/GlowDungeon.java
VaiTon: No, because to six there are seven iteration, to 6-1 there are six iteration, to 6-2 there are 4 iterations. EDIT: My fault, that's true

src/main/java/net/glowstone/generator/structures/GlowDungeon.java
VaiTon: Not on the walls

VaiTon:
Should I eliminate the message or set the level as DEBUG?

Paulomart:
@VaiTon86 would set it do debug

aramperes (approved):

src/main/java/net/glowstone/generator/structures/GlowDungeon.java
mastercoms: I think using `==` on the Material enum (`getType()` and `Material.COBBLESTONE`) would be better than using the equals() method, which simply just returns `this==other`.

mastercoms (approved):

PR #377, opened 2016-10-30
RegionFile improvements (memory and CPU)

minecrafter:
This should improve memory usage, CPU and performance, especially on servers that load and save a lot of chunks.

The changes (edited late on 10/30/16 to reflect new changes):
- `sectorFree` was changed from an `ArrayList<Boolean>` to a `BitSet`. This is where most of the memory savings comes from (replacing a collection of boxed primitives with a primitive array). The effect is most pronounced with more heavily populated region files - smaller regions will not see as much of an impact.
- `GZIPInputStream`, `InflaterInputStream` and `DeflaterOutputStream` should be buffered. This should increase performance, specifically for chunk loading and saving.
- `ChunkBuffer`'s buffer size was decreased to 4096 bytes, as most chunks are one sector in size.

Here's one instance from a heavily populated region (for this chunk, this was an approximately 15% decrease in memory usage). This is from a memory capture done soon after the server begins listening for connections (no connections were made).

Before:

screenshot from 2016-10-30 20-30-45

After:

screenshot from 2016-10-30 20-30-57

### DIFF
--- a/src/main/java/net/glowstone/io/anvil/RegionFile.java
+++ b/src/main/java/net/glowstone/io/anvil/RegionFile.java
@@ -65,6 +65,8 @@ package net.glowstone.io.anvil;
 import net.glowstone.GlowServer;
 
 import java.io.*;
+import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
 import java.util.BitSet;
 import java.util.zip.DeflaterOutputStream;
 import java.util.zip.GZIPInputStream;
@@ -138,10 +140,20 @@ public class RegionFile {
         sectorsUsed.set(0, true);
         sectorsUsed.set(1, true);
 
-        // read offsets from offset table
+        // read offset table and timestamp tables
         file.seek(0);
+        ByteBuffer header = ByteBuffer.allocate(8192);
+        while (header.hasRemaining()) {
+            if (file.getChannel().read(header) == -1) {
+                throw new EOFException();
+            }
+        }
+        header.clear();
+
+        // populate the tables
+        IntBuffer headerAsInts = header.asIntBuffer();
         for (int i = 0; i < SECTOR_INTS; ++i) {
-            int offset = file.readInt();
+            int offset = headerAsInts.get();
             offsets[i] = offset;
 
             int startSector = offset >> 8;
@@ -157,7 +169,7 @@ public class RegionFile {
         }
         // read timestamps from timestamp table
         for (int i = 0; i < SECTOR_INTS; ++i) {
-            chunkTimestamps[i] = file.readInt();
+            chunkTimestamps[i] = headerAsInts.get();
         }
     }
 
--- a/src/main/java/net/glowstone/io/anvil/RegionFile.java
+++ b/src/main/java/net/glowstone/io/anvil/RegionFile.java
@@ -200,11 +200,11 @@ public class RegionFile {
         if (version == VERSION_GZIP) {
             byte[] data = new byte[length - 1];
             file.read(data);
-            return new DataInputStream(new GZIPInputStream(new ByteArrayInputStream(data), 2 * SECTOR_BYTES));
+            return new DataInputStream(new BufferedInputStream(new GZIPInputStream(new ByteArrayInputStream(data))));
         } else if (version == VERSION_DEFLATE) {
             byte[] data = new byte[length - 1];
             file.read(data);
-            return new DataInputStream(new InflaterInputStream(new ByteArrayInputStream(data), new Inflater(), 2 * SECTOR_BYTES));
+            return new DataInputStream(new BufferedInputStream(new InflaterInputStream(new ByteArrayInputStream(data))));
         }
 
         throw new IOException("Unknown version: " + version);
@@ -212,7 +212,7 @@ public class RegionFile {
 
     public DataOutputStream getChunkDataOutputStream(int x, int z) {
         checkBounds(x, z);
-        return new DataOutputStream(new DeflaterOutputStream(new ChunkBuffer(x, z), new Deflater(Deflater.BEST_SPEED)));
+        return new DataOutputStream(new BufferedOutputStream(new DeflaterOutputStream(new ChunkBuffer(x, z))));
     }
 
     /* write a chunk at (x,z) with length bytes of data to disk */
@@ -335,7 +335,7 @@ public class RegionFile {
         private final int x, z;
 
         public ChunkBuffer(int x, int z) {
-            super(2 * SECTOR_BYTES); // initialize to 8KB
+            super(SECTOR_BYTES); // initialize to 4KB
             this.x = x;
             this.z = z;
         }
--- a/src/main/java/net/glowstone/io/anvil/RegionFile.java
+++ b/src/main/java/net/glowstone/io/anvil/RegionFile.java
@@ -65,11 +65,8 @@ package net.glowstone.io.anvil;
 import net.glowstone.GlowServer;
 
 import java.io.*;
-import java.util.ArrayList;
-import java.util.zip.Deflater;
-import java.util.zip.DeflaterOutputStream;
-import java.util.zip.GZIPInputStream;
-import java.util.zip.InflaterInputStream;
+import java.util.Arrays;
+import java.util.zip.*;
 
 public class RegionFile {
 
@@ -80,11 +77,13 @@ public class RegionFile {
     private static final int SECTOR_INTS = SECTOR_BYTES / 4;
 
     private static final int CHUNK_HEADER_SIZE = 5;
+    private static final int COMPRESSED_BUFFER_SIZE = 8192;
+
     private static final byte[] emptySector = new byte[SECTOR_BYTES];
     private final int[] offsets;
     private final int[] chunkTimestamps;
     private RandomAccessFile file;
-    private ArrayList<Boolean> sectorFree;
+    private boolean[] sectorFree;
     private int sizeDelta;
     private long lastModified;
 
@@ -128,13 +127,11 @@ public class RegionFile {
 
         // set up the available sector map
         int nSectors = (int) (file.length() / SECTOR_BYTES);
-        sectorFree = new ArrayList<>(nSectors);
-        for (int i = 0; i < nSectors; ++i) {
-            sectorFree.add(true);
-        }
+        sectorFree = new boolean[nSectors];
+        Arrays.fill(sectorFree, true);
 
-        sectorFree.set(0, false); // chunk offset table
-        sectorFree.set(1, false); // for the last modified info
+        sectorFree[0] = false; // chunk offset table
+        sectorFree[1] = false; // for the last modified info
 
         // read offsets from offset table
         file.seek(0);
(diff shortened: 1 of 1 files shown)


Paulomart (approved):

src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -333,7 +336,7 @@ public void close() throws IOException {
         private final int x, z;
 
         public ChunkBuffer(int x, int z) {
-            super(8096); // initialize to 8KB
+            super(8192); // initialize to 8KB
```
ghost: How about using constant `COMPRESSED_BUFFER_SIZE` here?

src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -80,11 +77,13 @@ would be in region (0, -1), and one at (70, -30) would be at (3, -1).
     private static final int SECTOR_INTS = SECTOR_BYTES / 4;
 
     private static final int CHUNK_HEADER_SIZE = 5;
+    private static final int COMPRESSED_BUFFER_SIZE = 8192;
```
mastercoms: This is just `SECTOR_BYTES * 2`. Is it really worth making a new constant? `SECTOR_BYTES * 2` is used throughout the code already. If you want to add this constant, name it something less specific, and use it everywhere.

mastercoms:
I noticed some for loops in the constructors which I think can be replaced with writing empty sectors instead. Could you change those for loops?

minecrafter:
@mastercoms The constructor handles edge cases such as misaligned boundaries where it is not easy to convert the code to writing only empty sectors. However, there is one instance writing empty sectors is acceptable (new region file), which I have implemented.

ghost:
Are virtual machine enough smart to caluclate SECTOR_SIZE \* 2 as constant to not waste cpu ticks to multiple every time? If no, it's better use another const.

minecrafter:
@ensirius You're verging on premature optimization territory.

I doubt this is worth a constant. I'm sure that `javac` would inline this for us anyway.

olliestanley:
IMO a constant is worth it, makes it more clear what it is. But SECTOR_SIZE \* 2 is also easy to understand so debatable I suppose

minecrafter:
After a few hours of experimentation, I now have a version that uses `BitSet` instead of an array of booleans. With this change, most `RegionFile` instances now use about the same amount of memory, give or take a few hundred bytes.

minecrafter:
I've now added proper buffering for the streams involved in compressing chunk data, which should improve performance for chunk loading and saving.

Here is a new comparison with a fresh world under the same conditions:

Before (original code):

screenshot from 2016-10-30 20-30-45

After (BitSet):

screenshot from 2016-10-30 20-30-57

I'd like to find some way to quantify chunk loading and saving performance.

src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -200,19 +214,19 @@ public DataInputStream getChunkDataInputStream(int x, int z) throws IOException
         if (version == VERSION_GZIP) {
             byte[] data = new byte[length - 1];
             file.read(data);
-            return new DataInputStream(new GZIPInputStream(new ByteArrayInputStream(data)));
+            return new DataInputStream(new BufferedInputStream(new GZIPInputStream(new ByteArrayInputStream(data))));
```
mastercoms: Isn't `GZIPInputStream` [already buffered](https://docs.oracle.com/javase/8/docs/api/java/util/zip/GZIPInputStream.html#GZIPInputStream-java.io.InputStream-int-)? 

I think a better optimization would be to investigate more performant binary serialization methods, like [`ByteBuffer`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html), or if you're feeling to have a little byte of fun, [`Unsafe`](https://mechanical-sympathy.blogspot.com/2012/07/native-cc-like-performance-for-java.html), though I would start with [`Kryo`](https://github.com/EsotericSoftware/kryo) for an established Unsafe-based IO/serializer before deciding to make your own or not.

src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -200,19 +214,19 @@ public DataInputStream getChunkDataInputStream(int x, int z) throws IOException
...
+            return new DataInputStream(new BufferedInputStream(new InflaterInputStream(new ByteArrayInputStream(data))));
         }
 
         throw new IOException("Unknown version: " + version);
     }
 
     public DataOutputStream getChunkDataOutputStream(int x, int z) {
         checkBounds(x, z);
-        return new DataOutputStream(new DeflaterOutputStream(new ChunkBuffer(x, z), new Deflater(Deflater.BEST_SPEED)));
+        return new DataOutputStream(new BufferedOutputStream(new DeflaterOutputStream(new ChunkBuffer(x, z))));
```
mastercoms: https://docs.oracle.com/javase/8/docs/api/java/util/zip/DeflaterOutputStream.html#DeflaterOutputStream-java.io.OutputStream-java.util.zip.Deflater-int-boolean-

src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -200,19 +214,19 @@ public DataInputStream getChunkDataInputStream(int x, int z) throws IOException
         if (version == VERSION_GZIP) {
             byte[] data = new byte[length - 1];
             file.read(data);
-            return new DataInputStream(new GZIPInputStream(new ByteArrayInputStream(data)));
+            return new DataInputStream(new BufferedInputStream(new GZIPInputStream(new ByteArrayInputStream(data))));
         } else if (version == VERSION_DEFLATE) {
             byte[] data = new byte[length - 1];
             file.read(data);
-            return new DataInputStream(new InflaterInputStream(new ByteArrayInputStream(data)));
+            return new DataInputStream(new BufferedInputStream(new InflaterInputStream(new ByteArrayInputStream(data))));
```
mastercoms: https://docs.oracle.com/javase/8/docs/api/java/util/zip/InflaterInputStream.html#InflaterInputStream-java.io.InputStream-java.util.zip.Inflater-int-

src/main/java/net/glowstone/io/anvil/RegionFile.java
minecrafter: `InflaterInputStream` and `GZIPInputStream` do have a buffer for reading from an input stream. However, they are used for different purposes: `InflaterInputStream` uses it to ensure that data is read from the backing `InputStream`, whereas the `BufferedInputStream` is meant to buffer the decompressed data. As a side effect, this reduces the number of inflations required to decompress the chunk.

src/main/java/net/glowstone/io/anvil/RegionFile.java
minecrafter: https://github.com/GlowstoneMC/Glowstone/pull/377#discussion_r85679496

src/main/java/net/glowstone/io/anvil/RegionFile.java
minecrafter: The buffer in `DeflateOutputStream` is used as a temporary working area that is eventually written to the `OutputStream`. Putting a `BufferedOutputStream` allows the writes to the stream to be batched so that the number of deflations is reduced.

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: But GZIPInputStream is the bottleneck in this case. The speed of using the compressed data is what is ultimately limiting the speed, not subsequently using the uncompressed data. There might be a small increase in performance by buffering the uncompressed data, but will it be worth it?

Because keep in mind that having a buffer for the compressed data is buffering more information in less bytes than the uncompressed stream.

I think it's better to tune the included buffers rather than adding a new `BufferedInputStream` on top of it, especially when you have its buffer for the uncompressed data much larger than than the inside buffers for the compressed data.

A benchmark will likely be required to investigate if we should add a `BufferedInputStream`, but what's clear to me is that the buffers for the gzip, deflater, and inflater streams must be tuned.

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: So, then do both, so we don't do as many deflations, and so we don't use the disk I/O too much, too.

minecrafter: You're writing to a `ChunkBuffer`, which is a `ByteArrayOutputStream` that flushes its output on `close()`. I/O is not the problem. The problem is the amount of deflations, which `BufferedOutputStream` reduces.

src/main/java/net/glowstone/io/anvil/RegionFile.java
minecrafter: I don't think I was being clear enough. Let me rephrase what I said earlier.

`InflaterInputStream` includes a built-in buffer so it can read data from the backing input stream, which in our case is a `ByteArrayInputStream`. `InflaterInputStream.read()` is implemented by inflating this buffer into an user-provided byte array.

Personally, I think it's better CPU-wise to decompress a few large chunks of data at a time than decompress a lot of small chunks.

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: Well, I was more speaking generally for all the inner streams I commented on (deflater, gzip, and inflater) when I mentioned disk IO. Why not buffer the uncompressed stream, and tune the buffer size for the inner stream as well?

minecrafter: If tuning the inner buffer is warranted, then I can do that (2048 bytes looks to be a good general fit that I've observed).

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: Ok, I agree with you now this use has been explained. So we should have and tune both.

Have you looked into the different serializing methods? Or do you think it's not in the scope of this PR?

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: We should benchmark and then tune each buffer accordingly, starting from a base buffer size of 2048 and testing to see what increases performance under general server stress.

minecrafter: It's late for me right now, so I'm not going to benchmark at the moment. You should try [JMH](http://openjdk.java.net/projects/code-tools/jmh/) if you're going to be writing a benchmark.

src/main/java/net/glowstone/io/anvil/RegionFile.java
minecrafter: The scope of this PR is purely to improve `RegionFile`'s memory and CPU consumption and increase its performance, so changing serialization is off the table.

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: Ok, just add the buffer size of 2048 for now.

src/main/java/net/glowstone/io/anvil/RegionFile.java
mastercoms: Ok. Would you work on that serialization PR?

minecrafter: Probably not.

minecrafter:
@mastercoms I've made the changes you've requested.

mastercoms:
@minecrafter Yes I saw but I haven't had time to do a final review and then merge it.

minecrafter:
Perfect, let me know if more changes should be made. :+1:

src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -333,7 +347,7 @@ public void close() throws IOException {
         private final int x, z;
 
         public ChunkBuffer(int x, int z) {
-            super(8096); // initialize to 8KB
+            super(SECTOR_BYTES); // initialize to 4KB
```
mastercoms: Why was the buffer reduced? Are most chunks 4KB?

mastercoms (approved):

PR #378, opened 2016-10-31
Entity loot tables

aramperes:
This PR implements a new loot table format. You can check the forum post about it [here](https://forums.glowstone.net/topic/14/looking-for-feedback-entity-loot-tables).

This also integrates my OneLineReflection util, in order to support in-line method calling for loot tables.

- [x] JSON Loot tables for all mobs
- [x] Implement the loot table format
  - [x] Reflective values
  - [x] Min/max randomisation
  - [x] Conditional values (overriding values depending on the properties of the mob)
  - [x] Review loot tables for validity and correct Material names
  - [ ] Add more default conditions
  - [ ] Conditional experience

Misc
 - [x] Put entities on fire when attacked with Fire Aspect enchanted items
 - [x] Deal damage every second when on fire


src/main/java/net/glowstone/GlowServer.java
```diff
@@ -476,6 +477,13 @@ public void start() {
         nameBans.load();
         ipBans.load();
 
+        try {
+            LootingManager.load();
+        } catch (Exception e) {
```
mastercoms, line 482: Is it possible to provide a more useful error message? Maybe be more fine with the exception catching, or include the exception in the logger message.

aramperes: Something like this?

mastercoms:
Do you think that what is in this PR is already good to be merged?

I would rather have some basic working functionality for loot tables now rather than the full thing some time later. We can always add the missing features later.

aramperes:
@mastercoms Yeah sure I don't think I'll have the time to complete what is left here for now, feel free to review/merge.

aramperes:
Going to update this to 1.11 now.

aramperes:
This should be ready for review.

PR #379, opened 2016-11-01
Blocks

mastercoms:
A rewrite of a lot of things in net.glowstone.blocks

I removed the Sponge stuff already, so don't worry about that stuff getting in.

Very WIP. Lots of work to still be done. Two major things that are pending are getting blocks to be single instance per Location, and rewriting the blocktype package to allow for better inheritance and custom blocks.


src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -212,48 +211,101 @@ public void setType(Material type, byte data, boolean applyPhysics) {
...
+            getChunk().setType(x & 0xf, z & 0xf, y, type);
+            this.type = type;
+            changed = true;
+        }
+        if (oldData != data) {
+            getChunk().setMetaData(x & 0xf, z & 0xf, y, data);
+            changed = true;
+        }
 
-        if (oldTypeId == Material.DOUBLE_PLANT && getRelative(BlockFace.UP).getType() == Material.DOUBLE_PLANT) {
```
mastercoms, line 233: This needs to be moved to the blocktype stuff once that's rewritten.

mastercoms:
I am also probably going to redo the work from SpaceManiac, rewritten to use Bukkit instead of Sponge, with some other enhancements in light of changes in Glowstone and the recent changes to the blocks branch.

src/main/java/net/glowstone/block/MaterialValueManager.java
```diff
@@ -98,15 +423,29 @@ public GlowMaterialBuilder setRandomTicks(boolean randomTicks) {
...
+                return pistonMoveReaction;
+            }
+
+            public GlowMaterialBuilder setPistonMoveReaction(PistonMoveReaction pistonMoveReaction) {
+                this.pistonMoveReaction = pistonMoveReaction;
+                return this;
+            }
+        }
+
+        GlowMaterial() {
```
mastercoms, line 568: This is a temporary constructor until all values are filled in.

src/main/java/net/glowstone/block/GlowBlock.java
```diff
@@ -212,48 +211,101 @@ public void setType(Material type, byte data, boolean applyPhysics) {
...
+
     @Override
     public boolean setTypeId(int type) {
         return setTypeId(type, true);
     }
 
     @Override
     public boolean setTypeId(int type, boolean applyPhysics) {
-        return setTypeIdAndData(type, (byte) 0, applyPhysics);
+        return setTypeIdAndData(type, getData(), applyPhysics);
```
mastercoms: The spec only tells us to set the type, but should we reset the data as well?

aramperes: As to keep the same functionality as Craftbukkit/Spigot, it should get reset to the default data value of the new block

gdude2002:
This has been here for a while - it also has some conflicts. Do we need to keep this PR going?

mastercoms:
Yes, I will get around to updating it soon, after I get the new chat stuff working.

mastercoms:
Closing this as a new PR is coming soon.

PR #293, opened 2016-06-11
1.10? 1.10!

aramperes:
This is the update for Glowstone++ for support for 1.10 (1.10.2).

TODO:
- [x] Protocol changes
- [x] Remove compatibility with 1.9.X
- [x] Update to Glowkit for Paper API 1.10 (@mastercoms)
- [x] Implement 1.10 API update
- Misc changes:
  - [x] Add Magma to Nether generation
  - [x] New blocks crafting recipes
  - [x] Polar Bear
  - [x] Command Targets and Tilda notation (https://github.com/GlowstoneMC/Glowkit/pull/6)
  - [x] `/teleport` command (awaiting https://github.com/GlowstoneMC/Glowkit/pull/6)
  - [x] New NBT Tags: `FallFlying` and `ZombieType`
  - [x] Add Magma Block functionality
- Other changes:
  - Fixed off-hand bugs, added switching items in main/off hand (#313)
  - Fixed wrong armor slot mapping (#298)

PR #385, opened 2016-11-17
[WIP] Exploration update? Exploration update! (1.11)

aramperes:
Let's start updating Glowstone to 1.11:

- [x] Protocol changes
- [x] Metadata update
  - [x] Add new Metadata entries
  - [x] Implement new entries once API is updated and implemented
- [x] Bukkit, SpigotAPI and Paper update to 1.11
- [x] Deploy Glowkit 1.11
- [x] Implement new Entity classes
  - [x] Horse classes
  - [x] Skeleton classes
  - [x] Zombie classes
  - [x] Tweak entity storage
  - [x] New entities: Vex, Evoker, Vindicator
- [x] Misc changes for the update

-----------

Other stuff:
 - Update the following libraries:
   - `sqlite-jdbc` 3.15.1
   - `mysql-connector-java` 5.1.39
   - Remove `commons-io`
 - Ambient sounds for all living entities


mastercoms:
This pull request will not be accepted until you come up with an original title. #293

phase:
I can't wait to see `1.12? 1.12!`

mastercoms:
There are a few dependency updates that we need to do for Glowstone 1.11.
In the pom:
`jline` to  `2.12.1`
`maven-jar-plugin` to `3.0.2`
(and maybe look into updating `maven-compiler-plugin` to `3.6.0`)
In the [Library Manager](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/util/LibraryManager.java):
`mysql-connector-java` to `5.1.39`
`sqlite-jdbc` to `3.15.1`
Remove `commons-io`

pom.xml
```diff
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>jline</groupId>
             <artifactId>jline</artifactId>
-            <version>2.11</version>
+            <version>2.12.1</version>
```
mastercoms, line 80: I should mention that 2.12 had a bug with console input, might require a rewrite of the console :)

mastercoms: Just kidding, that was 2.13

mastercoms:
A few general things that you can try to do:

* Stop using deprecated methods where it is easy to do so
* Don't use null for ItemStacks any more

aramperes:
@mastercoms 
I don't really know where to find such cases, in all honesty :P There are so many, and idk how to find that in IntelliJ

mastercoms:
Deprecated usages are listed when compiling Glowstone.

I'm not sure what to do for ItemStack, but I would look in some inventory code to start.

src/main/java/net/glowstone/block/blocktype/BlockFlowerPot.java
```diff
@@ -43,12 +42,6 @@ public TileEntity createTileEntity(GlowChunk chunk, int cx, int cy, int cz) {
     @Override
     public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face, Vector clickedLoc) {
         GlowBlockState state = block.getState();
-        MaterialData data = state.getData();
-
-        if (!(data instanceof FlowerPot)) {
```
mastercoms, line 48: That's there for old worlds, I think we should hang on to this one.

mastercoms:
I think we've made all the changes that are _required_ to add 1.11 support. Of course, we still have much to do, but I think that can be done in subsequent PRs or commits. @momothereal please correct me if I'm wrong.

aramperes:
There are probably more tests to be done for the changes to entity storage and legacy compatibility, as well as the changes for AIRx0, but I think this is ok regarding features.

mastercoms:
Considering everything looks good and changes to be made have settled a bit, I can say that the 1.11 branch is ready to be merged into master.

PR #400, opened 2016-12-23
1.11.2 :(

mastercoms:
this is the update, 1.11.2, that mojang released for minecraft. it's a small update, except for the fact that the spigot devs finally remembered to actually do things, so some new methods were added and were possibly implemented in this pr, if I felt like implementing them.

thanks for reviewing this pr.


aramperes (commented):
yes

PR #402, opened 2016-12-24
Add ability to register custom entities

BSFishy:
This allows for people to get the EntityManager from GlowServer#getEntityManager or GlowWorld#getEntityManger, and register their own custom entities, remove those entities, or move those entities. Before, the methods were package-private, so no one was able to use those methods.

**WIP:** Not actually sure how to get this working. I am going to be fiddling around with a couple of different things, until I get it working, then I will post a thread on the forums.

### DIFF
--- a/src/main/java/net/glowstone/entity/EntityManager.java
+++ b/src/main/java/net/glowstone/entity/EntityManager.java
@@ -68,7 +68,7 @@ public final class EntityManager implements Iterable<GlowEntity> {
      * @param entity The entity.
      */
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


gdude2002:
Did you do a test compile? I can't imagine this causing any problems, but better safe than sorry

BSFishy:
It compiles and runs as normally.

mastercoms:
Looks good.

BSFishy:
I realized that not only the EntityManager needs changing, but other classes do too. I am going to be working on that until I can get it working in my own plugin.

PR #403, opened 2016-12-24
Allow for custom entities

BSFishy:
In the beginning, I was thinking only the EntityManager needed to be changed, but I have realized that it is not the only thing. This is a WIP until I can get it working in my own plugin, which I will then make a thread on the forums on how to register custom entities.


BSFishy:
I realized that I can just use GlowWorld#spawn to spawn the entities during the runtime. I thought I had to register them myself. (No longer WIP)

PR #404, opened 2016-12-24
Allow plugins to have custom entities

BSFishy:
This is a WIP to allow plugins to have custom entities. As of now, it is not quite straightforward and has errors with saving and what not. I will tell you when it works, and I have it working with my plugin.

### DIFF
--- a/src/main/java/net/glowstone/command/SummonCommand.java
+++ b/src/main/java/net/glowstone/command/SummonCommand.java
@@ -1,6 +1,7 @@
 package net.glowstone.command;
 
 import com.google.common.collect.ImmutableList;
+import net.glowstone.GlowWorld;
 import net.glowstone.entity.EntityRegistry;
 import net.glowstone.entity.GlowEntity;
 import net.glowstone.io.entity.EntityStorage;
@@ -18,6 +19,7 @@ import org.bukkit.command.ConsoleCommandSender;
 import org.bukkit.command.defaults.BukkitCommand;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.Player;
+import org.bukkit.event.entity.CreatureSpawnEvent;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -71,15 +73,16 @@ public class SummonCommand extends BukkitCommand {
         }
 
         String entityName = args[0];
-        EntityType type = EntityType.fromName(entityName);
-        if (type == null) {
-            sender.sendMessage(ChatColor.RED + "Unknown entity type: " + entityName);
+        if (!checkSummon(sender, entityName)) {
             return true;
         }
-        if (!checkSummon(sender, type)) {
-            return true;
+        GlowEntity entity;
+        if (EntityType.fromName(entityName) != null) {
+            entity = (GlowEntity) location.getWorld().spawnEntity(location, EntityType.fromName(entityName));
+        } else {
+            Class<? extends GlowEntity> clazz = EntityRegistry.getCustomEntityDescriptor(entityName).getEntityClass();
+            entity = ((GlowWorld) location.getWorld()).spawn(location, clazz, CreatureSpawnEvent.SpawnReason.CUSTOM);
         }
-        GlowEntity entity = (GlowEntity) location.getWorld().spawnEntity(location, type);
         if (tag != null) {
             EntityStorage.load(entity, tag);
         }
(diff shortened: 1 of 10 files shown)


LousyLynx:
I forget, where is that used?

aramperes:
https://github.com/GlowstoneMC/Glowstone/pull/404/files#diff-8769715641781a189890669ce58619a0R110

Say you have a custom entity that extends GlowPig, when you register a new store (e.g. CustomPigStore) it will override the default's GlowPig store, which means that all pigs on the server will be loaded with the custom data you are assigning. This might or might not be what is intended when registering new entities.

BSFishy:
Wouldn't we just technically use the idTable then? That would remove the need for the classTable, and allow plugin developers to make custom entities. Or are the other consequences that I don't know about?

BSFishy:
I know in the original NMS it allowed you to use your own enum. I suppose that we could make that happen and have a specific format, with specific methods. If the enum does not have those methods, it throws an error. That way the developer can program it any way they want, and it still works on the server despite the many ways to do it.

BSFishy:
That saves the type as a string rather than an EntityType itself. Also, there is a constructor that takes in the string. It is a string because that is the only thing from EntityType that it is using.

BSFishy:
That should be about it. My entity is registering and saving, and the other entities are working like they were before I started messing with stuff. It compiles and works properly.

aramperes:
I will test this when I have time today

aramperes:
Functionality seems to be alright. This is the code I used to spawn a custom sheep that is always red, and can become dank:

**CustomEntityPlugin**
```java
public class CustomEntityPlugin extends JavaPlugin implements Listener {
    @Override
    public void onLoad() {
        EntityStorage.bind(new CustomSheepStore()); // bind the store before loading the world
    }

    @Override
    public void onEnable() {
        getServer().getPluginManager().registerEvents(this, this);
    }

    @EventHandler
    public void onJoin(PlayerJoinEvent event) {
        GlowWorld world = (GlowWorld) event.getPlayer().getWorld();
        world.spawn(event.getPlayer().getLocation(), CustomSheep.class, CreatureSpawnEvent.SpawnReason.CUSTOM); // Needs to use method GlowWorld#spawn(location, <class extends GlowEntity>, spawnReason) to force the use of custom entities
    }
}
```

**CustomSheep**
```java
public class CustomSheep extends GlowSheep {
    private boolean dank;

    public CustomSheep(Location location) {
        super(location);
        setColor(DyeColor.RED);
    }

    public boolean isDank() {
        return dank;
    }

    public void setDank(boolean dank) {
        this.dank = dank;
        setColor(dank ? DyeColor.GREEN : DyeColor.RED);
    }

    @Override
        ItemStack holding = InventoryUtil.itemOrEmpty(player.getItemInHand());
        if (holding.getType() == Material.LONG_GRASS && holding.getDurability() == 2 && !isDank()) {
            setDank(true);
            world.spawnParticle(Particle.SMOKE_LARGE, getLocation(), 1);
            // consume
            holding.setAmount(holding.getAmount() - 1);
            if (holding.getAmount() == 0) holding = InventoryUtil.createEmptyStack();
            player.setItemInHand(holding);
            return true;
        }
        return false;
    }
}
```

**CustomSheepStore**

```java
public class CustomSheepStore extends AnimalStore<CustomSheep> {
    public CustomSheepStore() {
        super(CustomSheep.class, "custom_sheep");
    }

    @Override
    public void save(CustomSheep entity, CompoundTag tag) {
        super.save(entity, tag);
        if (entity.isDank()) {
            tag.putBool("Dank", true);
        }
    }

    @Override
    public void load(CustomSheep entity, CompoundTag tag) {
        super.load(entity, tag);
        if (tag.isByte("Dank")) {
            entity.setDank(tag.getBool("Dank"));
        }
    }
}
```

edit: added danks

mastercoms:
I think it would be better to have some sort of wrapper class you can use to extend entities. That way we don't have to use strings instead of the enum.

aramperes:
It should also be possible to override/register new loot data for custom entities. Maybe look into if you can work something out, unless you think this is out of the scope of this PR, then it can be added later on.

aramperes:
Some additions I suggest to your PR: https://github.com/LousyLynx/Glowstone/pull/1

src/main/java/net/glowstone/command/SummonCommand.java
```diff
@@ -88,19 +91,24 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
-            EntityRegistry.getEntity(type).getConstructor(Location.class);
-        } catch (NoSuchMethodException e) {
-            if (sender != null) sender.sendMessage(ChatColor.RED + "The entity '" + type.getName() + "' cannot be summoned.");
+        if (entityType != null && EntityRegistry.getEntity(entityType) == null) {
+            if (sender != null)
+                sender.sendMessage(ChatColor.RED + "The entity '" + entityType.getName() + "' is not implemented yet.");
+            return false;
+        } else if (entityType == null && !EntityRegistry.isCustomEntityRegistered(type)) {
+            if (sender != null)
+                sender.sendMessage(ChatColor.RED + "The entity '" + type + "' is does not exist.");
```
aramperes: "is does not exist" - typo, my bad.

aramperes (approved):
I'm ok with this, we can always adjust the API if we need to in the future.

mastercoms (approved):
I would prefer using the EntityType enum, but I think it would be convoluted to have it all work, so I'm fine with this.

aramperes:
Documentation relative to these additions: https://github.com/GlowstoneMC/Glowstone/wiki/Creating-custom-entities

PR #406, opened 2017-01-04
Chat components

mastercoms:
This uses md5's API for 1.8's new chat features, so we can fully support the JSON formatting introduced then.

The PR is not working yet, lots of networking stuff and tests need to be fixed.


mastercoms:
Closing in favor of 1.12-staging.

PR #409, opened 2017-01-06
Title fix UPDATE

BSFishy:
This is the initial title fix pr, but on a new branch to make it easier for me and the original Glowstone contributors.

This is a WIP to fix the title command, to accept proper JSON, do proper formatting and not throw any exceptions.

### DIFF
--- a/setup.bat
+++ b/setup.bat
@@ -1,0 +1,1 @@
+mvn package
--- a/setup.bat
+++ b/setup.bat
@@ -1,1 +1,0 @@
-mvn package
--- a/src/main/java/net/glowstone/command/TitleCommand.java
+++ b/src/main/java/net/glowstone/command/TitleCommand.java
@@ -1,7 +1,7 @@
 package net.glowstone.command;
 
-import com.destroystokyo.paper.Title;
 import net.glowstone.entity.GlowPlayer;
+import net.glowstone.net.message.play.game.TitleMessage;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
@@ -126,13 +126,60 @@ public class TitleCommand extends BukkitCommand {
                 component = convertJson((JSONObject) parsed);
             }
 
-            player.sendTitle(new Title(component));
+            ((GlowPlayer) player).updateTitle(TitleMessage.Action.TITLE, component);
+            ((GlowPlayer) player).sendTitle();
 
             sender.sendMessage("Updated " + player.getName() + "'s title");
         } else if (action.equalsIgnoreCase("subtitle")) {
-            sender.sendMessage("This feature is not yet supported");
+            if (args.length < 3) {
+                sender.sendMessage(ChatColor.RED + "Usage: /title <player> " + action + " <raw json>");
+                return false;
+            }
+
+            StringBuilder message = new StringBuilder();
+
+            for (int i = 2; i < args.length; i++) {
+                if (i > 2) message.append(" ");
+                message.append(args[i]);
+            }
+
+            String raw = message.toString().trim();
+            if (!validJson(raw)) {
+                sender.sendMessage(ChatColor.RED + "Invalid JSON: Could not parse, invalid format?");
+                return false;
+            }
+
+            String component = raw;
+            Object parsed = JSONValue.parse(raw);
+            if (parsed instanceof JSONObject) {
+                component = convertJson((JSONObject) parsed);
+            }
+
+            ((GlowPlayer) player).updateTitle(TitleMessage.Action.SUBTITLE, component);
+
+            sender.sendMessage("Updated " + player.getName() + "'s subtitle");
         } else if (action.equalsIgnoreCase("times")) {
-            sender.sendMessage("This feature is not yet supported");
+            if (args.length != 5) {
+                sender.sendMessage(ChatColor.RED + "Usage: /title <player> " + action + " <fade in> <stay time> <fade out>");
+                return false;
+            }
+
+            if (!tryParseInt(args[2])) {
+                sender.sendMessage(ChatColor.RED + "'" + args[2] + "' is not a number");
+                return false;
+            }
+            if (!tryParseInt(args[3])) {
+                sender.sendMessage(ChatColor.RED + "'" + args[3] + "' is not a number");
+                return false;
+            }
+            if (!tryParseInt(args[4])) {
+                sender.sendMessage(ChatColor.RED + "'" + args[4] + "' is not a number");
+                return false;
+            }
+
+            ((GlowPlayer) player).updateTitle(TitleMessage.Action.TIMES, toInt(args[2]), toInt(args[3]), toInt(args[4]));
+
+            sender.sendMessage("Updated " + player.getName() + "'s times");
         } else {
             sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
             return false;
@@ -141,18 +188,18 @@ public class TitleCommand extends BukkitCommand {
         return true;
     }
 
-    private int tryParseInt(CommandSender sender, String number, int lastParse) {
-        if (lastParse == -1) {
-            return -1;
-        }
-
-        int n = -1;
+    private boolean tryParseInt(String number) {
         try {
-            n = Integer.parseInt(number);
-        } catch (NumberFormatException e) {
-            sender.sendMessage(ChatColor.RED + "'" + number + "' is not a number");
+            Integer.parseInt(number);
+        } catch (NumberFormatException | NullPointerException e) {
+            return false;
         }
-        return n;
+        // only got here if we didn't return false
+        return true;
+    }
+
+    private int toInt(String number) {
+        return Integer.parseInt(number.trim());
     }
 
     private boolean validJson(String raw) {
--- a/src/main/java/net/glowstone/command/TitleCommand.java
+++ b/src/main/java/net/glowstone/command/TitleCommand.java
@@ -7,11 +7,12 @@ import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.defaults.BukkitCommand;
 import org.bukkit.entity.Player;
-import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 import org.json.simple.JSONValue;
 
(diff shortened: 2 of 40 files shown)


BSFishy:
I am having trouble coming up with how to send the title, subtitle, and times all at once. Would I save it in the `GlowPlayer`'s class? I also want to know what `GlowPlayer#updateTitle` is supposed to be. Is that what I would be using to be able to do the subtitle and times?

BSFishy:
I have figured out what I am going to do to make the title, subtitle, and times all be sent at once. I am going to use the `Title.Builder` class. That allows us to not have to create a new instance of a `Title` every time we want to update it. I am also using the `GlowPlayer#updateTitle` command to update the title builder. I am also going to make a method that takes in the new value and a `TitleMessage.Action` to update a specific attribute.

BSFishy:
I think all of this code should work. The last time I compiled it everything worked except the times subcommand, and it should be fixed now. I am going to test it out one last time, and I will let you know if it works.

BSFishy:
Ok. I am 100% on it working now. Titles, subtitles, and timings work now. JSON also works, and no errors are thrown.

setup.bat
```diff
@@ -0,0 +1 @@
+mvn package
```
mastercoms: Add a newline here.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -2122,14 +2124,111 @@ public void sendTitle(Title title) {
...
+        sendTitle(currentTitle);
+        currentTitle = new Title.Builder();
+    }
+
     @Override
     public void updateTitle(Title title) {
-        sendTitle(title); // TODO: update existing title instead of sending a new title
+        Title builtTitle = currentTitle.build();
+
+        if (title.getTitle() != null) {
```
mastercoms: Title can't be null. It's useless to check for null.

https://github.com/PaperMC/Paper/blob/master/Spigot-API-Patches/0017-Player-Tab-List-and-Title-APIs.patch#L134

BSFishy: Right. I was just doing this because I was thinking someone might use another title, with times, or a subtitle to update the existing title... Maybe we use a 

    if (title.getTitle().equals("")) {
???

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -129,9 +126,60 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
             sender.sendMessage("Updated " + player.getName() + "'s title");
+        } else if (action.equalsIgnoreCase("subtitle")) {
+            if (args.length < 3) {
+                sender.sendMessage(ChatColor.RED + "Usage: /title <player> " + action + " <raw json>");
+                return false;
+            }
+
+            StringBuilder message = new StringBuilder();
+
+            for (int i = 2; i < args.length; i++) {
```
mastercoms, line 144: What's going on here? The raw json must be a single argument.

EDIT: never mind, vanilla Minecraft allows for spaces in the JSON.

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -129,9 +126,60 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
+                if (i > 2) message.append(" ");
```
mastercoms: You don't need to append spaces in JSON.

src/main/java/net/glowstone/entity/GlowPlayer.java
mastercoms: But why should a non empty title be blocked?

src/main/java/net/glowstone/net/message/play/game/TitleMessage.java
```diff
@@ -41,9 +41,12 @@ public TitleMessage(Action action) {
     }
 
     public static TitleMessage[] fromTitle(Title title) {
+        TextMessage titleMessage = title.getTitle() != null ? asTextMessage(BaseComponent.toLegacyText(title.getTitle())) : asTextMessage(null);
```
mastercoms: The API does not allow for title to be null, so this check is useless.

setup.bat
Paulomart: why not remove this?

mastercoms: now that I think about it, a setup script for windows is useless. It's only build servers which use our setup.sh command, and they run on Linux.

gdude2002: Nope, I'm sure we have developers that use it as well - That said, you can use cygwin on Windows.

mastercoms: And isn't there Bash on Windows on Windows 10 as well?

gdude2002: It's not fully-functional outside of insider preview yet.

src/main/java/net/glowstone/entity/GlowPlayer.java
BSFishy: I meant to say

    if (!title.getTitle().equals("")) {
😛

setup.bat
BSFishy: Bash is on Windows, although it is called Powershell, and does not have nearly any features you would expect. I just prefer to have the ability to type `setup` rather than `mvn package`.

aramperes: I second this, I don't think it hurts anyone to have Windows build scripts along with the Linux ones.

aramperes: Also, you could edit the README to reflect this addition.

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -27,62 +28,56 @@ public TitleCommand() {
...
+        for (ChatColor c : style) {
+            text = c + text;
         }
 
         return text;
     }
 
     private static ChatColor toColor(String name) {
-        if (name.equals("obfuscated"))
+        if (name.equalsIgnoreCase("obfuscated"))
```
mastercoms: Don't ignore case.

setup.bat
mastercoms: If you're adding Windows scripts, it should include equivalents to all of our scripts, and be in a separate PR.

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -129,9 +126,60 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
+                message.append(args[i]);
+            }
+
+            String raw = message.toString().trim();
+            if (!validJson(raw)) {
+                sender.sendMessage(ChatColor.RED + "Invalid JSON: Could not parse, invalid format?");
+                return false;
+            }
+
+            String component = raw;
```
mastercoms, line 154: I would appreciate it if you could use Gson instead.

See `JsonObject` and `JsonParser`.

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -129,9 +126,60 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
+
+            StringBuilder message = new StringBuilder();
+
+            for (int i = 2; i < args.length; i++) {
+                if (i > 2) message.append(" ");
+                message.append(args[i]);
+            }
+
+            String raw = message.toString().trim();
+            if (!validJson(raw)) {
```
mastercoms, line 149: It would be cool if you could add the column a parsing error occurred, like vanilla does.

src/main/java/net/glowstone/entity/GlowPlayer.java
mastercoms: It's better to use `isEmpty()`.

src/main/java/net/glowstone/command/TitleCommand.java
BSFishy: Well, I think that that is what `JSONValue` does. I don't think it uses Gson, but I do think it uses a `JsonParser`. I am just using what was previously in the file. I will try to change it.

mastercoms: `JSONValue` is for jsonsimple.

BSFishy: Ok. I will try to replace all of the instances of jsonsimple in the class with gson

src/main/java/net/glowstone/command/TitleCommand.java
BSFishy: I can try, but I am not sure how I would do that with the current parser(Am changing it to Gson)

BSFishy:
I have everything working, I just am having trouble pushing the commits. I will try to push it as soon as possible.

EDIT: There we go 😛

BSFishy:
Just to be clear, everything is fixed... that I know of 😛

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -129,9 +157,31 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
                 component = convertJson((JSONObject) parsed);
             }
 
-            player.sendTitle(new Title(component));
+            ((GlowPlayer) player).updateTitle(TitleMessage.Action.SUBTITLE, component);
```
mastercoms, line 160: Why is this casted? Should it not be in Glowkit?

src/main/java/net/glowstone/command/TitleCommand.java
```diff
@@ -129,9 +157,31 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
 
-            sender.sendMessage("Updated " + player.getName() + "'s title");
+            sender.sendMessage("Updated " + player.getName() + "'s subtitle");
+        } else if (action.equalsIgnoreCase("times")) {
+            if (args.length != 5) {
+                sender.sendMessage(ChatColor.RED + "Usage: /title <player> " + action + " <fade in> <stay time> <fade out>");
+                return false;
+            }
+
+            if (!tryParseInt(args[2])) {
```
mastercoms, line 169: Replace this chain of if statements with a for-loop

mastercoms:
Well, seeing as this PR is inactive and the remaining suggestions are non blocking, I'll merge this. Thank you for your contribution.

mastercoms (approved):

PR #416, opened 2017-01-21
Implement rail placement

### LINKED ISSUE #292: Rail placement not implemented
Also, no explanation needed :smile: 

https://gyazo.com/6b56105b8d0920ffd46196d9ee0eedd5.gif

picknchew:
This PR implements rail placement.
(Vanilla behavior is not guaranteed, a few quirks)
Fixes #292.

Looking for feedback.


src/main/java/net/glowstone/block/blocktype/BlockRails.java
```diff
@@ -3,50 +3,443 @@
...
+    private static RailDirection getRailDirection(GlowBlock rail) {
+        // north - 0, south - 1, east - 2, west - 3, ascending_north - 4, ascending_south - 5, ascending_east - 6, ascending_west - 7
+        boolean[] unconnectedRails = new boolean[8];
+
+        GlowBlock north = rail.getRelative(BlockFace.NORTH);
+        GlowBlock south = rail.getRelative(BlockFace.SOUTH);
+        GlowBlock east = rail.getRelative(BlockFace.EAST);
+        GlowBlock west = rail.getRelative(BlockFace.WEST);
+
+        unconnectedRails[0] = isUnstable(north, rail) || isUnstable(north.getRelative(BlockFace.DOWN), rail);
```
aramperes: Since you are using magic values here (array index) for directions, perhaps you could name them?
e.g. `private static final int NORTH = 0;` and then use `unconnectedRails[NORTH] = isUnstable(north, rail) [...]`

aramperes:
I think it's pretty good feature-wise! :+1:
Only inconsistency I've found is that the redstone direction-switch is not implemented, but that's not really important so I don't mind if it's implemented later.

aramperes (approved):

mastercoms (approved):

mastercoms:

PR #419, opened 2017-02-04
Store serverbootstrapchannel

Shevchik:
Without storing bootstrap channel it's almost impossible inject into GlowStone network reliably.


mastercoms (approved):
Looks good to me.

PR #425, opened 2017-02-26
Fix TextMessage#flatten() throwing cast exceptions

MrPowerGamerBR:


MrPowerGamerBR:
Note to self: Never send a change using GitHub's web editor, that thing sucks.

aramperes (approved):

src/main/java/net/glowstone/util/TextMessage.java
```diff
@@ -71,7 +70,7 @@ private static void flatten(StringBuilder dest, JSONObject obj) {
             dest.append(obj.get("text"));
         }
         if (obj.containsKey("extra")) {
-            JSONArray array = (JSONArray) obj.get("extra");
+            LinkedList<?> array = (LinkedList<?>) obj.get("extra");
```
Paulomart, line 73: Better cast to a `List` or an `Collection`?

mastercoms:
Thanks for your contribution!

PR #439, opened 2017-03-03
GPGPU Support

mastercoms:
WIP branch for implementing GPGPU support within Glowstone. In this branch, it is used for noise generation, but can be easily applied to other systems after this is merged.

The scope of this PR could change to remove the noise gen modification for quicker merging of OpenCL support, since the OpenCL backend is sufficiently completed, however, it could have better support of OpenCL code. Perhaps a folder dedicated for third party code that could be used by plugins.


Deamon5550:
Do you have any benchmarks for this with the noise gen?

mastercoms:
@Deamon5550 Yes, but right now it is not focused on speed, just on a working implementation. I am working on optimizing it now that it is working. Right now, it takes about 2.5 minutes to generate a world compared to 10 seconds with the original implementation, but this is because of unoptimized loops and bad memory management. For the OpenCL implementation to be faster, all of the noise generation needs to be implemented in OpenCL, so that's what I'm working on right now. Once I have that pushed, I'll publish benchmarks.

mastercoms:
Most work from this PR has been integrated into master with https://github.com/GlowstoneMC/Glowstone/commit/2b731f3a1d2709e15f5b9bcb9f63644be12a8953. This commit does not include any usage of OpenCL, unlike this PR. Usage of OpenCL will be done later in master.

PR #444, opened 2017-03-08
Implement chunk section palettes

Pokechu22:
This PR implements palettes for individual chunk sections, which should improve performance (bandwidth and memory wise), both on the server _and_ client sides.  I don't have any concrete benchmarks for this, but it should in most cases lead to chunk data being 4/13ths the size (when possible, though larger when needed).

The algorithm I'm using here is fairly simple:

1. Iterate through all blocks in the section.  For each, add the type to a FastUtil `IntList`, if the list doesn't already contain it.  This creates the palette.
2. Once the palette has been built, calculate the base-2 logarithm of the size to get the bits per block.
3. If the bits per block is less than 4, increase it to 4 (as required for the client).
4. If the bits per block is more than 8, set it to 13, and remove the palette.
5. Create a `VariableValueArray` with the bits per block.
6. Iterate through all blocks in the section, again.  For each, put it into the `VariableValueArray` the appropriate type (based off the brute-force based method `palette.indexOf(type)` if the palette isn't present) at the correct index.

Similarly, there's logic to expand the size of the palette when `set` is called, if needed.

That algorithm may be improvable (potentially by changing the order of the palette, so that something faster than `IntArrayList.indexOf` could be used), but it's a decent enough algorithm for the moment (if there are any suggestions on how to make it better, let me know!)

One thing to note is that the palette will slowly get larger within a chunk section until it is saved and reloaded; palette IDs aren't ever "freed" automatically.  This isn't ideal, but it's more expensive to try and automatically free them (and remap all of the items in the `VariableValueArray` when doing so) as far as I can tell.  Again, there may be better algorithms for that.  I did add an `optimize` method that manually rebalances the palette, but it isn't called anywhere yet (and may not need to be called anywhere).

Here's a test world that should be good for this (it contains sections with 4, 5, 6, 7, 8, and 13 bits per block, in large all-stone chunk columns): [TestWorldAllStone.zip](https://github.com/GlowstoneMC/Glowstone/files/826504/TestWorldAllStone.zip)

I made this PR to check my understanding of the palette system (while I was writing the [revised chunk format article](http://wiki.vg/User:Pokechu22/Chunk_Format)).


src/main/java/net/glowstone/chunk/GlowChunk.java
```diff
@@ -242,7 +246,7 @@ public void initializeSections(ChunkSection... initSections) {
         }
         //GlowServer.logger.log(Level.INFO, "Initializing chunk ({0},{1})", new Object[]{x, z});
 
-        sections = new ChunkSection[DEPTH / SEC_DEPTH];
+        sections = new ChunkSection[SEC_COUNT];
         System.arraycopy(initSections, 0, sections, 0, Math.min(sections.length, initSections.length));
```
mastercoms: Do you think it would be beneficial to add a method to init an individual section by index?

src/main/java/net/glowstone/io/anvil/AnvilChunkIoService.java
```diff
@@ -208,27 +208,29 @@ public void write(GlowChunk chunk) throws IOException {
             CompoundTag sectionTag = new CompoundTag();
```
mastercoms, line 205: If you add a individual init sections method, could you add methods for reading and writing individual sections?

mastercoms:
Looks good with my initial review. Suggestions so far (about individual chunk section handling) are not essential for merging this, but they would be nice to have.

src/main/java/net/glowstone/chunk/GlowChunk.java
Pokechu22: I'm not entirely sure what that would look like.  Do you mean replacing `initializeSections(ChunkSection... initSections)` with `initializeSection(int index, ChunkSection section)`?  There might be _some_ benefit, but I don't know where that would be used (all places where `initializeSections` are used already build an array).  It would get rid of the array with null members being passed, though.

mastercoms: Yeah, I meant for it be used to skip passing null members.

Pokechu22: Alright, I'll try to do that, shouldn't be too hard.

Pokechu22: Done (correctly, this time).

src/main/java/net/glowstone/io/anvil/AnvilChunkIoService.java
Pokechu22: Done.

mastercoms (approved):

mastercoms:
@Pokechu22 Seems like this PR made lighting stop working. Everything is bright during the night.

RhvsW:
Before (https://github.com/GlowstoneMC/Glowstone/commit/90eb1e3d2812969381342c16d89a25c1aca0edfc)

before

After (https://github.com/GlowstoneMC/Glowstone/commit/96bd6de81c09f202e44defcab7a9b70b837a7b1d)

after
after underground

Pokechu22:
Hm, I might have broken something with sky light (I did refactor the way that it's loaded).  It's probably a simple mistake, I'll look.

mastercoms:
Ok thanks. I didn't think to test sky light because I saw that it was just a small refactor. Oh well. Thanks for your quick response and help for this issue.

Pokechu22:
Can you take a screenshot with F3 opened?  (I don't have access to a client right now; F3 would let me see what types of light are set - SL and BL)

RhvsW:
Build before PR: before
Latest build: latest

Pokechu22:
> Light: 15 (0 sky, 15 block)

I did not expect _that_.  Hm.  Oh, I see what happened...  I'm pre-populating the blocklight array with 0xF, when I meant to populate the skylight array with 0xF.  That's easy to fix.

PR #445, opened 2017-03-09
Unofficial PacketAPI

kamcio96:


mastercoms:
Are you continuing work for this, or would you like for an initial API to be added at this state?

kamcio96:
No, only simplest API :wink:

ghost:
Events is not thread-safe. You need call both events in `GlowSession`. There must be something like read/write queue that's is proccessed every tick.

kamcio96:
That events are async (don't contain `Async` in name but are async). As I said, its simplest api.
For call in tick:
```java
    private Set<Message> ignoreMessage = new HashSet<>();
    
    @EventHandler
    public void onPacketSend(PacketSendEvent event) {
        Message message = event.getMessage();

        if (ignoreMessage.remove(message)) {
            return;
        }
        
        event.setCancelled(true);

        new BukkitRunnable() {
            @Override
            public void run() {
                // do something with message
                ignoreMessage.add(message);
                event.getSession().send(message);
            }
        }.runTask(this);
    }

```

ghost:
Why not make it thread-safe?

kamcio96:
It is `thread-safe`. Reading packets in main tick can cause big lags.

MrPowerGamerBR:
@ensirius if Glowstone's networking system is all async, why run the event in the main thread?

Paulomart:
As far as I see, the example provided is not even thread safe.

mastercoms:
I think it will be best to figure out a packet API as a group, during the designated update month for networking, where we will most likely also switch to using native netty.

Issue #454, opened 2017-04-15
Entity#setGravity(false) leaves gravity enabled


satoshinm:
When `.setGravity(false)` is called on an `Entity`, physics are supposed to be disabled allowing the entity to float mid-air, but instead the call appears to have no effect and entities still fall to the ground.

Looking at the code, https://github.com/GlowstoneMC/Glowstone/blob/47b0e0b699d607f9d1d196a1aa9897d055928aa9/src/main/java/net/glowstone/entity/GlowEntity.java#L1056 setGravity sets `this.gravity` but I can't find where it is accessed, may be a simple check `if (this.gravity)` that is missing surrounding the relevant physics code.

Glowstone build 444

VaiTon:
> physics are supposed to be disabled

Not all physics, just gravity, so that entities can be still pushed (TNT, other entities, damage, water, etc...)

[Here](http://gph.is/2oxmkTu) is an example
So we should modify only these two lines ([1](https://github.com/GlowstoneMC/Glowstone/blob/47b0e0b699d607f9d1d196a1aa9897d055928aa9/src/main/java/net/glowstone/entity/GlowEntity.java#L720) , [2](https://github.com/GlowstoneMC/Glowstone/blob/47b0e0b699d607f9d1d196a1aa9897d055928aa9/src/main/java/net/glowstone/entity/GlowEntity.java#L774))

mastercoms:
Yep, I forgot about that method.

PR #469, opened 2017-05-02
Respect Entity#setGravity(false). Closes GH-454

satoshinm:
Adds support for calling `setGravity(false)` to disable gravity physics on an entity. This is useful for plugins to make their own entities which don't respect gravity:

screen shot 2017-05-01 at 9 59 22 pm

Implemented by adding a `getGravityAccel()` accessor to complement `setGravityAccel()` (and `getVelocity()`), which returns a zero vector if `!this.gravity`.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowEntity.java
@@ -317,6 +317,14 @@ public abstract class GlowEntity implements Entity {
         this.gravityAccel = gravity;
     }
 
+    public Vector getGravityAccel() {
+        if (this.gravity) {
+            return this.gravityAccel;
+        } else {
+            return this.zeroG;
+        }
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -716,6 +724,7 @@ public boolean intersects(BoundingBox box) {
      * Gravity acceleration applied each tick.
      */
     protected Vector gravityAccel = new Vector(0, -0.04, 0);
+    private static final Vector zeroG = new Vector(0, 0, 0);
```
mastercoms, line 727: This can be `new Vector()`, FYI.

mastercoms:
Merged in 447c945c9ee660a4775e794555037fa5567c17f5.

Issue #462, opened 2017-04-20
[#458] Dispensers don't apply armour to nearby players


gdude2002:
This issue has been posted to separate the following ticket into smaller issues: #458

mastercoms:
Fixed in https://github.com/GlowstoneMC/Glowstone/commit/f171c8203586d1a6e46b2a0d63f0ad7261078f44.

PR #470, opened 2017-05-07
Added dispense armor feature

Christian7573:
I've changed two things:

1. I've added the feature reported missing in #462, the ability to equip armor to a player with a dispenser.
Note: This is WIP because upon writing this pull request I realized that I only included armor, while there are other equitable items such as heads and elytras.

2. During testing of item 1, I discovered that items with a unique dispenser behavior would only use that behavior properly "when they felt like it." (I didn't understand the dispenser slot randomization code) I rewrote `getDispenseSlot()` of GlowDispenser to fix this.


src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,104 @@
...
+import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.Location;
+import org.bukkit.World;
+import org.bukkit.entity.Player;
+import java.util.List;
+
+/**
+ *
+ * @author cag45
```
aramperes: We don't really use `@author` tags in Glowstone, could you remove this comment block?

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,104 @@
...
+            case GOLD_LEGGINGS: armor = Material.DIAMOND_LEGGINGS; break;
+            case IRON_LEGGINGS: armor = Material.DIAMOND_LEGGINGS; break;
+            case CHAINMAIL_LEGGINGS: armor = Material.DIAMOND_LEGGINGS; break;
+            case LEATHER_CHESTPLATE: armor = Material.DIAMOND_CHESTPLATE; break;
+            case GOLD_CHESTPLATE: armor = Material.DIAMOND_CHESTPLATE; break;
+            case IRON_CHESTPLATE: armor = Material.DIAMOND_CHESTPLATE; break;
+            case CHAINMAIL_CHESTPLATE: armor = Material.DIAMOND_CHESTPLATE; break;
+        }
+        
+        List<Player> players = world.getPlayers();
```
aramperes: I think this functionality applies to other entities as well (those who can equip armor). You can use [`world.getNearbyEntities(location, x, y, z)`](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#getNearbyEntities(org.bukkit.Location,%20double,%20double,%20double)) to look through entities in an area.

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -68,17 +89,44 @@ public boolean dispense() {
...
-
-        int slot = -1;
-        int randomChance = 1;
-
-        for (int i = 0; i < inv.getSize(); i++) {
-            if (inv.getItem(i) != null && random.nextInt(randomChance++) == 0) {
-                slot = i;
+        
+        boolean allAir = true;
+        if (inv.getItem(0).getType() != Material.AIR) allAir = false;
```
aramperes: I would use a for-loop here... but can't you use `isEmpty()` in this case?

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -68,17 +89,44 @@ public boolean dispense() {
...
+        if (inv.getItem(3).getType() != Material.AIR) allAir = false;
+        if (inv.getItem(4).getType() != Material.AIR) allAir = false;
+        if (inv.getItem(5).getType() != Material.AIR) allAir = false;
+        if (inv.getItem(6).getType() != Material.AIR) allAir = false;
+        if (inv.getItem(7).getType() != Material.AIR) allAir = false;
+        if (inv.getItem(8).getType() != Material.AIR) allAir = false;
+        if (allAir) { //Determines if the dispenser is empty, if so return slot 0
+            return 0;
+        } else {
+
```
aramperes: Extra whitespace.

aramperes (changes requested):

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Sure, forgot that NetBeans automatically adds those.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Ok, will implement that

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Didn't know that was a function, yes I will use that.

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Will be removed.

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -68,17 +90,37 @@ public boolean dispense() {
...
-        int slot = -1;
-        int randomChance = 1;
-
-        for (int i = 0; i < inv.getSize(); i++) {
-            if (inv.getItem(i) != null && random.nextInt(randomChance++) == 0) {
-                slot = i;
+        
+        boolean isEmpty = true;
+        for (ItemStack stack : inv.getContents()) {
+            if (InventoryUtil.isEmpty(stack) == false) isEmpty = false;
```
aramperes: For efficiency, you should break your loop after finding a non-empty stack. Also, use `!InventoryUtil.isEmpty(stack)`.

aramperes (changes requested):

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Thanks for the advice, changes made.

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -68,17 +90,37 @@ public boolean dispense() {
...
+        } 
+        if (isEmpty == false) {
```
aramperes: Same thing here, you should use `!isEmpty`.

aramperes (dismissed):

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Ok

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -68,17 +90,37 @@ public boolean dispense() {
...
+            int slot = 0;
+            int tries = 0;
+            while (true) {
+                //Gets a random slot and determines if that slot has an item in it
+                slot = random.nextInt(9);
+                if (inv.getItem(slot).getType() != Material.AIR) break;
+                
+                //Eliminates the possibility of an infinite loop, goes in order upon reaching 20 random tries
+                tries += 1;
+                if (tries == 20) {
```
picknchew: Algorithm is unreliable and the number chosen is arbitrary. A better way of selecting an item would be finding all slots that have items and choosing one.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,95 @@
...
+        Player player;
+        Location location;
+        Boolean location1Test;
+        Boolean location2Test;
```
picknchew: Use primitive for boolean instead. These aren't needed anyways. Just compare them directly.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,95 @@
...
+        int i = 0;
+        //Loop through players to see if any are in the location where armor would be dispensed
+        while (i < players.size()) {
```
picknchew: For loop is much more suitable for this

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,95 @@
...
+public class ArmorDispenseBehavior extends DefaultDispenseBehavior {
+    DefaultDispenseBehavior defaultBehavior = new DefaultDispenseBehavior();
+    
+    @Override
+    protected ItemStack dispenseStack(GlowBlock block, ItemStack stack) {
+        Material armor = stack.clone().getType();
+        BlockFace facing = BlockDispenser.getFacing(block);
+        GlowBlock target = block.getRelative(facing);
+        Location location1 = target.getLocation();
+        Location location2 = location1.clone();
```
picknchew: There is no need to create a new location to compare them.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,95 @@
...
+import org.bukkit.World;
+import org.bukkit.entity.Player;
+import java.util.List;
+
+public class ArmorDispenseBehavior extends DefaultDispenseBehavior {
+    DefaultDispenseBehavior defaultBehavior = new DefaultDispenseBehavior();
+    
+    @Override
+    protected ItemStack dispenseStack(GlowBlock block, ItemStack stack) {
+        Material armor = stack.clone().getType();
```
picknchew: No need to clone

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,95 @@
...
+            player = players.get(i);            
+            location = player.getLocation().clone();
+            location.setX(location.getBlockX());
+            location.setY(location.getBlockY());
+            location.setZ(location.getBlockZ());            
+            location1Test = (location.getX() == location1.getX() && location.getY() == location1.getY() && location.getZ() == location1.getZ());
+            location2Test = (location.getX() == location2.getX() && location.getY() == location2.getY() && location.getZ() == location2.getZ());
+            if (location1Test || location2Test) {
+                //Check if the aproperiate armor slot is empty, if so equip armor.
+                if (armor == Material.DIAMOND_BOOTS) {
```
picknchew: Instead of doing this, it'd be much better to find a way to differentiate the different types of armor internally.

picknchew (changes requested):

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
aramperes: Yes, and the same system would be useful when we implement entities picking up armor on the ground (and upgrading their equipment)

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: It's not that, if a dispenser is placed at your head level it should also be able to equip items. That is why there is 2 locations, 1 for the block which the dispenser if facing and another one for 1 block below it. If you're referring to `facing` and `target` those were copy-pasted from `BucketDispenseBehavior`

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Those were put there to keep the if statement lines short, but I'll change it if you wish.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Agreed

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: So perhaps add an integer called Wearable to ItemType with 0 meaning cannot be worn, 1 for head, 2 for chestplate, 3 for leggings and 4 for boots?

aramperes: Perhaps, or an enum could be better. Would also need a "value" level to determine which armor material is better than another, etc.

Christian7573: Okee, I'll see what I can do.

Christian7573:
Ability to add armor to armor stands is not going to be implemented yet because of #471

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -130,15 +130,17 @@ protected final void setMaxStackSize(int maxStackSize) {
      * @param holding The ItemStack the player was holding
      */
     public void rightClickAir(GlowPlayer player, ItemStack holding) {
-        // nothing by default
+        if (fastEquip) {
```
mastercoms: This will cause dupe usage, as the client sends a right click and use packet. This is not meant to check if self usage is only allowed.

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -130,15 +130,17 @@ protected final void setMaxStackSize(int maxStackSize) {
...
+            equipStack(player, holding);
```
mastercoms: This will cause dupe usage, as the client sends a right click and use packet. This is not meant to check if self usage is only allowed.

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -130,15 +130,17 @@ protected final void setMaxStackSize(int maxStackSize) {
...
+        }
     }
 
     /**
      * If this item can only be used without any context (essentially used in air)
      * @return If this item can only be used without any context
      */
     public boolean canOnlyUseSelf() {
-        return false;
+        return fastEquip;
```
mastercoms: This is not supposed to be a getter for any fields. It's there for subclasses to define themselves as items with such behavior.

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -164,10 +166,67 @@ public final String toString() {
...
     
+    //Code for wearable items
+    private ItemWearablePosition wearablePosition = ItemWearablePosition.NON_WEARABLE;
+    private int armorPoints = 0;
+    private boolean fastEquip = false;
+    
     public ItemWearablePosition getWearablePosition() {
         return wearablePosition;
     }
     public void setWearablePosition(ItemWearablePosition position) {
```
mastercoms: This is invalid for most items. Keep the subclass for wearable items. Is there a reason the class was removed? And what were the use cases of the class and these methods?

mastercoms (changes requested):

src/main/java/net/glowstone/block/ItemTable.java
```diff
@@ -301,6 +299,31 @@ private void registerBuiltins() {
         reg(Material.HOPPER_MINECART, new ItemMinecart(GlowMinecart.MinecartType.HOPPER));
         reg(Material.POWERED_MINECART, new ItemMinecart(GlowMinecart.MinecartType.FURNACE));
         reg(Material.STORAGE_MINECART, new ItemMinecart(GlowMinecart.MinecartType.CHEST));
+        
+        //Wearable Items
+        reg(Material.LEATHER_BOOTS, new ItemType().wear(1, ItemWearablePosition.FEET, true));
```
mastercoms: Make a subclass with a constructor for this stuff, after explaining why this functionality is needed.

mastercoms (changes requested):

src/main/java/net/glowstone/block/ItemTable.java
mastercoms: Ah, I see it's for determining the armor type of the material? You should use `EnchantmentTarget` for that, so just revert all of this unless there's more uses for `ItemWearable`.

Christian7573: Well items such as heads, pumpkins, and elytras can be work but cannot be enchanted. The integer is to determine how many armor points wearing that item gives you. 

>  Would also need a "value" level to determine which armor material is better than another.

Then for where you wear this item and then if you should be able to equip this item by right clicking it.

src/main/java/net/glowstone/block/itemtype/ItemType.java
Christian7573: This is for the ability to equip armor by right clicking it, as I remember checking and it not being implemented.

src/main/java/net/glowstone/block/itemtype/ItemType.java
Christian7573: This is for the ability to equip armor by right clicking it, as I remember checking and it not being implemented. If it can be right clicked it should not be placed.

src/main/java/net/glowstone/block/itemtype/ItemType.java
Christian7573: I removed it because Skull and Pumpkin had different classes that both extended from different classes, so simply extending them from ItemWearable wouldn't work

src/main/java/net/glowstone/block/ItemTable.java
mastercoms: Equipping armor with right click was working before. Don't include a fix for it in this PR, since it's a regression somewhere else and adding a new implementation for this feature won't fix the root issue. Though, I appreciate pointing my attention to this new bug.

src/main/java/net/glowstone/block/itemtype/ItemType.java
mastercoms: Then use an interface with default methods for now until the new types system comes around, which will fix these inheritance issues.

src/main/java/net/glowstone/block/itemtype/ItemType.java
mastercoms: See https://github.com/GlowstoneMC/Glowstone/pull/470#discussion_r115345972.

src/main/java/net/glowstone/block/itemtype/ItemType.java
mastercoms: See https://github.com/GlowstoneMC/Glowstone/pull/470#discussion_r115345972.

src/main/java/net/glowstone/block/ItemTable.java
Christian7573: Ok, and how did you reference a comment?

src/main/java/net/glowstone/block/itemtype/ItemType.java
Christian7573: Ok.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
picknchew: There is no need to clone the objects to compare them. A simple check between the values will do.

picknchew (changes requested):

src/main/java/net/glowstone/block/ItemTable.java
mastercoms: Just posted the comment link and GitHub handles formatting.

src/main/java/net/glowstone/block/itemtype/ItemWearablePosition.java
```diff
@@ -0,0 +1,9 @@
+package net.glowstone.block.itemtype;
+
+public enum ItemWearablePosition {
```
mastercoms: This class is redundant, use EquipmentSlot.

src/main/java/net/glowstone/util/Typeless.java
```diff
@@ -0,0 +1,12 @@
+package net.glowstone.util;
+
+public class Typeless<T> {
```
mastercoms: What is this for?

src/main/java/net/glowstone/block/itemtype/ItemWearable.java
```diff
@@ -0,0 +1,60 @@
+package net.glowstone.block.itemtype;
+
+import org.bukkit.Material;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.inventory.ItemStack; 
+
+public class ItemWearable extends ItemType {
```
mastercoms: I still don't understand what ItemWearable is used for the dispenser logic.

If you're going to use it to equip find equipment slot type, just use `EnchantmentTarget.includes` for normal armor, and then a few extra boolean checks for the other non-enchantable materials, like pumpkins or skulls. 

You could make some sort of utility class for it too, to check for those extra materials, as well as to do the correct equipping for each equipment slot type.

I just don't think this data needs to be instanced per material in a new wearable class.

mastercoms (changes requested):

src/main/java/net/glowstone/util/Typeless.java
Christian7573: Pretty sure I deleted that, must have accidentally restored it when restoring other commits.

src/main/java/net/glowstone/block/itemtype/ItemWearable.java
Christian7573: Oh, sorry, did not realize that EnchantmentTarget included things like ARMOR_FEET, I just saw ARMOR.

Christian7573:
Note: Latest commits have not been tested yet.

src/main/java/net/glowstone/util/Wearable.java
```diff
@@ -0,0 +1,120 @@
...
+            return wearablePosition;
+        }
+        public int armorPoints() {
+            return armorPoints;
+        }
+    }
+    
+    protected Map<Material, WearableReturn> createMap() {
+        Map<Material, WearableReturn> map2 = new HashMap<>();
+        map2.put(Material.LEATHER_BOOTS, new WearableReturn(EnchantmentTarget.ARMOR_FEET, 1));
```
mastercoms: Don't use `EnchantmentTarget` for specifying slot, only for determining it. Slot should be specified with `EquipmentSlot` (if needed) and it is not needed to define armor to be put in a certain slot, as that can be checked with the `includes(Material)` method in `EnchantmentTarget`. For example, `EnchantmentTarget.ARMOR_HEAD.includes(Material.GOLD_HELMET)` would return true.

src/main/java/net/glowstone/util/Wearable.java
```diff
@@ -0,0 +1,120 @@
...
+        private int armorPoints;
+        
+        public WearableReturn(EnchantmentTarget position, int points) {
+            this.wearablePosition = position;
+            this.armorPoints = points;
+        }
+        public EnchantmentTarget wearPosition() {
+            return wearablePosition;
+        }
+        public int armorPoints() {
```
mastercoms: In Java, accessors are typically named with the prefix `get`, so for example, your method should be called `getArmorPoints()` instead of just `armorPoints()`.

src/main/java/net/glowstone/util/Wearable.java
```diff
@@ -0,0 +1,120 @@
...
+        map2.put(Material.DIAMOND_LEGGINGS, new WearableReturn(EnchantmentTarget.ARMOR_LEGS, 6));
+        map2.put(Material.DIAMOND_CHESTPLATE, new WearableReturn(EnchantmentTarget.ARMOR_TORSO, 8));
+        map2.put(Material.DIAMOND_HELMET, new WearableReturn(EnchantmentTarget.ARMOR_HEAD, 3));
+        map2.put(Material.SKULL_ITEM, new WearableReturn(EnchantmentTarget.ARMOR_HEAD, 0));
+        map2.put(Material.ELYTRA, new WearableReturn(EnchantmentTarget.ARMOR_TORSO, 0));
+        map2.put(Material.PUMPKIN, new WearableReturn(EnchantmentTarget.ARMOR_HEAD, 0));
+        return map;
+    }
+    
+    public static void intalize() {
```
mastercoms: Since your map is static, you can use a static method to **initialize** your map. So, if you were to keep `createMap()` add `static` after `protected`.

Then, you can run that static method using a static block:

```java
static {
  createMap();
}
```

src/main/java/net/glowstone/util/Wearable.java
```diff
@@ -0,0 +1,120 @@
...
+        public EnchantmentTarget wearPosition() {
+            return wearablePosition;
+        }
+        public int armorPoints() {
+            return armorPoints;
+        }
+    }
+    
+    protected Map<Material, WearableReturn> createMap() {
+        Map<Material, WearableReturn> map2 = new HashMap<>();
```
mastercoms: You can assign `map` to `new HashMap<>()` in the field declaration. It doesn't need to be separated.

Then, you can directly add entries to `map` instead of using `map2` and then assigning `map` to `map2`.

src/main/java/net/glowstone/util/Wearable.java
```diff
@@ -0,0 +1,120 @@
...
+
+import org.bukkit.enchantments.EnchantmentTarget;
+import org.bukkit.Material;
+import java.util.Map;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.inventory.ItemStack;
+import java.util.HashMap;
+import java.util.Set;
+
+public final class Wearable {
```
mastercoms: After seeing this class, I have understood what you're trying to do. This behavior is already implemented in `GlowPlayerInventory`. Use `tryToFillSlots(armor, 36, 40)`

I have put some comments about your code in case you need to use similar Java code in the future.

mastercoms (changes requested):

src/main/java/net/glowstone/util/Wearable.java
Christian7573: Oh, didn't think of that.

src/main/java/net/glowstone/util/Wearable.java
Christian7573: Since I didn't realize I could set map as `new HashMap<>()` before, I was defining map2 as a local variable, but whenever I tried to access it NetBeans yelled at me for "accessing a non-static variable from a static method"

src/main/java/net/glowstone/util/Wearable.java
Christian7573: Ok.

src/main/java/net/glowstone/util/Wearable.java
Christian7573: Ok

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Is this not supposed to be arbitrary?

Christian7573:
So `tryToFillSlots()` from `GlowInventory` returns null if all items are placed successfully. Can I return null from `ArmorDispenseBehavior` or do I have to check if it's null and return `new ItemStack(Material.AIR)`?

mastercoms:
`null` is fine, but you can use `InventoryUtil.itemOrEmpty()`

Christian7573:
Ok, thanks.

src/main/java/net/glowstone/block/ItemTable.java
```diff
@@ -234,7 +234,6 @@ private void registerBuiltins() {
         reg(Material.BREWING_STAND_ITEM, new ItemPlaceAs(Material.BREWING_STAND));
         reg(Material.CAULDRON_ITEM, new ItemPlaceAs(Material.CAULDRON));
         reg(Material.FLOWER_POT_ITEM, new ItemPlaceAs(Material.FLOWER_POT));
-        reg(Material.SKULL_ITEM, new ItemPlaceAs(Material.SKULL));
```
mastercoms: I don't think this diff is needed.

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -32,11 +33,34 @@ public static DispenseBehaviorRegistry getDispenseBehaviorRegistry() {
     public static void register() {
         // register all dispense behaviors
         DefaultDispenseBehavior bucketDispenseBehavior = new BucketDispenseBehavior();
+        ArmorDispenseBehavior armorDispenseBehavior = new ArmorDispenseBehavior();
```
mastercoms: Move this line to right before you register the behavior for all of the new materials. (line 42 in current file)

src/main/java/net/glowstone/block/state/GlowDispenser.java
mastercoms: Generally, you shouldn't use tries for this. An inventory isn't a black box where we don't know how many slots there are or if any of the have items in it. We can gather all slots with items through inventory methods.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,56 @@
...
+        BlockFace facing = BlockDispenser.getFacing(block);
+        GlowBlock target = block.getRelative(facing);
+        Location location1 = target.getLocation();
+        Location location2 = location1.clone();
+        location2.setY(location2.getY() - 1);
+        World world = location1.getWorld();
+        
+        //Find all nearby entities and see if they are players or armor stands
+        List<LivingEntity> entities = new ArrayList<>();
+        for (Entity cEntity : world.getNearbyEntities(location1, 3, 3, 3)) {
```
mastercoms: Any reason why this is named `cEntity` instead of `entity`?

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class ArmorDispenseBehavior extends DefaultDispenseBehavior {
+    DefaultDispenseBehavior defaultBehavior = new DefaultDispenseBehavior();
+    
+    @Override
+    protected ItemStack dispenseStack(GlowBlock block, ItemStack stack) {
+        Material armor = stack.getType();
+        BlockFace facing = BlockDispenser.getFacing(block);
+        GlowBlock target = block.getRelative(facing);
+        Location location1 = target.getLocation();
```
mastercoms: Could you explain to me what these locations are used for?

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,56 @@
...
+        List<LivingEntity> entities = new ArrayList<>();
+        for (Entity cEntity : world.getNearbyEntities(location1, 3, 3, 3)) {
+            switch (cEntity.getType()) {
+                case PLAYER: entities.add((LivingEntity) cEntity); break;
+                //case ARMOR_STAND: entities.add((LivingEntity) cEntity); break;
+            }
+        }  
+        
+        Location location;
+        Boolean location1Test;
```
mastercoms: Use the `boolean` primitive type.

mastercoms (changes requested):

src/main/java/net/glowstone/block/ItemTable.java
Christian7573: What is "this diff"
If you're wondering why I removed it it's just been moved to the bottom of the list, I can't remember why though. I think it was because I used to have all wearables in one place.

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Ok

src/main/java/net/glowstone/block/ItemTable.java
mastercoms: A diff is your changes from the previous revision of a file. I am just saying that it isn't necessary to move the line down.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: `Location1` is the absolute location of the block the dispenser if facing. `Location2` is one block below that, because in my vanilla tests dispensers can equip armor to you if they are at your head level.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: I think I used to have a variable called entity, so I put a `c` infront of it for `currentEntity`. I'll change it to entity if you like.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Ok

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
mastercoms: You can just compare
entity.getLocation() and entity.getEyeLocation() to the target location.

Christian7573:
For whatever reason, GitHub isn't letting me respond to your comment about tries. First of all, I assumed from the previous code for getting a slot that it was a random decision, if it's not let me know and I'll make it go in order. If it is, I'm hearing that you want to randomly select from the only the ones that have items in them, which makes sense.

src/main/java/net/glowstone/block/ItemTable.java
Christian7573: Oh ok.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Is `getEyeLocation()` a block location?

mastercoms: It is a precise location so you'll have to compare the block coordinates.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
mastercoms: Yes please.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Ok

mastercoms:
Yes, when there is more than one slot occupied by items in a dispenser, a random occupied slot is dispensed.

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -68,17 +94,7 @@ public boolean dispense() {
 
     public int getDispenseSlot() {
         Inventory inv = getInventory();
```
mastercoms: I don't think this is needed. Reference `getInventory()` directly in the `getRandomSlot()` call.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class ArmorDispenseBehavior extends DefaultDispenseBehavior {
+    DefaultDispenseBehavior defaultBehavior = new DefaultDispenseBehavior();
+    
+    @Override
+    protected ItemStack dispenseStack(GlowBlock block, ItemStack stack) {
+        Material armor = stack.getType();
+        BlockFace facing = BlockDispenser.getFacing(block);
+        GlowBlock target = block.getRelative(facing);
+        Location location1 = target.getLocation();
```
mastercoms: Rename this to `targetLocation`, please.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,54 @@
...
+        World world = location1.getWorld();
+        
+        //Find all nearby entities and see if they are players or armor stands
+        List<LivingEntity> entities = new ArrayList<>();
+        for (Entity entity : world.getNearbyEntities(location1, 3, 3, 3)) {
```
mastercoms: Inline `targetLocation.getWorld()`.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,54 @@
...
+                //case ARMOR_STAND: entities.add((LivingEntity) cEntity); break;
+            }
+        }  
+        
+        Location location;
+        boolean location1Test;
+        boolean location2Test;
+        //Loop through entities to see if any are in the location where armor would be dispensed
+        for (LivingEntity player : entities) {          
+            location = player.getLocation().clone();
```
mastercoms: I don't think this is needed. Just use the block coordinate getters.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,54 @@
...
+            location.setX(location.getBlockX());
+            location.setY(location.getBlockY());
+            location.setZ(location.getBlockZ());            
+            location1Test = (location.getX() == location1.getX() && location.getY() == location1.getY() && location.getZ() == location1.getZ());
+            location2Test = (location.getX() == player.getEyeLocation().getBlockX() && player.getEyeLocation().getBlockY() == location1.getY() && player.getEyeLocation().getBlockZ() == location1.getZ());
```
mastercoms: Compare the target's location to the player's eye location. Right now, the code is comparing the player's location to their eye location.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,54 @@
...
+            if ((location1Test || location2Test)) {
+                return ((GlowInventory) ((Player) player).getInventory()).tryToFillSlots(stack, 36, 40);
+            }
+        }
+        return defaultBehavior.dispense(block, stack); //Fallback
```
mastercoms: Comments should have a space after the `//`

mastercoms (changes requested):

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Nice idea

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Sure

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Ok.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: By "block coordinate getters" are you referring to `location.getBlockX()` and Y and Z?

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Oh wow can't believe I did that. Will be fixed.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Ok

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -66,21 +90,6 @@ public boolean dispense() {
         return true;
     }
 
-    public int getDispenseSlot() {
```
mastercoms: I meant
```java
public int getDispenseSlot() {
  return InventoryUtil.getRandomSlot(random, getInventory(), true);
}
```

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -32,11 +33,34 @@ public static DispenseBehaviorRegistry getDispenseBehaviorRegistry() {
     public static void register() {
         // register all dispense behaviors
         DefaultDispenseBehavior bucketDispenseBehavior = new BucketDispenseBehavior();
+        ArmorDispenseBehavior armorDispenseBehavior = new ArmorDispenseBehavior();
         getDispenseBehaviorRegistry().putBehavior(Material.WATER_BUCKET, bucketDispenseBehavior);
         getDispenseBehaviorRegistry().putBehavior(Material.LAVA_BUCKET, bucketDispenseBehavior);
         getDispenseBehaviorRegistry().putBehavior(Material.BUCKET, new EmptyBucketDispenseBehavior());
         getDispenseBehaviorRegistry().putBehavior(Material.FLINT_AND_STEEL, new FlintAndSteelDispenseBehavior());
         getDispenseBehaviorRegistry().putBehavior(Material.TNT, new TNTDispenseBehavior());
```
mastercoms: Move `ArmorDispenseBehavior armorDispenseBehavior = new ArmorDispenseBehavior();` to here

src/main/java/net/glowstone/block/ItemTable.java
mastercoms: Here's just a reminder to do this :)

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,50 @@
...
+        Location location;
+        boolean targetLocationTest1;
+        boolean targetLocationTest2;
+        // Loop through entities to see if any are in the location where armor would be dispensed
+        for (LivingEntity player : entities) {          
+            location = player.getLocation().clone();
+            location.setX(location.getBlockX());
+            location.setY(location.getBlockY());
+            location.setZ(location.getBlockZ());            
+            targetLocationTest1 = (player.getLocation().getBlockX() == targetLocation.getX() && player.getLocation().getBlockY() == targetLocation.getY() && player.getLocation().getBlockZ() == targetLocation.getZ());
```
mastercoms: Make sure you're using getBlockX, getBlockY and getBlockZ for all locations.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,50 @@
...
+        // Find all nearby entities and see if they are players or armor stands
+        List<LivingEntity> entities = new ArrayList<>();
+        for (Entity entity : targetLocation.getWorld().getNearbyEntities(targetLocation, 3, 3, 3)) {
+            switch (entity.getType()) {
+                case PLAYER: entities.add((LivingEntity) entity); break;
+                // case ARMOR_STAND: entities.add((LivingEntity) cEntity); break;
+            }
+        }  
+        
+        Location location;
```
mastercoms: This location variable isn't needed anymore since you're accessing the player location directly.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
```diff
@@ -0,0 +1,50 @@
...
+import org.bukkit.entity.Player;
+import net.glowstone.inventory.GlowInventory;
+
+public class ArmorDispenseBehavior extends DefaultDispenseBehavior {
+    DefaultDispenseBehavior defaultBehavior = new DefaultDispenseBehavior();
+    
+    @Override
+    protected ItemStack dispenseStack(GlowBlock block, ItemStack stack) {
+        BlockFace facing = BlockDispenser.getFacing(block);
+        GlowBlock target = block.getRelative(facing);
```
mastercoms: It is a bit expensive to get a block, so try cloning `block`'s location and then adding facing.getModX,Y and Z to that location to get `targetLocation`.

mastercoms (changes requested):

src/main/java/net/glowstone/block/ItemTable.java
Christian7573: Oops, forgot. Thanks for the reminder.

src/main/java/net/glowstone/block/state/GlowDispenser.java
Christian7573: Oh ok.

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Sure

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Will be removed

src/main/java/net/glowstone/dispenser/ArmorDispenseBehavior.java
Christian7573: Ok

Christian7573: Wait, since `targetLocation` is already a block location do I need to use it there too?

mastercoms: Oh whoops, you're right. I wasn't thinking about that. Simple getX and similar are fine for locations of blocks.

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -57,30 +82,19 @@ public boolean dispense() {
             block.getWorld().playEffect(block.getLocation(), Effect.CLICK1, 0);
             return false;
         }
-
+
```
mastercoms, line 85: Extra whitespace

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -57,30 +82,19 @@ public boolean dispense() {
...
+        
         ItemStack origItems = getInventory().getItem(dispenseSlot);
 
         DispenseBehavior behavior = getDispenseBehaviorRegistry().getBehavior(origItems.getType());
         ItemStack result = behavior.dispense(block, origItems);
         getInventory().setItem(dispenseSlot, result);
         return true;
     }
-
+
```
mastercoms, line 93: Extra whitespace

src/main/java/net/glowstone/block/state/GlowDispenser.java
```diff
@@ -57,30 +82,19 @@ public boolean dispense() {
...
-            if (inv.getItem(i) != null && random.nextInt(randomChance++) == 0) {
-                slot = i;
-            }
-        }
-
-        return slot;
+        return InventoryUtil.getRandomSlot(random, getInventory(), true);
     }
-
+
```
mastercoms, line 97: Extra whitespace

mastercoms (approved):

mastercoms:
Hi, this looks good to go, there's just some extra whitespace, but that's no big deal. I can clean that up, just remember to clean up your whitespace next time please.

Christian7573:
Ok

Christian7573:
Ok

aramperes (approved):

mastercoms:

PR #473, opened 2017-05-14
Implement block breaking animation

satoshinm:
- [x] Add block break animation message and codec
- [x] Broadcast message to nearby clients
- [x] Calculate block break stage when breaking
- [x] Clear block break animation when stopped
- ~~factoring in tool effectiveness vs hardness~~ - out of scope of this pull request
- ~~BlockCrackEvent API event? (fire in between BlockDamageEvent and BlockBreakEvent)~~ - maybe later

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -2080,18 +2080,6 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
         }
     }
 
-    public void sendBlockBreakAnimation(Player player, Location loc, int destroyStage) {
-        sendBlockBreakAnimation(new BlockBreakAnimationMessage(player.getEntityId(), loc.getBlockX(), loc.getBlockY(), loc.getBlockZ(), destroyStage));
-    }
-
-    public void sendBlockBreakAnimation(BlockBreakAnimationMessage message) {
-        // only send message if the chunk is within visible range
-        Key key = new Key(message.getX() >> 4, message.getZ() >> 4);
-        if (canSeeChunk(key)) {
-            afterBlockChanges.add(message);
-        }
-    }
-
     @Override
     public boolean sendChunkChange(Location loc, int sx, int sy, int sz, byte[] data) {
         throw new UnsupportedOperationException("Not supported yet.");
@@ -2844,6 +2832,19 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
 
     public GlowBlock getDigging() {
         return digging;
+    }
+
+
+    public void sendBlockBreakAnimation(Player player, Location loc, int destroyStage) {
+        sendBlockBreakAnimation(new BlockBreakAnimationMessage(player.getEntityId(), loc.getBlockX(), loc.getBlockY(), loc.getBlockZ(), destroyStage));
+    }
+
+    public void sendBlockBreakAnimation(BlockBreakAnimationMessage message) {
+        // only send message if the chunk is within visible range
+        Key key = new Key(message.getX() >> 4, message.getZ() >> 4);
+        if (canSeeChunk(key)) {
+            afterBlockChanges.add(message);
+        }
     }
 
     private void broadcastBlockBreakAnimation(GlowBlock block, int destroyStage) {
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -2873,8 +2873,12 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
     }
 
     private void pulseDigging() {
-        // TODO: take into account the tool used to mine (ineffective=5x, effective=1.5x, material multiplier, etc.)
         float hardness = digging.getMaterialValues().getHardness() * 1000;
+
+        // TODO: take into account the tool used to mine (ineffective=5x, effective=1.5x, material multiplier, etc.)
+        // for now, assuming hands are used and the block is not dropped
+        hardness *= 5;
+
         long duration = System.currentTimeMillis() - diggingStarted;
         double completion = (double) duration / hardness;
         int stage = (int) (completion * 10);
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -2852,7 +2852,9 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
         GlowChunk.Key key = new GlowChunk.Key(block.getChunk().getX(), block.getChunk().getZ());
         // TODO: should canSeeChunk filtering take place here, in GlowPlayer, or both?
         block.getWorld().getRawPlayers().stream().filter(player1 -> player1.canSeeChunk(key)).forEach(player2 -> {
-            player2.sendBlockBreakAnimation(player2, block.getLocation(), destroyStage);
+            if (player2 != this) {
+                player2.sendBlockBreakAnimation(player2, block.getLocation(), destroyStage);
+            }
         });
     }
 
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -2838,8 +2838,22 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
         return digging;
     }
 
+    private void broadcastBlockBreakAnimation(GlowBlock block, int destroyStage) {
+        if (block == null) return;
+
+        GlowChunk.Key key = new GlowChunk.Key(block.getChunk().getX(), block.getChunk().getZ());
+        // TODO: should canSeeChunk filtering take place here, in GlowPlayer, or both?
+        block.getWorld().getRawPlayers().stream().filter(player1 -> player1.canSeeChunk(key)).forEach(player2 -> {
+            player2.sendBlockBreakAnimation(player2, block.getLocation(), destroyStage);
+        });
+    }
+
     public void setDigging(GlowBlock block) {
         digging = block;
+
+        // show other clients the block is beginning to crack
+        // TODO: show incremental stages
+        broadcastBlockBreakAnimation(block, 0);
     }
 
     @Override
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -284,9 +284,9 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
      */
     private GlowBlock digging;
     /**
-     * The time the player started digging.
+     * The number of ticks elapsed since the player started digging.
      */
-    private long diggingStarted = -1;
+    private long diggingTicks = 0;
 
     public Location teleportedTo = null;
     /**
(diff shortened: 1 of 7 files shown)


satoshinm:
Added the 0x08 codec, but that's the easy part, I'll have to dig into the code further to see how to send this message to nearby clients appropriately (or if anyone wants to lend me some pointers how this would be done?). Note this message allows clients to see the block break stages when other players are digging blocks.

aramperes:
You can use something like [this](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/block/entity/BlockEntity.java#L43) to update in range.

src/main/java/net/glowstone/net/message/play/game/BlockBreakAnimationMessage.java
```diff
@@ -0,0 +1,10 @@
+package net.glowstone.net.message.play.game;
+
+import com.flowpowered.network.Message;
+import lombok.Data;
+
+@Data
+public class BlockBreakAnimationMessage implements Message {
+
+    private final int entityID, x, y, z, destroyStage;
```
aramperes: For the sake of continuity with other entity Message classes, change `entityID` to `id`.

src/main/java/net/glowstone/net/codec/play/game/BlockBreakAnimationCodec.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public ByteBuf encode(ByteBuf buf, BlockBreakAnimationMessage message) throws IOException {
+        ByteBufUtils.writeVarInt(buf, message.getEntityID());
+        GlowBufUtils.writeBlockPosition(buf, message.getX(), message.getY(), message.getZ());
+        buf.writeByte(message.getDestroyStage());
+        return buf;
+    }
+}
+
```
aramperes: Extra whitespace.

aramperes (changes requested):

src/main/java/net/glowstone/net/codec/play/game/BlockBreakAnimationCodec.java
mastercoms: Is that not a newline at the end of file?

src/main/java/net/glowstone/net/message/play/game/BlockBreakAnimationMessage.java
```diff
@@ -0,0 +1,10 @@
...
+}
```
mastercoms: Add a newline here please.

mastercoms (changes requested):

src/main/java/net/glowstone/net/codec/play/game/BlockBreakAnimationCodec.java
satoshinm: It's an extra newline, I'll remove it *update: removed*

mastercoms:
Ah, I thought GitHub was supposed to show newlines.

satoshinm:
Added an initial block break animation

satoshinm:
screen shot 2017-05-15 at 9 44 07 pm

Added an initial block break animation broadcast, other clients can now see the initial block break stage. However, the digging handler isn't continuously called, only when digging starts/stops. I think to implement this properly, the server would need to know each blocks' hardness to calculate the stages, then increment accordingly. Looking at another open source server (Cuberite) it seems they haven't implemented this part yet either: https://github.com/cuberite/cuberite/blob/01d9beee9ee60aa1ead52f7796c762a8372b9ca4/src/ClientHandle.cpp#L1223-L1231 "TODO: Send animation packets even without receiving any other packets" edit: actually, they update the stage in a "client tick" method: https://github.com/cuberite/cuberite/blob/01d9beee9ee60aa1ead52f7796c762a8372b9ca4/src/ClientHandle.cpp#L2134-L2147 - does Glowstone have something equivalent to this?

mastercoms:
Yeah but I don't think it is good to have it in a client tick. Digging is caused by the packet, and so any tasks should off shoot from there. 

About hardness, that info is stored in the material values collection.

satoshinm:
> Yeah but I don't think it is good to have it in a client tick. Digging is caused by the packet, and so any tasks should off shoot from there.

Start an asynchronous task if `message.getState() == DiggingMessage.START_DIGGING`, sending increasingly block break destroy stage packets at an interval, then cancel the task when the block breaks or digging stops (`message.getState() == DiggingMessage.FINISH_DIGGING`)?

[materialValues.yml](https://github.com/GlowstoneMC/Glowstone/blob/77e30f78e4e323ae61a999c7dad3075731985f77/src/main/resources/builtin/materialValues.yml) has a `hardness` field but only values used are: 1 (default), -1 (bedrock), 0 (soft blocks, plants etc.), and 50 (bedrock).

Hardness values to populate the yaml can be found at: http://minecraft.gamepedia.com/Breaking#Blocks_by_hardness

mastercoms:
This PR would probably depend on [#379](https://github.com/GlowstoneMC/Glowstone/pull/379/files#diff-9f83ce130d9b3bff618cb81945f17116R14), then.

About the async task, I think it could work like that.

satoshinm:
Hm, didn't see https://github.com/GlowstoneMC/Glowstone/pull/379/files#diff-9f83ce130d9b3bff618cb81945f17116R14 .. are the hardnesses moving to BuiltinMaterialValueManager.java out of materialValues.yml? Went ahead and added to .yml before I saw this, http://minecraft.gamepedia.com/Breaking#Blocks_by_hardness has all the hardness values for reference except skulls but http://minecraft.gamepedia.com/Mob_head says skull hardness = 1.

mastercoms:
Yeah they're being moved to the class so it can better support custom materials and programmable modifications of vanilla materials.

However, your latest commit should be fine. It removes the need to wait for #379 to be merged first.

src/main/java/net/glowstone/net/message/play/game/BlockBreakAnimationMessage.java
satoshinm: Added newline after the variable declarations

src/main/java/net/glowstone/net/message/play/game/BlockBreakAnimationMessage.java
satoshinm: Changed to id

satoshinm:
Addressed the requested changes but they still show as changes requested, how can I mark them addressed?

Implemented the block break animation stages, used player pulse() seemed to make the most sense there. Added cancel digging support. This is functional now however, I'm finding the hardness differs from what the player sees:

screen shot 2017-05-21 at 8 06 07 pm

left = the player digging, right = the block break animation from another client (too fast).

It isn't taking into account the tool/effectiveness versus hardness. From http://minecraft.gamepedia.com/Breaking#Blocks_by_hardness - stone "hardness" is 1.5 seconds, but when mined with a bare hand it is 7.5 seconds. The equation is described in http://minecraft.gamepedia.com/Breaking#Speed:

> The base time in seconds is the block's hardness multiplied by 1.5 if the player can harvest the block with the current tool, or multiplied by 5 if the player cannot.

mastercoms:
The reviewer has to approve changes. Nothing you need to do.

Due to the complexity of this issue, I believe it is best to leave it for another PR. Initial block break animation support is sufficient, in my opinion.

mastercoms (approved):

satoshinm:
Agreed, sounds good. I changed it for now to use the correct hardness for if the block is mined with bare hands and is not dropped, for example, hands on stone -- the animations now match. The consequences of incorrect block break animation are purely cosmetic, tool effectiveness can be factored in later, I think this is good enough for now.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -2826,10 +2846,47 @@ public GlowBlock getDigging() {
...
     }
 
+    private void pulseDigging() {
+        float hardness = digging.getMaterialValues().getHardness() * 1000;
+
+        // TODO: take into account the tool used to mine (ineffective=5x, effective=1.5x, material multiplier, etc.)
+        // for now, assuming hands are used and the block is not dropped
+        hardness *= 5;
+
+        long duration = System.currentTimeMillis() - diggingStarted;
```
aramperes: Wouldn't it be better to base this off world thread ticking, instead of system time in milliseconds?

satoshinm: Good point, will fix *update*: changed to count ticks

aramperes (approved):

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -2826,10 +2846,47 @@ public GlowBlock getDigging() {
         return digging;
     }
 
+    private void broadcastBlockBreakAnimation(GlowBlock block, int destroyStage) {
+        if (block == null) return;
```
mastercoms: Wouldn't this work better as a precondition?

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -2826,10 +2846,47 @@ public GlowBlock getDigging() {
...
+
+        GlowChunk.Key key = new GlowChunk.Key(block.getChunk().getX(), block.getChunk().getZ());
+        // TODO: should canSeeChunk filtering take place here, in GlowPlayer, or both?
+        block.getWorld().getRawPlayers().stream().filter(player1 -> player1.canSeeChunk(key)).forEach(player2 -> {
```
mastercoms: Could you explain to me what this code is doing?

mastercoms (changes requested):

src/main/java/net/glowstone/entity/GlowPlayer.java
satoshinm: Agreed, changed to move the check outside the function - it should only be called with non-null block now

src/main/java/net/glowstone/entity/GlowPlayer.java
satoshinm: It is getting the chunk of the block and filtering the list of players to those who can see the chunk, so it only sends the block breaking animation to the right players. I wasn't familiar with the Streams API and Java lambdas but it seems to be a common pattern in Glowstone:

```
src/main/java/net/glowstone/block/entity/BlockEntity.java
45:        block.getWorld().getRawPlayers().stream().filter(player -> player.canSeeChunk(key)).forEach(this::update);

src/main/java/net/glowstone/block/state/GlowNoteBlock.java
189:        getWorld().getRawPlayers().stream().filter(player -> player.canSeeChunk(key)).forEach(player -> player.playNote(location, instrument, note));

src/main/java/net/glowstone/entity/GlowEntity.java
1024:        world.getRawPlayers().stream().filter(player -> player.canSeeEntity(this)).forEach(player -> player.getSession().send(message));

src/main/java/net/glowstone/entity/objects/GlowItem.java
71:            world.getRawPlayers().stream().filter(other -> other.canSeeEntity(this)).forEach(other -> other.getSession().send(message));
```

but the players are filtered twice for chunk visibility, both here and in sendBlockBreakAnimation:

```java
    public void sendBlockBreakAnimation(BlockBreakAnimationMessage message) {
        // only send message if the chunk is within visible range
        Key key = new Key(message.getX() >> 4, message.getZ() >> 4);
        if (canSeeChunk(key)) {
            afterBlockChanges.add(message);
        }
    }
```

leaning towards removing this second check and only performing the filtering in broadcastBlockBreakAnimation **update**: changed to filter only once

mastercoms: Ok. Why did you name your variables player1 and player2? Can't they both be player?

Also you have an if statement within the for-each lambda that could be part of the filter predicate.

mastercoms (changes requested):

mastercoms:
@satoshinm GitHub is wrongly hiding my comment here: https://github.com/GlowstoneMC/Glowstone/pull/473#discussion_r117816322

I'm not sure if you saw it because of this.

satoshinm:
@mastercoms thanks didn't see it, will fix *update*: made the changes

src/main/java/net/glowstone/entity/GlowPlayer.java
satoshinm: Made both changes, also removed unnecessary Player parameter

mastercoms (approved):

mastercoms:
Alright, looks good. Thank you for your contribution to Glowstone.

mastercoms:
By the way, did you see our invite at https://github.com/GlowstoneMC?

(As per our contribution policy: https://github.com/GlowstoneMC/Glowstone/blob/master/.github/CONTRIBUTING.md)

Issue #478, opened 2017-05-25
Jack-o'-lanterns always place facing north


satoshinm:
When a jack-o'-lantern block is placed, the face always faces north (as if the player is looking north, the face towards the opposite, southern direction). Pumpkins do not have this problem, and correctly place facing the direction the player is facing.

Glowstone 496

PR #479, opened 2017-05-25
Fix jack-o'-lantern orientation on placement. Closes GH-478

satoshinm:
Move the Pumpkin interface facing logic up to a BlockPumpkinBase class
shared by both BlockPumpkin and BlockJackOLantern. Placed
jack-o'-lanterns now face the player, just as pumpkins face the player.


src/main/java/net/glowstone/block/ItemTable.java
```diff
@@ -205,7 +205,7 @@ private void registerBuiltins() {
         reg(Material.TNT, new BlockTNT());
         reg(Material.DOUBLE_PLANT, new BlockDoublePlant());
         reg(Material.PUMPKIN, new BlockPumpkin());
-        reg(Material.JACK_O_LANTERN, new BlockDirectDrops(Material.JACK_O_LANTERN));
+        reg(Material.JACK_O_LANTERN, new BlockJackOLantern());
```
mastercoms: Why create a new class instead of just using the constructor for BlockPumpkinBase?

mastercoms (changes requested):

src/main/java/net/glowstone/block/blocktype/BlockPumpkinBase.java
```diff
@@ -0,0 +1,30 @@
...
+import net.glowstone.entity.GlowPlayer;
+import net.glowstone.inventory.ToolType;
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.material.MaterialData;
+import org.bukkit.material.Pumpkin;
+import org.bukkit.util.Vector;
+
+public abstract class BlockPumpkinBase extends BlockDirectDrops {
```
mastercoms: I don't think this class needs to be abstract.

mastercoms (changes requested):

src/main/java/net/glowstone/block/ItemTable.java
satoshinm: I guess I could change it to `new BlockPumpkinBase(Material.JACK_O_LANTERN)`

src/main/java/net/glowstone/block/blocktype/BlockPumpkinBase.java
satoshinm: Removed abstract (was needed in an earlier iteration when I was changing the constructors but no longer necessary)

mastercoms (approved):

mastercoms:
Looks good. Thanks for your contribution to Glowstone!

I'm a bit conflicted about merging this right now, since for thematic reasons it would be best to wait until Halloween.

PR #480, opened 2017-05-27
Use IdleStateHandler for both ReadTimeoutHandler and WriteTimeoutHandler

louxiu:
`IdleStateHandler` has the function of `ReadTimeoutHandler`. So no extra `ReadTimeoutHandler` is needed


mastercoms (changes requested):
Thank you for your contribution to Glowstone! This PR looks good to me, but I think while we're modifying this, we can also move the idle state handler up to be the first one added to the pipeline.

louxiu:
@mastercoms That make sense. Updated

mastercoms (approved):

mastercoms:
Alright, thank you for your contribution to Glowstone.

Issue #486, opened 2017-06-01
NPE when picking up crafting result on workbench


satoshinm:
Place a crafting table, right-click to open it, place items to craft something (I used 8 planks in a square to make a chest), then click on the crafting output to pick it up. Glowstone build 520 errors out:

```
23:55:28 [SEVERE] Error while handling WindowClickMessage(id=1, slot=0, button=0, transaction=12, mode=0, item=ItemStack{CHEST x 1}) (handler: WindowClickHandler)
java.lang.NullPointerException
	at net.glowstone.inventory.crafting.CraftingManager.removeItems(CraftingManager.java:96)
	at net.glowstone.inventory.GlowCraftingInventory.craft(GlowCraftingInventory.java:96)
	at net.glowstone.net.handler.play.inv.WindowClickHandler.process(WindowClickHandler.java:400)
	at net.glowstone.net.handler.play.inv.WindowClickHandler.handle(WindowClickHandler.java:30)
	at net.glowstone.net.handler.play.inv.WindowClickHandler.handle(WindowClickHandler.java:25)
	at com.flowpowered.network.session.BasicSession.handleMessage(BasicSession.java:80)
	at com.flowpowered.network.session.BasicSession.messageReceived(BasicSession.java:139)
	at net.glowstone.net.GlowSession.pulse(GlowSession.java:472)
	at java.util.concurrent.ConcurrentHashMap$KeySetView.forEach(ConcurrentHashMap.java:4649)
	at net.glowstone.net.SessionRegistry.pulse(SessionRegistry.java:23)
	at net.glowstone.scheduler.GlowScheduler.pulse(GlowScheduler.java:144)
	at net.glowstone.scheduler.GlowScheduler.lambda$start$0(GlowScheduler.java:83)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #487, opened 2017-06-01
Fix inverted condition causing NPE crafting. Closes GH-486

satoshinm:
Fixes https://github.com/GlowstoneMC/Glowstone/issues/486


aramperes (approved):
Seems like it was a mistake when switching from `null` checks to `isEmpty()`.

Issue #490, opened 2017-06-07
Server fails to start after initial worldgen, NPE at net.glowstone.GlowWorld.prepareSpawn(GlowWorld.java:796)


satoshinm:
Since build 538 with, Glowstone server is able to generate a fresh initial world but subsequent invocations fail with a null pointer exception and stop the server. If I revert https://github.com/GlowstoneMC/Glowstone/commit/f5abedc5d720dbd4c6f3cb8e1a1a711516e0fa69 I am able to start the server successfully. Full console log:

```
tmp $ java -jar /Users/admin/Downloads/glowstone.jar 
23:37:07 [INFO] This server is running Glowstone version 2017.6.0-SNAPSHOT.f5abedc-MC1.11.2 (Implementing API version 1.11.2-R0.1-SNAPSHOT)
23:37:07 [INFO] Creating default config: config/glowstone.yml
23:37:07 [INFO] Creating default world config: config/worlds.yml
23:37:07 [INFO] Recipes: 299 shaped, 63 shapeless, 26 furnace, 10 dynamic, 28 fuels.
23:37:07 [INFO] Downloading sqlite-jdbc 3.15.1...
23:37:07 [INFO] Downloading slf4j-jdk14 1.7.15...
23:37:07 [INFO] Downloading mysql-connector-java 5.1.39...
23:37:10 [INFO] Downloaded slf4j-jdk14 1.7.15.
23:37:10 [INFO] Downloaded mysql-connector-java 5.1.39.
23:37:10 [INFO] Downloaded sqlite-jdbc 3.15.1.
23:37:10 [INFO] Scanning plugins...
23:37:10 [INFO] Created default config: config/help.yml
23:37:10 [INFO] Preparing spawn for world...
23:37:12 [INFO] Preparing spawn for world: 0%
23:37:13 [INFO] Preparing spawn for world: 3%
23:37:14 [INFO] Preparing spawn for world: 10%
23:37:15 [INFO] Preparing spawn for world: 24%
23:37:16 [INFO] Preparing spawn for world: 38%
23:37:17 [INFO] Preparing spawn for world: 52%
23:37:18 [INFO] Preparing spawn for world: 70%
23:37:19 [INFO] Preparing spawn for world: 92%
23:37:19 [INFO] Preparing spawn for world: done
23:37:19 [INFO] Preparing spawn for world_nether...
23:37:20 [INFO] Preparing spawn for world_nether: 43%
[...]

PR #492, opened 2017-06-08
Revert "Enhance prepare spawn code". Closes GH-490

satoshinm:
Fixes https://github.com/GlowstoneMC/Glowstone/issues/490, allowing the server to start 

Didn't look into too deeply why it was failing to develop a more targeted fix, but reverting works


aramperes:
Implemented a fix for the issue in 8b3d13e32acb707f7db808b628d22840cef760a1. A revert shouldn't be needed.

PR #509, opened 2017-06-27
Make explosions visible

### LINKED ISSUE #508: ExplosionMessage is only sent to players that get damaged
Hi,

Did some more testing. The explosion particles are not visible when you are not in the range of the explosion.
How to reproduce:
1) Place a tnt block
2) ignite it
3) Go a few block (20 or so) away

In vanilla, the explosion is visible from far away.

In glowstone, the explosion is not visible.
However, when you get damaged by the explosion, you will see the particles.

Version:
`
This server is running Glowstone version 2017.6.1-SNAPSHOT.0826543-MC1.12 (Implementing API version 1.12-R0.1-SNAPSHOT)
`

best regards
Postremus

Postremus:
After some testing in vanilla, I figured out that explosions are visible in a range of 64 blocks.  

Also, separate damage and visibility calculations.
closes #508

### DIFF
--- a/src/main/java/net/glowstone/Explosion.java
+++ b/src/main/java/net/glowstone/Explosion.java
@@ -259,10 +259,8 @@ public final class Explosion {
             }
             entity.damage(damage, source, damageCause);
 
-            if (entity instanceof GlowPlayer) {
-                if (((GlowPlayer) entity).isFlying()) {
-                    continue;
-                }
+            if (entity instanceof GlowPlayer && ((GlowPlayer) entity).isFlying()) {
+                continue;
             }
 
             Vector rayLength = RayUtil.getVelocityRay(distanceToHead(entity));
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/Explosion.java
```diff
@@ -255,7 +260,6 @@ private void setBlockOnFire(GlowBlock block) {
             entity.damage(damage, source, damageCause);
 
             if (entity instanceof GlowPlayer) {
```
mastercoms: Merge the two if statements, please

mastercoms (changes requested):

gdude2002 (approved):
:+1:

src/main/java/net/glowstone/Explosion.java
Postremus: Will do.
Actually, you pointed another possible error with explosions out.
Flying players will, at least in paper 1.10, take damage from explosions.
However, this will need more testing -> another pr.

mastercoms (approved):

gdude2002:
Thanks for the contribution!

PR #513, opened 2017-07-01
respawn on ground

### LINKED ISSUE #460: [#458] Server places you above the map when respawning
This issue has been posted to separate the following ticket into smaller issues: #458

Postremus:
This pr attempts to fix the sending order of the position and chunkdata, as well as some other minor stuff.

Currently the problem is, we send chunkdata before the position update on respawn. The client gets confused, and uses `x=8.5 y=? z=8.5` as its coordinates. These coordinates will also be send as first position update from the client. x and z appear to be static, y is somewhere on ground.
After this initial confusion, the server sends the real position of the player. The client accepts these coordinates correctly through a TeleportConfirmMessage.
However, when his first coordinates where above the real spawn location, he will fall to the ground, causing issue #460.

I adjusted the sending order to be (hopefully) vanilla like.

The same problem occurs on inital join in a freshly generated world, but vanilla does not handle this case as well -> no need to change something here.

I adjusted the sending order of the SPAWN_POSITION, TIME and WORLD_BORDER message for both respawn and join.

Logs used to figure this out:
join on paper: https://gist.github.com/Postremus/9ea304d277ec2f1850fded5252c158ef
join on glowstone: https://gist.github.com/Postremus/b8c590817b3a18ab3ce46df92f316a0f

/kill (respawn) on paper: https://gist.github.com/Postremus/e8f0e9c924b3e73e8199f80592cee020
/kill (respawn) on glowstone: https://gist.github.com/Postremus/57066a2bea42338d4b048bd3147fc42b

closes #460

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -355,7 +355,45 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
                 return loc;
             }
         }
-        return session.getServer().getWorlds().get(0).getSpawnLocation();
+
+        return findSafeSpawnLocation(session.getServer().getWorlds().get(0).getSpawnLocation());
+    }
+
+    /**
+     * Find a a Location obove or below the specified Location, which
+     * is on ground.
+     * The returned Location will be at the center of the block, X and Y wise.
+     *
+     * @param spawn The Location a safe spawn position should be found at.
+     * @return The location to spawn the player at.
+     */
+    private static Location findSafeSpawnLocation(Location spawn) {
+        World world = spawn.getWorld();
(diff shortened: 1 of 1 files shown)


aramperes (changes requested):
From my testing with these changes, it looks like the player spawns at a specific Y-coordinate -- if you block this position with blocks, it can cause suffocation when spawning.

Wouldn't it be better if the player spawned at a safe place instead of inside the ground? I'm not sure about this behavior in Vanilla though. If this is consistent with how the game does it, then that's okay.

sc

Also, I think the player should spawn at X+0.5 and Z+0.5 (center of the block), instead of X.00 and Z.00

aramperes (approved):

aramperes:

PR #517, opened 2017-07-02
Don't fake explode blocks

### LINKED ISSUE #515: Explosions fake destroy wrong blocks
Hello,

Explosions seem to be able to "fake" destroy blocks, like obsidian. The server sends that the block got destroyed in the ExplosionMessage. After a rejoin, the obsidian will be there again.

Hot to reproduce:
1) Build something like this:
grafik
2) Let it explode
3) The Obsidian blocks seems to get destroyed
4) disconnect and connect to the server again
5) The Obsidian block is there again.

regards
Postremus

Postremus:
This PR attempts to fix multiple explosion related problems.

The offset of records in ExplosionMessage's was wrong. Changed it according to http://wiki.vg/Protocol#Explosion.

The destroyed blocks of an explosion can be changed via the API.

Also, only calculate the vectors for the rays once.

closes #515

### DIFF
--- a/src/main/java/net/glowstone/Explosion.java
+++ b/src/main/java/net/glowstone/Explosion.java
@@ -47,7 +47,6 @@ public final class Explosion {
     private static final List<Vector> RAY_DIRECTIONS = new ArrayList<>();
 
     static {
-
         int value = 16;
         for (int x = 0; x < value; x++) {
             for (int y = 0; y < value; y++) {
@@ -335,9 +334,10 @@ public final class Explosion {
         Collection<Record> records = new ArrayList<>();
 
         for (Block block : blocks) {
-            byte x = (byte) (block.getLocation().getBlockX() - (int) location.getX());
-            byte y = (byte) (block.getLocation().getBlockY() - (int) location.getY());
-            byte z = (byte) (block.getLocation().getBlockZ() - (int) location.getZ());
+            Location blockLocation = block.getLocation();
+            byte x = (byte) (blockLocation.getBlockX() - (int) this.location.getX());
+            byte y = (byte) (blockLocation.getBlockY() - (int) this.location.getY());
+            byte z = (byte) (blockLocation.getBlockZ() - (int) this.location.getZ());
             records.add(new Record(x, y, z));
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/Explosion.java
```diff
@@ -327,13 +331,13 @@ private void playOutSoundAndParticles() {
...
-    private void playOutExplosion(GlowPlayer player, Iterable<BlockVector> blocks) {
+    private void playOutExplosion(GlowPlayer player, Iterable<Block> blocks) {
         Collection<Record> records = new ArrayList<>();
 
-        for (BlockVector block : blocks) {
-            byte x = (byte) (block.getBlockX() - location.getBlockX());
-            byte y = (byte) (block.getBlockY() - location.getBlockY());
-            byte z = (byte) (block.getBlockZ() - location.getBlockZ());
+        for (Block block : blocks) {
+            byte x = (byte) (block.getLocation().getBlockX() - (int) location.getX());
```
aramperes: Is there a reason to use `(int) location.getX()` instead of `location.getBlockX()`? Also, please cache the `block.getLocation()` object because the method creates new Location instances each time.

src/main/java/net/glowstone/Explosion.java
```diff
@@ -44,6 +44,29 @@
     public static final int POWER_ENDER_CRYSTAL = 6;
     private static final Random random = new Random();
     public static final int EXPLOSION_VISIBILITY_RADIUS = 64;
+    private static final List<Vector> RAY_DIRECTIONS = new ArrayList<>();
+
+    static {
+
```
aramperes: Extra whitespace.

aramperes (changes requested):

src/main/java/net/glowstone/Explosion.java
Postremus: As example, lets take X = -103.351. When casting to int, we get -103. The .getBlockX() returns, as expected, -104.
The client apparently needs the offsets to be calculated from the casted location.

aramperes (approved):

aramperes:

Issue #499, opened 2017-06-23
Implementing vanilla commands


aramperes:
Most vanilla commands were [removed](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/a1cc4b769caf690b2f075f833c83e76d6431f9a7) from Bukkit as part of the 1.12 update. Therefore, we need to re-implement them ourselves to get them back in Glowstone.

This is a list of commands that need to be implemented:

- [x] ~~`/achievement`~~ (removed from Minecraft, not needed)
- [ ] `/advancement` (Advancements need to be implemented first)
- [x] `/ban`
- [x] `/ban-ip`
- [x] `/banlist` (#518)
- [ ] `/blockdata`
- [x] **`/clear`**
- [x] `/clone` (#590)
- [x] `/defaultgamemode` (#521)
- [x] `/deop`
- [x] **`/difficulty`**
- [ ] `/effect`
- [x] `/enchant` (#560)
- [ ] `/entitydata`
- [ ] `/execute`
- [ ] `/fill`
- [x] `/function`
- [x] `/gamemode`
- [x] `/gamerule`
- [x] `/give`
- [x] `/kick`
- [x] **`/kill`**
- [x] `/list`
- [ ] `/locate`
- [x] `/me` (#521)
- [x] `/op`
- [x] **`/pardon`**
- [x] **`/pardon-ip`**
- [ ] `/particle`
- [x] `/playsound` (#531)
- [ ] `/recipe` (New recipes system needs to be implemented, might wait for 1.13+)
- [ ] `/replaceitem`
- [x] **`/save-all`, `/save-on`, `/save-off`**
- [x] `/say`
- [ ] `/scoreboard`
- [x] `/seed` (#521)
- [x] `/setidletimeout` (#523)
- [x] `/setblock` (6d00594775acc96f88d21f4ac8ad7ba54b8c0fad)
- [x] `/setworldspawn` (#523)
- [x] `/spawnpoint` (#523)
- [ ] `/spreadplayers`
- [x] `/stop`
- [x] **`/tp`**
- [x] `/tell`
- [x] `/testfor` (#573)
- [x] `/testforblock`
- [x] `/testforblocks`
- [x] **`/time`**
- [x] `/toggledownfall` (#523)
- [ ] `/trigger`
- [x] **`/weather`**
- [x] **`/whitelist`**
- [x] `/xp` (#521)

[...]

PR #518, opened 2017-07-05
#499 - Adds banlist command

FlorentClarret:
Hi,

This PR relates to this issues : https://github.com/GlowstoneMC/Glowstone/issues/499. It doesn't close it, but contributes to :)

It reimplements the vanilla banlist command line.

### DIFF
--- a/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
@@ -7,7 +7,6 @@ import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.defaults.VanillaCommand;
-import org.bukkit.util.StringUtil;
 
 import java.util.*;
 import java.util.stream.Collectors;
--- a/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
@@ -7,6 +7,7 @@ import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.defaults.VanillaCommand;
+import org.bukkit.util.StringUtil;
 
 import java.util.*;
 import java.util.stream.Collectors;
@@ -54,6 +55,10 @@ public class BanListCommand extends VanillaCommand {
 
     @Override
     public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
-        return args.length == 0 ? super.tabComplete(sender, alias, args) : Collections.emptyList();
+        if (args.length == 1) {
+            return StringUtil.copyPartialMatches(args[0], BAN_TYPES, new ArrayList(BAN_TYPES.size()));
+        } else {
+            return args.length == 0 ? super.tabComplete(sender, alias, args) : Collections.emptyList();
+        }
     }
 }
--- a/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
@@ -1,0 +1,86 @@
+package net.glowstone.command.minecraft;
+
+import org.bukkit.BanEntry;
+import org.bukkit.BanList;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.command.CommandSender;
+import org.bukkit.command.defaults.VanillaCommand;
+import org.bukkit.util.StringUtil;
+
+import java.util.*;
+
+public class BanListCommand extends VanillaCommand {
+
+    private static final List<String> BAN_TYPES;
+
+    private static final int BAN_TYPES_SIZE;
+
+    static {
+        BAN_TYPES = Arrays.asList("ips", "players");
+        BAN_TYPES_SIZE = BAN_TYPES.size();
+    }
+
+    public BanListCommand() {
+        super("banlist", "Displays the server's blacklist.", "/banlist [ips|players]", Collections.emptyList());
+        setPermission("minecraft.command.ban.list");
+    }
+
+    @Override
+    public boolean execute(CommandSender sender, String commandLabel, String[] args) {
+        if (!testPermission(sender)) return false;
+
+        BanList.Type banType;
+
+        if (args.length > 0) {
+            final String parameter = args[0];
+            switch (parameter) {
+                case "ips" :
+                    banType = BanList.Type.IP;
+                    break;
+                case "players" :
+                    banType = BanList.Type.NAME;
+                    break;
+                default:
+                    sender.sendMessage(ChatColor.RED + "Invalid parameter '" + parameter + "'. Usage : " + usageMessage);
+                    return false;
+            }
+        } else {
+             banType = BanList.Type.NAME;
+        }
+
+        final Set<BanEntry> banEntries = Bukkit.getBanList(banType).getBanEntries();
+
+        if (banEntries.isEmpty()) {
+            sender.sendMessage("There are no banned players");
+        } else {
+            int index = 0, size = banEntries.size();
+            final StringBuilder banList = new StringBuilder(200);
+
+            for (final BanEntry banEntry : banEntries) {
+                banList.append(banEntry.getTarget());
+                if (index != size - 1) {
+                    if (index == size - 2) {
+                        banList.append(" and ");
+                    } else {
+                        banList.append(", ");
+                    }
+                }
+                ++index;
+            }
+
+            sender.sendMessage("There are " + size + " banned players : ");
+            sender.sendMessage(banList.toString());
+        }
+
+        return true;
+    }
+
+    @Override
+    public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
+        if (args.length == 1) {
+            return StringUtil.copyPartialMatches(args[0], BAN_TYPES, new ArrayList(BAN_TYPES_SIZE));
+        }
+        return super.tabComplete(sender, alias, args);
+    }
+}
--- a/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/BanListCommand.java
@@ -55,10 +55,6 @@ public class BanListCommand extends VanillaCommand {
 
     @Override
     public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+
+        return true;
+    }
+
+    @Override
+    public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
+        if (args.length == 1) {
+            return StringUtil.copyPartialMatches(args[0], BAN_TYPES, new ArrayList(BAN_TYPES_SIZE));
+        }
+        return super.tabComplete(sender, alias, args);
```
aramperes: I think this should return `Collections.emptyList()` if `args.length` is over 1 (this will remove player auto-completion, which is not necessary here)

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+
+        final Set<BanEntry> banEntries = Bukkit.getBanList(banType).getBanEntries();
+
+        if (banEntries.isEmpty()) {
+            sender.sendMessage("There are no banned players");
+        } else {
+            int index = 0, size = banEntries.size();
+            final StringBuilder banList = new StringBuilder(200);
+
+            for (final BanEntry banEntry : banEntries) {
```
aramperes: Please use https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/command/CommandUtils.java#L105 if possible

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+            final String parameter = args[0];
+            switch (parameter) {
+                case "ips" :
+                    banType = BanList.Type.IP;
+                    break;
+                case "players" :
+                    banType = BanList.Type.NAME;
+                    break;
+                default:
+                    sender.sendMessage(ChatColor.RED + "Invalid parameter '" + parameter + "'. Usage : " + usageMessage);
```
aramperes: Remove space before `:` (`"Usage : "`)

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.*;
+
+public class BanListCommand extends VanillaCommand {
+
+    private static final List<String> BAN_TYPES;
+
+    private static final int BAN_TYPES_SIZE;
+
+    static {
```
aramperes: Is there a reason for a static initialization block here? I think you could just to:

```java
private static final List<String> BAN_TYPES = Arrays.asList("ips", "players");
```

No need to store `BAN_TYPES_SIZE` in a variable, just use `BAN_TYPES.size()` when needed.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+             banType = BanList.Type.NAME;
+        }
+
+        final Set<BanEntry> banEntries = Bukkit.getBanList(banType).getBanEntries();
+
+        if (banEntries.isEmpty()) {
+            sender.sendMessage("There are no banned players");
+        } else {
+            int index = 0, size = banEntries.size();
+            final StringBuilder banList = new StringBuilder(200);
```
aramperes: I don't think there's a need for an initial capacity value here?

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+                    if (index == size - 2) {
+                        banList.append(" and ");
+                    } else {
+                        banList.append(", ");
+                    }
+                }
+                ++index;
+            }
+
+            sender.sendMessage("There are " + size + " banned players : ");
```
aramperes: Remove space before `:` (`" banned players : "`)

aramperes (changes requested):

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public boolean execute(CommandSender sender, String commandLabel, String[] args) {
+        if (!testPermission(sender)) return false;
+
+        BanList.Type banType;
+
+        if (args.length > 0) {
+            final String parameter = args[0];
+            switch (parameter) {
```
LuminousLoreBloom: Make this non case-sensitive?

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: Good point, no need to store it in this case.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: No problem

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: French guy here, don't blame me.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: Indeed. No need in this case.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: Oh. Didn't see that one. Thanks

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: French guy here again, don't blame me.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: Fair point.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -33,17 +28,13 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
...
-                case "ips" :
-                    banType = BanList.Type.IP;
-                    break;
-                case "players" :
-                    banType = BanList.Type.NAME;
-                    break;
-                default:
-                    sender.sendMessage(ChatColor.RED + "Invalid parameter '" + parameter + "'. Usage : " + usageMessage);
-                    return false;
+            if ("ips".equalsIgnoreCase(args[0])) {
```
picknchew: A switch can still be used. Just make the first argument lowercase.

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        final Set<BanEntry> banEntries = Bukkit.getBanList(banType).getBanEntries();
+
+        if (banEntries.isEmpty()) {
+            sender.sendMessage("There are no banned players");
+        } else {
+            final List<String> targets = banEntries.stream().map(BanEntry::getTarget).collect(Collectors.toList());
+            sender.sendMessage("There are " + banEntries.size() + " banned players: ");
+            sender.sendMessage(CommandUtils.prettyPrint(targets.toArray(new String[targets.size()])));
```
LuminousLoreBloom, line 50: Perhaps check if there are no banned players here so an empty message isn't sent?

aramperes: @Mystiflow That check is done here: https://github.com/GlowstoneMC/Glowstone/pull/518/files/946cda8cadfbdf72f3a29d6a4fae84f80276bb56..779d13ed59cef62cd218cdc6d6fed30cb814e40f#diff-d9df5a37e1afb33765f601eec8d00d0aL55

FlorentClarret: Yup, the check is done

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: Yes, less modifications would have been performed

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
LuminousLoreBloom: Ah, whoops, completely skimmed over it

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -0,0 +1,64 @@
...
+
+        return true;
+    }
+
+    @Override
+    public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
+        if (args.length == 1) {
+            return StringUtil.copyPartialMatches(args[0], BAN_TYPES, new ArrayList(BAN_TYPES.size()));
+        } else {
+            return Collections.emptyList();
```
aramperes: You should still return `super.[...]` if `args.length == 0` (sorry if I wasn't clear in my previous review). For example: `return args.length == 0 ? super.[...] : Collections.emptyList();`

aramperes (changes requested):

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
FlorentClarret: My bad. I updated it too quickly

src/main/java/net/glowstone/command/minecraft/BanListCommand.java
```diff
@@ -55,10 +55,6 @@ public boolean execute(CommandSender sender, String commandLabel, String[] args)
 
     @Override
     public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
-        if (args.length == 1) {
```
aramperes: No... you still need that check :)

```java
if (args.length == 1) {
    return StringUtil.copyPartialMatches(args[0], BAN_TYPES, new ArrayList(BAN_TYPES.size()));
}
return args.length == 0 ? super.tabComplete(sender, alias, args) : Collections.emptyList();
```

aramperes (changes requested):

aramperes (approved):

aramperes:
Thank you for your contribution! :)

Issue #519, opened 2017-07-05
Named sound id are wrong in some cases


Postremus:
Hello,

I wanted to implement equipment sounds for armor stand (see #471).
"Sound.ITEM_ARMOR_EQUIP_LEATHER" maps to id "minecraft:item.armor.equip.leather". According to http://pokechu22.github.io/Burger/1.11.html#sounds however, it should be "minecraft:item.armor.equip_leather".

The same is the case for multiple, if not all, of the armor_equip sounds.

regards
Postremus

aramperes:
https://github.com/GlowstoneMC/Glowstone/blob/c07c74e068ce95a2cb33918360d9e630b0777465/src/main/java/net/glowstone/constants/GlowSound.java#L36

We should either have an actual map, or adjust this way of converting to a Vanilla ID.

mastercoms:
I honestly think the sounds map is useless for 1.11+, so that should be removed as well. Instead, there should be a map of sound ID exceptions.

PR #520, opened 2017-07-06
Add off hand support for Armor stands

### LINKED ISSUE #471: Armor Stands aren't functional
Armor stands just don't work.

As I was programming in #470, I was having extreme trouble getting armor stands to work. A thought occurred to me that perhaps armor stands were broken in general. I placed one down, grabbed a diamond helmet and a diamond chestplate, and tried to equip them to the armor stand. It didn't equip nor take it out of my inventory.

Problems:
Equipping armor stands in game does not work.
Equipping armor using `Entity.getEquipment().setHelmet()` and similar methods do not work
Equipping armor using `ArmorStand.setHelmet()` do not work.

Tested in build 478 (2017.5.0.0)

Postremus:
Add equipment sounds for armorstands. They are played to the player who equipped the armor stand. Currently no sound is played because of #519.

Implement EntityEquipment. This allows the armorstand to use the equipmentmonitor.

Save the off hand item of the livingentity. Use the tags HandItems, ArmorItems, HandDropChances, ArmorDropChances for this. Keep the old Equipment and DropChances tags for backwards compatibility.

Add off hand support for armor stands.

closes #471


src/main/java/net/glowstone/entity/objects/GlowArmorStand.java
```diff
@@ -1,13 +1,20 @@
 package net.glowstone.entity.objects;
 
+
```
aramperes: Whitespace! ;)

src/main/java/net/glowstone/io/entity/LivingEntityStore.java
```diff
@@ -148,6 +130,68 @@ public void load(T entity, CompoundTag compound) {
         }
     }
 
+    private void loadEquipment(T entity, EntityEquipment equip, CompoundTag compound) {
+        // Deprecated since 15w31a, left here for compatibilty for now
+        if (compound.isList("Equipment", TagType.COMPOUND)) {
+            List<CompoundTag> list = compound.getCompoundList("Equipment");
+            if (list.size() == 5) {
```
aramperes: Would it be better to allow for a variable size of this list? For example, having a list with only 1 element would only set the "item in hand" slot, and so forth.

aramperes (changes requested):

src/main/java/net/glowstone/io/entity/LivingEntityStore.java
Postremus: I can't see a Situation where:
- A world is saved with an incomplete list of "Equipment", "HandItems" or "ArmorItems"
- and the client still working

Just tried this situation out. Client is just black, and I can't move.

Since it is not safe to have incomplete lists in vanilla, I don't see how useful handling loading of them is going to be for us.

Postremus: Okay, more testing revealed, that paper has no problem with incomplete *Items lists.

aramperes: The only situation I could see this happening is when using data-tags in commands (/summon, for example): it would be possible for a command to not have all the items in the list, and the Entity Store would have to handle that.

aramperes:
@Postremus I don't have time to test this on my own system right now. Can you confirm that your changes work without major breakage? If not, I'll have a look myself tonight or during the weekend.

Postremus:
It seems to work without breaking anything.   

Executing:
`/summon armor_stand ~ ~ ~ {ShowArms:1,Invisible:1,HandItems:[{Count:1b,id:276s},{Count:1b,id:442s}], ArmorItems:[{Count:1b,id:313s}, {Count:1b,id:312s}, {Count:1b,id:311s}]}`  
Spawns the armorstand with 3 pieces of armor (foot, leg, chest), and 2 handitems, correctly.

No other entity uses an EntityEquipment right now in Glowstone.

aramperes:
@Postremus Alright, thank you for confirming this works :)

Perhaps adding EntityEquipment support for other entities can be done in a future PR. The scope of your PR is mostly about Armor Stands anyways.

aramperes (approved):

aramperes:
Thank you again for your contribution!

Issue #504, opened 2017-06-24
[#396] Entities: Implementation [$50 awarded]


gdude2002:
[![Bountysource bounty](https://api.bountysource.com/badge/issue?issue_id=46548049)](https://www.bountysource.com/issues/46548049-396-entities-implementation)

This issue is part of [this story](https://github.com/GlowstoneMC/Glowstone/issues/396), *Completing entities in Glowstone* (#396). To make the work easier and bounties fairer, we've split it into multiple issues.

Anyone can participate in the completion of this issue, which will be a major turning point in the development and the possible use-cases for Glowstone as a Minecraft server alternative.

---

**Before getting started!** If you are unfamiliar with Glowstone, please remember that Glowstone is a complete rewrite of the Minecraft vanilla server, aimed to be a better alternative to other software like Spigot. You are by no means allowed to push code to this repository based on Mojang's own platform (whether it be Forge, MCP, decompiled binaries, etc.) Doing so will cause in the rejection of your Pull Request. You are allowed to use other resources like the numerous wikis that exist on the Internet (Minecraft Wiki, wiki.vg, etc.) as well as trial-and-error.

---

Entities to be implemented
==================

Some entities are simply not implemented in Glowstone. Implementing an entity is a task split in multiple aspects, such as Metadata (networking), Storage (I/O), basic interactions, API implementation, etc.

These entities include:

* ~Armor Stand (need to be fixed, see #471)~ (done in #520)
* ~Boat (including steering and the networking behind it)~ (done in #529)
* ~Ender Crystal~ (done in #530)
* ~Experience Orb (and dropping them when killing entities, based on the loot tables that have been implemented in the past)~ (done in a7de3bda58b37dbee29e5e96d4e51e6da9f27017)
* ~Firework~ (done in #542)
* ~Leash (and leashing entities, preferably)~ (done in #538)
* ~Lightning~ (done in #546)
* ~Paintings~ (done in #535)
* Horses (mostly implemented, ~but the passenger system in completely broken and needs to be fixed~ (done in #524))

gdude2002:
As @Postremus has done almost all of this ticket himself and the $20 bounty on it was placed by him originally, I've decided to double it. 

Since it looks like you're gonna get the bounty when this ticket is closed, Postremus, grats and thanks for contributing!

aramperes:
Since all the tasks have been completed, the issue can be closed. The bounty will be awarded shortly to @Postremus.

[...]

PR #524, opened 2017-07-09
Support multiple passengers

### LINKED ISSUE #522: Multiple passengers need to be implemented
Currently, Glowstone only supports one passenger. Since 15w41a, multiple are possible through the `Passengers` tag in vanilla.  
This tag needs to be implemented correctly, and the api implementation has to be adjusted.

See:
http://minecraft.gamepedia.com/15w41a
http://wiki.vg/Protocol#Set_Passengers
https://beta.glowstone.net/jd/glowkit/org/bukkit/entity/Entity.html#getPassengers--
https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/entity/GlowEntity.java#L1113

Postremus:
This PR attempts to implement multiple passengers per vehicle.

Save and load Passengers. When an entity is a passenger, it will only be saved as child of its vehicle, and never alone.
Adjust api implementation to be (hopefully) in line with papers.
Also set the size for armorstands and horses.

closes #522
related to #504

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowEntity.java
@@ -899,9 +899,6 @@ public abstract class GlowEntity implements Entity {
             setRawLocation(location.clone().add(velocity));
         }
 
-        // apply location changes to all passengers
-
-
         // apply friction and gravity
         if (location.getBlock().getType() == Material.WATER) {
             velocity.multiply(liquidDrag);
@@ -1177,11 +1174,11 @@ public abstract class GlowEntity implements Entity {
 
     @Override
     public boolean setPassenger(Entity bPassenger) {
-        Preconditions.checkArgument(bPassenger != this, "Entity cannot ride itself.");
+        Preconditions.checkArgument(!this.equals(bPassenger), "Entity cannot ride itself.");
 
         boolean result = false;
         for (Entity passenger : Lists.newArrayList(passengers)) {
-            if (passenger != bPassenger) {
+            if (!Objects.equals(passenger, bPassenger)) {
                 result = !removePassenger(passenger);
             }
         }
--- a/src/main/java/net/glowstone/entity/GlowEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowEntity.java
@@ -999,6 +999,7 @@ public abstract class GlowEntity implements Entity {
         boundingBox = null;
         world.getEntityManager().unregister(this);
         server.getEntityIdManager().deallocate(this);
+        this.setPassenger(null);
     }
 
     ////////////////////////////////////////////////////////////////////////////
@@ -1122,7 +1123,7 @@ public abstract class GlowEntity implements Entity {
 
     @Override
     public boolean removePassenger(Entity passenger) {
-        Preconditions.checkArgument(passenger != this, "Entity cannot ride itself.");
+        Preconditions.checkArgument(!this.equals(passenger), "Entity cannot ride itself.");
 
         if (passenger == null || !passengers.contains(passenger)) return false; // nothing changed
 
@@ -1139,12 +1140,14 @@ public abstract class GlowEntity implements Entity {
         passengerChanged = true;
         glowPassenger.vehicle = null;
 
+        glowPassenger.teleport(getDismountLocation());
+
         return passengers.remove(passenger);
     }
 
     @Override
     public boolean addPassenger(Entity passenger) {
-        Preconditions.checkArgument(passenger != this, "Entity cannot ride itself.");
+        Preconditions.checkArgument(!this.equals(passenger), "Entity cannot ride itself.");
 
         if (passenger == null || passengers.contains(passenger)) return false; // nothing changed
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -1105,73 +1109,92 @@ public void setCustomNameVisible(boolean flag) {
...
     @Override
     public List<Entity> getPassengers() {
-        return Arrays.asList(new Entity[]{passenger}); // TODO: multiple passengers
+        return Collections.unmodifiableList(passengers);
     }
 
     @Override
     public boolean removePassenger(Entity passenger) {
-        return passenger.leaveVehicle();
+        Preconditions.checkArgument(passenger != this, "Entity cannot ride itself.");
```
aramperes: Replace `passenger != this` with `!Objects.equals(passenger, this)`

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -1105,73 +1109,92 @@ public void setCustomNameVisible(boolean flag) {
...
+        glowPassenger.vehicle = null;
+
+        return passengers.remove(passenger);
     }
 
     @Override
     public boolean addPassenger(Entity passenger) {
-        return setPassenger(passenger); // TODO: multiple passengers
-    }
+        Preconditions.checkArgument(passenger != this, "Entity cannot ride itself.");
```
aramperes: See previous review comment (don't use `!=`/`==` for equality checks)

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -1105,73 +1109,92 @@ public void setCustomNameVisible(boolean flag) {
...
-            if (!(bPassenger instanceof GlowEntity)) {
-                return false;
-            }
+        this.passengers.add(passenger);
+        glowPassenger.vehicle = this;
+        this.passengerChanged = true;
 
-            GlowEntity passenger = (GlowEntity) bPassenger;
+        Location onTopOfVehicle = this.location.clone().add(0, this.getHeight(), 0);
+        glowPassenger.teleport(onTopOfVehicle);
```
aramperes: I think you would also need to teleport the passenger after dismounting, no?

src/main/java/net/glowstone/io/entity/EntityStore.java
```diff
@@ -98,6 +102,37 @@ public void load(T entity, CompoundTag tag) {
             UUID uuid = UUID.fromString(tag.getString("UUID"));
             entity.setUniqueId(uuid);
         }
+
+        if (tag.isList("Passengers", TagType.COMPOUND)) {
+            for (CompoundTag entityTag : tag.getCompoundList("Passengers")) {
+                Entity passenger = loadPassenger(entity, entityTag);
+                if (passenger != null) {
+                    entity.addPassenger(passenger);
```
aramperes, line 110: Perhaps you should make another ("internal") method when setting the passengers from NBT -- you wouldn't want to fire these events when loading the world.

aramperes (changes requested):

src/main/java/net/glowstone/io/entity/EntityStore.java
Postremus: I have to agree with you, that you don't want to do this normally. Spigot however does throw them during world load..

aramperes: Ah okay, that's fine then.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -898,6 +899,9 @@ protected void pulsePhysics() {
             setRawLocation(location.clone().add(velocity));
         }
 
+        // apply location changes to all passengers
```
aramperes: Empty comment block? Or is there something missing here?

src/main/java/net/glowstone/io/entity/EntityStore.java
```diff
@@ -98,6 +102,37 @@ public void load(T entity, CompoundTag tag) {
...
+            // to the right coordinates.
+            NbtSerialization.locationToListTags(vehicle.getLocation(), compoundTag);
+        }
+
+        try {
+            // note that creating the entity is sufficient to add it to the world
+            return EntityStorage.loadEntity(vehicle.getWorld(), compoundTag);
+        } catch (Exception e) {
+            String id = compoundTag.isString("id") ? compoundTag.getString("id") : "<missing>";
+            GlowServer.logger.warning("Skipping Entity with id " + id);
```
aramperes: Log the exception here

aramperes (changes requested):

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -134,7 +135,7 @@
     /**
      * Passenger
      */
-    private GlowEntity passenger;
+    private List<Entity> passengers = new ArrayList<>();
```
aramperes: This field can be final (and perhaps `protected` too)

aramperes (changes requested):

aramperes (approved):

Issue #526, opened 2017-07-11
Result slot not being updated when crafting more than one of a specific item


puigru:
Hello,

While testing Glowstone on the test server, whenever I attempted to craft more than one of a specific item, the result slot on crafting inventories wouldn't update, leaving it empty.

![2017-07-12_00 11 47](https://user-images.githubusercontent.com/7153662/28093240-c06620f0-6696-11e7-97ed-af68d8049d6a.png)

Glowstone build: 983

Issue #525, opened 2017-07-11
Crafting inventories don't support shift-clicking


puigru:
Hello,

While testing Glowstone on the test server, I noticed that if you attempt to retrieve an item from the result slot in a crafting inventory by shift-clicking, the item will disappear server-side.

Glowstone build: 983

aramperes:
Are there any relevant error logs for this?

puigru:
@momothereal I just tested the latest artifact (985) locally, as my tests were done on mc.glowstone.net, and there were no errors shown in the console.

aramperes:
Fixed in 0ca671017009c8822a99bb88b6feb3fe53b85341

PR #527, opened 2017-07-13
#526 and #525 - Fix shift click when crafting and refresh the crafting result slot

FlorentClarret:
Hi,

This PR relates to the following issues :

- https://github.com/GlowstoneMC/Glowstone/issues/525
- https://github.com/GlowstoneMC/Glowstone/issues/526

The crafting result slot is now updated if a new item can be crafted. Also, I fixed the shift click on crafting as well. 

There is still one little thing I need to fix : when shift clicking, the amount of layers is first reduced by 1 and right after completely reset. The result is correct, all the items are crafted and the required items removed, but the layers are supposed to be removed all at once.

### DIFF
--- a/src/main/java/net/glowstone/inventory/GlowCraftingInventory.java
+++ b/src/main/java/net/glowstone/inventory/GlowCraftingInventory.java
@@ -67,16 +67,15 @@ public class GlowCraftingInventory extends GlowInventory implements CraftingInve
             // Place the items in the player's inventory (right to left)
             player.getInventory().tryToFillSlots(clickedItem, 8, -1, 35, 8);
 
-            // Craft the items, removing the ingredients from the crafting matrix
-            for (int i = 0; i < recipeAmount; i++) {
-                craft();
-            }
+            // Avoid calling craft because we already know the player can craft 'recipeAmount' of this item
+            CraftingManager cm = ((GlowServer) Bukkit.getServer()).getCraftingManager();
+            // Removing all the items at once will avoid multiple useless calls to craft (and all of its sub methods like getRecipe)
+            cm.removeItems(getMatrix(), this, recipeAmount);
         } else {
             // Clicked in the crafting grid, no special handling required (just place them left to right)
             clickedItem = player.getInventory().tryToFillSlots(clickedItem, 9, 36, 0, 9);
             view.setItem(clickedSlot, clickedItem);
         }
-
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


FlorentClarret:
For some reason, when shift clicking, a single click seems to be performed on the client side before the server get the shift click. This is where the - 1 on the layers comes from. On debug mode, i didn't manage to catch this "single click" before the shift click, and the shift click is working properly.

Do you have any idea where this comes from ?

I will try to fix this a bit later this week.

FlorentClarret:
I didn't manage to find out why we have this first layer removed on the client side before the server remove all the layers. However, this PR closes the two issues so if it's ok for you we could merge it.

If you have any idea why we have this, feel free to tell me :)

src/main/java/net/glowstone/inventory/GlowCraftingInventory.java
```diff
@@ -60,23 +55,24 @@ public void handleShiftClick(GlowPlayer player, InventoryView view, int clickedS
...
 
             // Place the items in the player's inventory (right to left)
             player.getInventory().tryToFillSlots(clickedItem, 8, -1, 35, 8);
 
-            // Craft the items, removing the ingredients from the crafting matrix
-            for (int i = 0; i < recipeAmount; i++) {
-                craft();
-            }
+            // Avoid calling craft because we already know the player can craft 'recipeAmount' of this item
+            CraftingManager cm = ((GlowServer) Bukkit.getServer()).getCraftingManager();
```
aramperes: You can use `player.getServer()` which returns `GlowServer`

src/main/java/net/glowstone/inventory/crafting/CraftingManager.java
```diff
@@ -115,7 +129,7 @@ public void removeItems(ItemStack[] items, GlowCraftingInventory inv) {
     public static int getLayers(ItemStack... items) {
         int layers = 0;
         for (ItemStack item : items) {
-            if (item != null && (item.getAmount() < layers || layers == 0)) {
+            if (item != null && item.getAmount() != 0 && (item.getAmount() < layers || layers == 0)) {
```
aramperes: Use `!InventoryUtil.isEmpty(item)` instead of amount/null check

aramperes (changes requested):

aramperes (approved):

aramperes:
Thank you for your contributions :)

The layer-removing glitch is very minor, but you can feel free to open an issue about it.

PR #528, opened 2017-07-15
GPGPU Surface Noise

mastercoms:
**Problem:** Terrain generation is slow and uses a lot of our CPU time, which could be used for other game tasks, like AI.

**PR Goal:** Offload overworld surface noise operations onto the GPU.

**PR Changes:** This PR adds a basic noise generation system using the Glowstone OpenCL API. For the purpose of these changes, this PR also adds support for builtin OpenCL programs and also improves logging for OpenCL program loading, due to variants in cases.

**Notes:** The PR should be tested with the `advanced.gpgpu` config option set to `true` in `glowstone.yml`. It is recommended to run a client on a GPU different than the one Glowstone uses for GPGPU. Finally, the noise algorithm is not well developed, but it should work fine for a first revision of this non-default terrain generation system.


mastercoms:
Well, that's all the improvements I can think of. Should be good enough to review now.

aramperes (approved):
Looks good to me!

PR #529, opened 2017-07-15
Implement boats

Postremus:
This pr attempts to implement boats, excluding:
- boat specific physics
 - Boats seem to be affected by gravity, and therefore sink when placed
 - Most of the deprecated API methods are not implemented

Some basic interactions, like damage calculation and sounds are also missing. They will be done in a future pr, since this one is already to big for my taste.

related to #504

### DIFF
--- a/src/main/java/net/glowstone/net/handler/play/inv/VehicleMoveHandler.java
+++ b/src/main/java/net/glowstone/net/handler/play/inv/VehicleMoveHandler.java
@@ -1,7 +1,6 @@
 package net.glowstone.net.handler.play.inv;
 
 import com.flowpowered.network.MessageHandler;
-import java.util.Objects;
 import net.glowstone.EventFactory;
 import net.glowstone.GlowServer;
 import net.glowstone.entity.GlowEntity;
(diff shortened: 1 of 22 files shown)


src/main/java/net/glowstone/net/handler/play/inv/VehicleMoveHandler.java
```diff
@@ -1,12 +1,119 @@
...
+            vehicle.teleport(oldLocation);
+            return;
+        }
+
+        // call move event if movement actually occurred and there are handlers registered
+        if (!oldLocation.equals(newLocation) && VehicleMoveEvent.getHandlerList().getRegisteredListeners().length > 0) {
+            VehicleMoveEvent event = EventFactory.callEvent(new VehicleMoveEvent((Vehicle) vehicle, oldLocation, newLocation));
+
+            if (!event.getTo().equals(newLocation)) {
+                vehicle.teleport(event.getTo(), TeleportCause.PLUGIN);
```
aramperes: Is there a reason to set this to `TeleportCause.PLUGIN`?

Postremus: Changed the implementation to be more in line with papers. Changing the to or from location of this event does not seem to do anything there.

aramperes (approved):

aramperes:
Thank you for your contribution :)

PR #530, opened 2017-07-16
Implement ender crystals

Postremus:
This pr attempts to implement ender crystals, excluding
- Spawning them in the end
- enderdragons charging from them
- destruction by pushing them with pistons

related to #504

### DIFF
--- a/src/main/java/net/glowstone/entity/meta/MetadataMap.java
+++ b/src/main/java/net/glowstone/entity/meta/MetadataMap.java
@@ -112,6 +112,11 @@ public class MetadataMap {
         return get(index, MetadataType.ITEM, null);
     }
 
+    /**
+     * Get an optional position
+     * @param index of the optposition
+     * @return either an instanc of a BlockVector, or null if not present
+     */
     public BlockVector getOptPosition(MetadataIndex index) {
         return get(index, MetadataType.OPTPOSITION, null);
     }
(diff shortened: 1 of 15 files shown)


src/main/java/net/glowstone/entity/meta/MetadataMap.java
```diff
@@ -111,6 +112,15 @@ public ItemStack getItem(MetadataIndex index) {
         return get(index, MetadataType.ITEM, null);
     }
 
+    /**
+     * Get an optional position
```
aramperes:
```
* Gets the optional position value for the given MetadataIndex
* 
* @param the MetadataIndex of the optional position
* @return the position value as a BlockVector, null if the value is not present
```

src/test/java/net/glowstone/util/PositionTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Test
+    public void testGetPositionAsBlockVector() throws Exception {
+        BlockVector blockVector = new BlockVector(-5, 15, -20);
+        long position = -1373315792916L;
+
+        assertThat(Position.getPosition(position), is(blockVector));
+    }
+
+}
```
aramperes: Add a new line at the end (LF)

start.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-java -Xms1G -Xmx1G -XX:+UseG1GC -jar glowstone.jar
+java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 -Xms1G -Xmx1G -XX:+UseG1GC -jar glowstone.jar
```
aramperes: Is this intentional?

aramperes (changes requested):

aramperes (approved):

PR #531, opened 2017-07-17
#499 - Implement playsound command

FlorentClarret:
Hi,

Another PR on the https://github.com/GlowstoneMC/Glowstone/issues/499 issue. This time, I added playsound.

### DIFF
--- a/src/main/java/net/glowstone/command/minecraft/PlaySoundCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/PlaySoundCommand.java
@@ -1,0 +1,174 @@
+package net.glowstone.command.minecraft;
+
+import net.glowstone.command.CommandTarget;
+import net.glowstone.command.CommandUtils;
+import net.glowstone.constants.GlowSound;
+import net.glowstone.entity.GlowPlayer;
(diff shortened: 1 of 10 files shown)


src/main/java/net/glowstone/command/minecraft/PlaySoundCommand.java
```diff
@@ -0,0 +1,174 @@
...
+
+        return true;
+    }
+
+    @Override
+    public List<String> tabComplete(CommandSender sender, String alias, String[] args) throws IllegalArgumentException {
+        if (args == null) {
+            return Collections.emptyList();
+        } else if (args.length == 1) {
+            return StringUtil.copyPartialMatches(args[0], SOUNDS, new ArrayList(SOUNDS.size()));
```
aramperes: You can do something similar to what I did for the [Give command](https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/command/minecraft/GiveCommand.java#L95-L100) (in order to keep the same behavior as vanilla)

aramperes (approved):

PR #532, opened 2017-07-22
Cleanup itemframes

Postremus:
Moved facing attribute into new GlowHangingEntity super class.
Remove all methods used to set or get by a "facingNumber", and use the new enum HangingFace instead.
Pull save and load of "Facing" attribute up into new super class "HangingStore".
call some more of the hanging* events, but there are for sure some instances missing where more events should be called.

related to #504, since this is the groundwork for paintings

### DIFF
--- a/src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
+++ b/src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
@@ -80,6 +80,7 @@ public class GlowItemFrame extends GlowHangingEntity implements ItemFrame {
                 }
                 remove();
             } else {
+                world.dropItemNaturally(location, getItem().clone());
                 setItem(new ItemStack(Material.AIR));
                 setRotation(Rotation.NONE);
             }
@@ -94,6 +95,9 @@ public class GlowItemFrame extends GlowHangingEntity implements ItemFrame {
         if (ticksLived % 11 == 0) {
 
             if (location.getBlock().getRelative(getAttachedFace()).getType() == Material.AIR) {
+                if (EventFactory.callEvent(new HangingBreakEvent(this, RemoveCause.PHYSICS)).isCancelled()) {
+                    return;
+                }
                 world.dropItemNaturally(location, new ItemStack(Material.ITEM_FRAME));
                 if (!isEmpty()) {
                     world.dropItemNaturally(location, getItem().clone());
@@ -196,10 +200,10 @@ public class GlowItemFrame extends GlowHangingEntity implements ItemFrame {
 
     @Override
     public void setItem(ItemStack is) {
-        if (is == null) {
-            is = new ItemStack(Material.AIR, 1);
-        }
+        is = InventoryUtil.itemOrEmpty(is);
+        is = is.clone();
         is.setAmount(1);
+
         metadata.set(MetadataIndex.ITEM_FRAME_ITEM, is);
         setRotation(Rotation.NONE);
     }
--- a/src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
+++ b/src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
@@ -1,8 +1,9 @@
 package net.glowstone.entity.objects;
 
 import com.flowpowered.network.Message;
+import net.glowstone.EventFactory;
 import net.glowstone.chunk.GlowChunk.Key;
-import net.glowstone.entity.GlowEntity;
+import net.glowstone.entity.GlowHangingEntity;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.entity.meta.MetadataIndex;
 import net.glowstone.net.message.play.entity.EntityMetadataMessage;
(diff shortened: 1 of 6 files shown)


src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
```diff
@@ -249,60 +177,46 @@ public Location getOrigin() {
...
+    public void setFacingDirection(BlockFace blockface) {
+        facing = HangingFace.getByBlockFace(blockface);
+        createTeleportMessage(blockface);
     }
 
     @Override
     public void setItem(ItemStack is) {
-        setItemInFrame(is);
+        if (is == null) {
+            is = new ItemStack(Material.AIR, 1);
```
aramperes: You can use `is = InventoryUtil.itemOrEmpty(is)`

aramperes (changes requested):

src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
```diff
@@ -196,10 +200,10 @@ public void setFacingDirection(BlockFace blockface) {
 
     @Override
     public void setItem(ItemStack is) {
-        if (is == null) {
-            is = new ItemStack(Material.AIR, 1);
-        }
+        is = InventoryUtil.itemOrEmpty(is);
+        is = is.clone();
         is.setAmount(1);
```
aramperes: I'm not sure what you're doing here. If a plugin uses this method with say, `ItemStack(Material.STONE, 0)` it should be treated as empty instead of updating its amount to 1 (unless vanilla does it differently?). Cloning the ItemStack should not be necessary either.

Postremus: Cloning is required, imo.
During testing, I had the following scenario:
- I have 4 Rails in my Inventory
- I put one into the itemframe. Now I have 3 Rails in my Inventory.
- I rightclick onto the itemframe, and the items drop.
- I pick them up, and now I have 6 Rails in my Inventory.
Cloning solved this scenario.

Also, at least on paper, the amount of the item seems to be always set to 1.

aramperes: Alright. You can still simplify this with `is = InventoryUtil.itemOrEmpty(is).clone()`.

src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
```diff
@@ -44,81 +43,21 @@ public GlowItemFrame(GlowPlayer player, Location location, BlockFace clickedface
...
-                return BlockFace.WEST;
-            case 2:
-                return BlockFace.NORTH;
-            case 3:
-                return BlockFace.EAST;
-            default:
-                return BlockFace.SOUTH;
-        }
+        setRotation(Rotation.NONE);
+        setItem(new ItemStack(Material.AIR));
```
aramperes: `InventoryUtil.createEmptyStack()`

src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
```diff
@@ -44,81 +43,21 @@ public GlowItemFrame(GlowPlayer player, Location location, BlockFace clickedface
...
-    public void setItemFrameRotation(int rotation) {
-        metadata.set(MetadataIndex.ITEM_FRAME_ROTATION, rotation);
-    }
-
     @Override
     public boolean entityInteract(GlowPlayer player, InteractEntityMessage message) {
         if (message.getAction() == Action.INTERACT.ordinal()) {
-            if (itemInFrame == Material.AIR) {
+            if (InventoryUtil.isEmpty(getItem())) {
                 ItemStack isInHand = player.getItemInHand();
```
aramperes: Please take the hand into consideration -- see https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/net/message/play/player/InteractEntityMessage.java#L25 and https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/net/handler/play/player/InteractEntityHandler.java#L52

src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
```diff
@@ -129,19 +68,21 @@ public boolean entityInteract(GlowPlayer player, InteractEntityMessage message)
...
+                // TODO: use correct RemoveCause
+                if (EventFactory.callEvent(new HangingBreakEvent(this, RemoveCause.DEFAULT)).isCancelled()) {
+                    return false;
+                }
                 remove();
             } else {
-                setItemInFrame(new ItemStack(Material.AIR));
-                rot = 0;
+                world.dropItemNaturally(location, getItem().clone());
+                setItem(new ItemStack(Material.AIR));
```
aramperes: `InventoryUtil.createEmptyStack()`

aramperes (changes requested):

src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
aramperes: I'll correct myself: it doesn't seem like you can place items from the off-hand for Item Frames and Armor Stands (see: https://bugs.mojang.com/browse/MC-84771). Please add a comment or `todo` around this check in case this vanilla bug is fixed in the future (although it has been opened for nearly 2 years). The check remains OK as long as they don't fix this.

aramperes (approved):

PR #533, opened 2017-07-26
Clean-up unnecessary comments and sout

louxiu:
1.Interface has comments already, so the comments in implements is
unnecessary
2. Replace sout with log

### DIFF
--- a/src/main/java/net/glowstone/net/handler/play/player/AdvancementTabHandler.java
+++ b/src/main/java/net/glowstone/net/handler/play/player/AdvancementTabHandler.java
@@ -1,16 +1,19 @@
 package net.glowstone.net.handler.play.player;
 
 import com.flowpowered.network.MessageHandler;
+import net.glowstone.GlowServer;
 import net.glowstone.net.GlowSession;
 import net.glowstone.net.message.play.player.AdvancementTabMessage;
+
+import java.util.logging.Level;
 
 public class AdvancementTabHandler implements MessageHandler<GlowSession, AdvancementTabMessage> {
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/net/handler/play/player/AdvancementTabHandler.java
```diff
@@ -1,16 +1,19 @@
...
 import net.glowstone.net.message.play.player.AdvancementTabMessage;
 
+import java.util.logging.Level;
+
 public class AdvancementTabHandler implements MessageHandler<GlowSession, AdvancementTabMessage> {
     @Override
     public void handle(GlowSession session, AdvancementTabMessage message) {
         if (message.getAction() == AdvancementTabMessage.ACTION_CLOSE) {
-            System.out.println("Received AdvancementTabHandler (close)");
+            GlowServer.logger.log(Level.SEVERE, "Received AdvancementTabHandler (close)");
```
aramperes: `Level.DEBUG`/`Level.INFO` perhaps?

aramperes (changes requested):

louxiu:
@momothereal Oh, I didn't comprehend the `Level` correctly. I think `DEBUG`(Level.FINE?) is suitable.

src/main/java/net/glowstone/net/handler/play/player/AdvancementTabHandler.java
louxiu: done

aramperes (approved):

PR #535, opened 2017-07-29
Implement paintings

Postremus:
This PR attempts to implement paintings.

GlowHangingEntity and HangingStore are reused for this, since the facing value is the same as for itemframes.

Stores the "Motive" of a painting in CamelCase, like vanilla does it.

When "Painting.setFacing" or "Painting.setArt" is called on paper, a DestroyEntities and SpawnPainting Message is sent. A new id is used for spawning, which probably means a new entity is created.
Changing of the entity id can not be easily done here, and does not make much sense, therefore not implemented.

getHeight() and getWidth() return regardless of the art size 0.5 each, which is the same as paper.

The BoundingBox x and z of a painting is defined as "type width or 0.0625 (depth) ". I had to adjust the EntityBoundingBox to make it possible to set width and depth separately.

A painting has multiple important locations, each relative to the topleftcorner of the painting:
- center of art: Defined as (max(0, artWidth / 2 - 1), artHeight / 2)). This one is needed for spawning this entity. It is also used as center of rotation.
- center of painting entity: Defined as (artWidth / 2, artHeight / 2). This one is the center, and location, of the entity.

related to #504


src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+    public void setFacingDirection(BlockFace blockFace) {
+        setFacingDirection(blockFace, false);
+    }
+
+    @Override
+    public void pulse() {
+        super.pulse();
+
+        if (ticksLived % 11 == 0) {
+
```
aramperes: Whitespace

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+        return HangingFace.values()[(facing.ordinal() + 1) % 4].getBlockFace();
+    }
+
+    private BlockFace getRightFace() {
+        return getLeftFace().getOppositeFace();
+    }
+
+    @Override
+    public double getWidth() {
+        return 0.5;
```
aramperes: Add a note about this value being static (like you mentioned in the PR description)

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+    @Override
+    public void setFacingDirection(BlockFace blockFace) {
+        setFacingDirection(blockFace, false);
+    }
+
+    @Override
+    public void pulse() {
+        super.pulse();
+
+        if (ticksLived % 11 == 0) {
```
aramperes: Where does the `11` value come from? Do paintings only update every 0.55 seconds?

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+            updateBoundingBox();
+            return false;
+        }
+
+        respawn();
+
+        return true;
+    }
+
+    private void respawn() {
```
aramperes: Add documentation about this method (sends destroy and then spawn messages).
I would rename this method `refresh` and make it available publicly for plugins wanting to use it.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+        Message[] messages = new Message[] {destroyMessage, spawnMessages.get(0)};
+
+        getWorld()
+            .getRawPlayers()
+            .stream()
+            .filter(p -> p.canSeeEntity(this))
+            .forEach(p -> p.getSession().sendAll(messages));
+    }
+
+    public void setArtInternal(Art art) {
```
aramperes: Add documentation about this method because it is public.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }
+    }
+
+    @Override
+    protected void pulsePhysics() {
+        // not affected by physics
+    }
+
+    public boolean isObstructed() {
```
aramperes: Add documentation about this method because it is public.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+                if (!canHoldPainting(current)) {
+                    return true;
+                }
+                current = current.getBlock().getRelative(right).getLocation();
+            }
+
+            current = current.getBlock().getRelative(BlockFace.DOWN).getLocation();
+
+            // reset x and z
+            current.subtract(right.getModX() * art.getBlockWidth(), 0, right.getModZ() * art.getBlockWidth());
```
aramperes, line 275: Since you are using `subtract`, make sure the locations are cloned before updating them, IIRC it can update the block's location too (unless that only happens on Spigot?)

src/main/java/net/glowstone/net/message/play/entity/SpawnPaintingMessage.java
```diff
@@ -1,12 +1,14 @@
...
 
 import com.flowpowered.network.Message;
+import java.util.UUID;
 import lombok.Data;
 
 @Data
 public final class SpawnPaintingMessage implements Message {
 
     private final int id;
+    private final UUID uuid;
```
aramperes: Rename this field to `uniqueId`.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,334 @@
...
+        );
+    }
+
+    @Override
+    public Art getArt() {
+        return art;
+    }
+
+    @Override
+    public boolean setArt(Art art) {
```
aramperes, line 127: Check if `art` is not null, default to Kebab if it is.

aramperes (changes requested):

aramperes:
This will likely be our last feature for the `2017.7` release which is set to Monday/Tuesday. I'd like to thank you for all of your contributions you've done recently!

src/main/java/net/glowstone/entity/objects/GlowPainting.java
Postremus: I took the value from the itemframes, since they are hanging entities as well. I will recheck this.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
Postremus: Paper just throws a NPE in this case: https://gist.github.com/Postremus/a2efa23ead7915e95183f6d93610454c  
I changed setArt and setFacing to do the same.

aramperes: I wouldn't throw the exception. That's a bug with their implementation and is not an API design choice - use notnull validation here.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
Postremus: `getTopLeftCorner()`, which is used to obtain the starting location, returns a cloned location.  
`GlowBlock.getLocation()` always returns a new Location, therefore the location of the block is not updated.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
Postremus: Did some testing on vanilla using a clock.
The hanging entity is only destroyed once `ticksLived % 5 seconds == 0`.
This is true for paintings as well as item frames.  

On paper, this holds true as well.
https://gist.github.com/Postremus/5358c837099817bbfd98bcdbad01ee1e

Postremus: I adjusted these checks to be more in line with vanilla.

src/main/java/net/glowstone/entity/objects/GlowPainting.java
```diff
@@ -0,0 +1,365 @@
...
+                current = current.getBlock().getRelative(right).getLocation();
+            }
+
+            current = current.getBlock().getRelative(BlockFace.DOWN).getLocation();
+
+            // reset x and z
+            current.subtract(right.getModX() * art.getBlockWidth(), 0, right.getModZ() * art.getBlockWidth());
+        }
+
+        return !this.world.getEntityManager().getEntitiesInside(this.boundingBox, this).isEmpty();
```
aramperes: What happens if a player 'collides' with the painting? Wouldn't it be destroyed here?

Postremus: Changed this so only hanging entities are checked for.

aramperes (approved):

PR #536, opened 2017-07-31
Properly handle generators with null spawns

mastercoms:
This fixes a null pointer exception when a generator returns a null spawn. This is valid according to the Bukkit spec and means that the server should find a random spawn point.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -766,7 +766,20 @@ public final class GlowWorld implements World {
             // find a spawn if needed
             Location spawn = generator.getFixedSpawnLocation(this, random);
             // we're already going to anchor if told to, so don't request another anchor
-            setSpawnLocation(spawn.getBlockX(), spawn.getBlockY(), spawn.getBlockZ(), false);
+            if (spawn == null) {
+                // determine a location randomly
+                int spawnX = random.nextInt(256) - 128, spawnZ = random.nextInt(256) - 128;
+                getChunkAt(spawnX >> 4, spawnZ >> 4).load(true);  // I'm not sure there's a sane way around this
+
+                for (int tries = 0; tries < 1000 && !generator.canSpawn(this, spawnX, spawnZ); ++tries) {
+                    spawnX += random.nextInt(256) - 128;
+                    spawnZ += random.nextInt(256) - 128;
+                }
+                setSpawnLocation(spawnX, getHighestBlockYAt(spawnX, spawnZ), spawnZ);
+                needSpawn = false;
+            } else {
+                setSpawnLocation(spawn.getBlockX(), spawn.getBlockY(), spawn.getBlockZ(), false);
+            }
         }
 
         if (spawnChunkLock == null) {


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -766,7 +766,20 @@ public void setKeepSpawnInMemory(boolean keepLoaded) {
...
+            if (spawn == null) {
+                // determine a location randomly
+                int spawnX = random.nextInt(256) - 128, spawnZ = random.nextInt(256) - 128;
+                getChunkAt(spawnX >> 4, spawnZ >> 4).load(true);  // I'm not sure there's a sane way around this
+
+                for (int tries = 0; tries < 1000 && !generator.canSpawn(this, spawnX, spawnZ); ++tries) {
+                    spawnX += random.nextInt(256) - 128;
+                    spawnZ += random.nextInt(256) - 128;
+                }
+                setSpawnLocation(spawnX, getHighestBlockYAt(spawnX, spawnZ), spawnZ);
```
aramperes, line 778: Wouldn't it be `getHighestBlockYAt(spawnX, spawnZ) + 1`? Or is the 1 offset handled when actually spawning the player?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -766,7 +766,20 @@ public void setKeepSpawnInMemory(boolean keepLoaded) {
             // find a spawn if needed
             Location spawn = generator.getFixedSpawnLocation(this, random);
             // we're already going to anchor if told to, so don't request another anchor
-            setSpawnLocation(spawn.getBlockX(), spawn.getBlockY(), spawn.getBlockZ(), false);
+            if (spawn == null) {
+                // determine a location randomly
+                int spawnX = random.nextInt(256) - 128, spawnZ = random.nextInt(256) - 128;
+                getChunkAt(spawnX >> 4, spawnZ >> 4).load(true);  // I'm not sure there's a sane way around this
+
+                for (int tries = 0; tries < 1000 && !generator.canSpawn(this, spawnX, spawnZ); ++tries) {
```
aramperes, line 774: 1000 seems rather high, especially if all attempts going to fail because the spawn chunk is in the ocean or something.

mastercoms: https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/generator/GlowChunkGenerator.java#L66

src/main/java/net/glowstone/GlowWorld.java
mastercoms: Spawn point is a solid block

src/main/java/net/glowstone/GlowWorld.java
aramperes: I'm sure I'm missing something, but why not just use that method instead of duping the code? Also, that doesn't change that 1000 is a lot :P

mastercoms: Because that method isn't static.

I didn't change the value because that's outside of the scope of the PR. I could expand the scope to change the loop to a more appropriate value, and add a static helper method for default random spawn finding code.

aramperes: `generator.getFixedSpawnLocation(this, random)`? It wouldn't need to be static

src/main/java/net/glowstone/GlowWorld.java
aramperes: Are you sure? If you use `/setworldspawn`, it would set the spawn at the location of the executor which is 1 block above the surface.

src/main/java/net/glowstone/GlowWorld.java
mastercoms: https://github.com/GlowstoneMC/Glowstone/pull/536/files#diff-cea1d25e5518f2495b11e2d20021b0a0R767

The whole point of this PR is to add a case for when `generator.getFixedSpawnLocation(this, random)` returns null (as mentioned in the PR post "This fixes a null pointer exception when a generator returns a null spawn").

aramperes: Oh, I see. I should have read the PR description better then, my bad :)

src/main/java/net/glowstone/GlowWorld.java
mastercoms: My mistake, but I know that `getHighestBlockYAt` produces the correct Y value for the spawn point, whatever that value is supposed to be.

aramperes: Alright, I guess the Y position doesn't really matter anyways because it is adjusted to the correct height when a player spawns (to avoid suffocation).

aramperes (approved):

PR #537, opened 2017-08-06
Use map instead of list for CommandFunction

louxiu:


src/main/java/net/glowstone/command/minecraft/FunctionCommand.java
```diff
@@ -37,13 +37,7 @@ public boolean execute(CommandSender sender, String label, String[] args) {
...
         String functionName = args[0];
-        CommandFunction function = null;
-        for (CommandFunction f : world.getFunctions()) {
-            if (f.getFullName().equals(functionName)) {
-                function = f;
-                break;
-            }
-        }
+        CommandFunction function = world.getFunctions().get(functionName);
         if (function == null) {
```
aramperes: Use `getFunctions().containsKey()` check here, and `get` the function after that check.

aramperes (changes requested):

louxiu:
@momothereal done

aramperes (approved):
Looks good! Please avoid re-writing commits/history though, splitting in multiple commits is easier for reviewing changes - we squash everything when merging anyways :)

louxiu:
@momothereal ok

PR #538, opened 2017-08-06
Implement Leashes

Postremus:
This PR implements Leashes, and leashing of entities.

GlowHangingEntity is reused for this, since LeashHitches are HangingEntities.

The leash holder is stored as part of a Living Entity ("Leash" Compound).  
Either as X/Y/Z if the holder is a leash hitch, or as a uuid, if the holder is another living entity.  
Other non living leash holders are not persisted, though the leash still has to be dropped on the next world load since it basically broke.

The facing value of leash hitches most likely does nothing.

Leashed entities try to follow their holder. The physics behind this is not even close to vanilla-like though.

related to #504

### DIFF
--- a/src/main/java/net/glowstone/GlowLeashHitch.java
+++ b/src/main/java/net/glowstone/GlowLeashHitch.java
@@ -130,7 +130,7 @@ public class GlowLeashHitch extends GlowHangingEntity implements LeashHitch {
         );
 
         Optional<LeashHitch> first = sorted.findFirst();
-        return first.orElse(block.getWorld().spawn(block.getLocation(), LeashHitch.class));
+        return first.orElseGet(() -> first.orElse(block.getWorld().spawn(block.getLocation(), LeashHitch.class)));
     }
 
     /**
--- a/src/main/java/net/glowstone/GlowLeashHitch.java
+++ b/src/main/java/net/glowstone/GlowLeashHitch.java
@@ -46,7 +46,6 @@ public class GlowLeashHitch extends GlowHangingEntity implements LeashHitch {
 
     @Override
     public List<Message> createSpawnMessage() {
-
         int x = location.getBlockX();
         int y = location.getBlockY();
         int z = location.getBlockZ();
@@ -79,8 +78,8 @@ public class GlowLeashHitch extends GlowHangingEntity implements LeashHitch {
 
     @Override
     public boolean shouldSave() {
-        // The Glowhitch on its own should never be saved
-        // It is saved in the leashed entity
+        // The GlowLeashHitch on its own should never be saved
+        // It is saved as part of the leashed living entity
         return false;
     }
 
--- a/src/main/java/net/glowstone/GlowLeashHitch.java
+++ b/src/main/java/net/glowstone/GlowLeashHitch.java
@@ -133,13 +133,23 @@ public class GlowLeashHitch extends GlowHangingEntity implements LeashHitch {
         return first.orElse(block.getWorld().spawn(block.getLocation(), LeashHitch.class));
     }
 
+    /**
+     * Checks if an Entity of the specified type is allowed to be a leash holder
+     *
+     * @param type type of the entity which wishes to become a leash holder
+     * @return if the type is allowed as a leash holder true, otherwise false
+     */
+    public static boolean isAllowedLeashHolder(EntityType type) {
+        return !(EntityType.ENDER_DRAGON.equals(type) || EntityType.WITHER.equals(type) || EntityType.PLAYER.equals(type) || EntityType.BAT.equals(type));
+    }
+
     @Override
     public boolean entityInteract(GlowPlayer player, InteractEntityMessage message) {
         if ((message.getAction() == Action.ATTACK.ordinal()) && message.getHandSlot() == EquipmentSlot.HAND) {
             remove();
         }
 
-        if ((message.getAction() == Action.INTERACT_AT.ordinal()) && message.getHandSlot() == EquipmentSlot.HAND) {
+        if ((message.getAction() == Action.INTERACT.ordinal()) && message.getHandSlot() == EquipmentSlot.HAND) {
             if (player.getLeashedEntities().isEmpty()) {
                 List<GlowEntity> entities = ImmutableList.copyOf(getLeashedEntities());
                 for (GlowEntity leashedEntity : entities) {
--- a/src/main/java/net/glowstone/GlowLeashHitch.java
+++ b/src/main/java/net/glowstone/GlowLeashHitch.java
@@ -64,6 +64,9 @@ public class GlowLeashHitch extends GlowHangingEntity implements LeashHitch {
 
     @Override
     public boolean setFacingDirection(BlockFace blockFace, boolean force) {
+        if (blockFace == null) {
+            return false;
+        }
         // facing most likely just does nothing for the leash hitch
         this.facing = HangingFace.getByBlockFace(blockFace);
         return true;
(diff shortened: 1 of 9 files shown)


src/main/java/net/glowstone/GlowLeashHitch.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public GlowLeashHitch(Location location, BlockFace clickedface) {
+        super(location, clickedface);
+        setSize(0.375f, 0.5f);
+        setGravity(false);
+    }
+
+    @Override
+    public List<Message> createSpawnMessage() {
+
```
aramperes: Whitespace

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -534,6 +611,25 @@ public void setRawLocation(Location location) {
     public abstract List<Message> createSpawnMessage();
 
     /**
+     * Creates a {@link Message} which can be sent to a client directly after the entity is spawned.
```
aramperes: In this case it would represent a *List* of messages to be sent, not a single message.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -558,6 +558,7 @@ public void remove() {
             scoreboard.unsubscribe(this);
             scoreboard = null;
         }
+
```
aramperes: Whitespace

src/main/java/net/glowstone/io/entity/LivingEntityStore.java
```diff
@@ -130,6 +133,26 @@ public void load(T entity, CompoundTag compound) {
...
+            // We know that there was something leashed, but not what entity it was
+            // This can happen, when for example Minecart got leashed
+            // We still have to make sure that we drop a Leash Item
+            entity.setLeashHolderUniqueID(UUID.randomUUID());
+        } else if (compound.isCompound("Leash")) {
+            CompoundTag leash = compound.getCompound("Leash");
+            if (leash.isLong("UUIDMost") && leash.isLong("UUIDLeast")) {
+                UUID uuid = new UUID(leash.getLong("UUIDMost"), leash.getLong("UUIDLeast"));
+                entity.setLeashHolderUniqueID(uuid);
+            } else if (leash.isInt("X") || leash.isInt("Y") || leash.isInt("Z")) {
```
aramperes: I think this should check if all coordinates exist (use `&&` instead of `||`)

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -1003,5 +991,36 @@ public AttributeInstance getAttribute(Attribute attribute) {
     public TaskManager getTaskManager() {
         return taskManager;
     }
+
+    @Override
+    public boolean entityInteract(GlowPlayer player, InteractEntityMessage message) {
+        super.entityInteract(player, message);
+        if (message.getHandSlot() == EquipmentSlot.HAND) {
+            if (!InventoryUtil.isEmpty(player.getItemInHand()) && player.getItemInHand().getType() == Material.LEASH) {
```
aramperes: Please reduce the use of empty lines/whitespace in this method

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -1435,4 +1553,55 @@ protected String disambiguate(Entity subject, String metadataKey) {
...
+        if (!isLeashed()) {
+            throw new IllegalStateException("Entity not leashed");
+        }
+
+        return leashHolder;
+    }
+
+    public boolean setLeashHolder(Entity holder) {
+        // "This method has no effect on EnderDragons, Withers, Players, or Bats"
+        if (this instanceof EnderDragon || this instanceof Wither || this instanceof Player || this instanceof Bat) {
```
aramperes: Maybe use `getType()` here instead?

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -599,6 +695,16 @@ public void setRawLocation(Location location) {
             passengerChanged = false;
         }
 
+        if (leashHolderChanged) {
+            int attached = isLeashed() && session.getPlayer().getEntityId() == leashHolder.getEntityId() ? 0 : this.getEntityId();
+            int holder = !isLeashed() ? -1 : leashHolder.getEntityId();
+
+            // When the leashHolder is not see able, the AttachEntityMessage will be created in createAfterSpawnMessage()
```
aramperes: "see able" -> "visible"

aramperes (changes requested):

src/main/java/net/glowstone/entity/GlowEntity.java
Postremus: yep, makes sense.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -997,13 +997,23 @@ public boolean entityInteract(GlowPlayer player, InteractEntityMessage message)
         super.entityInteract(player, message);
 
         if (message.getHandSlot() == EquipmentSlot.HAND) {
-            if (!InventoryUtil.isEmpty(player.getItemInHand()) && player.getItemInHand().getType() == Material.LEASH) {
+            ItemStack handUten = player.getItemInHand();
```
aramperes: It appears you can use the off-hand with leads, and `getItemInHand()` actually gets the main hand.

Also, you are calling the leashing event without checking if the player is 'allowed' to do it with that entity type. I think the type check should also be done here.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -997,13 +997,23 @@ public boolean entityInteract(GlowPlayer player, InteractEntityMessage message)
...
+            if (!InventoryUtil.isEmpty(handUten) && handUten.getType() == Material.LEASH) {
                 if (this.isLeashed() || EventFactory.callEvent(new PlayerLeashEntityEvent(this, player, player)).isCancelled()) {
                     return false;
                 }
 
+                if (handUten.getType() == Material.SADDLE && player.getGameMode() != GameMode.CREATIVE) {
```
aramperes: Is this the proper way to do this? I would've thought that this would be done in the `GlowPig` implementation...

Postremus: 😅 
Yeah, shit like that happens when you're not fully concentrated on one task.

aramperes:
Code looks good to me, but I only see 1 caveat: if you're in survival mode and you use a lead on an animal using the main hand, the lead is broken immediately - this doesn't happen with the off-hand for some reason.

Postremus:
resolved this issue with 7da64e9

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -997,8 +998,22 @@ public TaskManager getTaskManager() {
     public boolean entityInteract(GlowPlayer player, InteractEntityMessage message) {
         super.entityInteract(player, message);
 
+        if (message.getAction() != Action.INTERACT.ordinal()) {
```
aramperes: I think this check should be done inside the `isLeashed() && [...]` check, because you don't want to return `false` for every other action type that other entities can implement.

Postremus: I actually want to return false, since the super implementation does the same in every case: https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/entity/GlowEntity.java#L1395

aramperes: Ah, I see then.

aramperes (approved):

aramperes:
Looks like there was a little bug that we overlooked when we introduced the leashes: #549

PR #542, opened 2017-08-11
Implement Firework Entity

Postremus:
Hello,

This PR implements the firework_rocket entity, excluding proper physics.

Firework items can be used in the air (for boosting a player), as well as on blocks. Only one of this interactions was possible before. I changed this using the `Context` (and its value AIR_BLOCK) so that both interactions are possible. This replaces the old `canOnlyUseSelf` of ItemType.
Since 16w20a, the firework recipe produces 3 rockets.

### Physics during normal flight
The initial velocity is random for vx and vz, and static for vy. This matches the output that I got from paper (see sources), the values are most likely a bit imprecise though.

**During flight**, vx and vz are each multiplied by 1.15. This was calculated using 
* -0.001125*x^30=-0.086 =>  x = 1.5552
*  -0.0405*x^10=-0.164125 => x = 1.1502  

Vy is only increased by 0.04 each tick during flight, which appears to be a static value.

### Physics of boosting an entity
These are just assumptions.
When you are only boosted by one firework rocket, vx, vy, and vz never go above 1.5.
When you are boosted by multiple firework rockets, your velocity goes slightly above 1.5.

I played around till I found a way to replicate vanilla, but its off by like 1.1. Most likely air drag or gravity which needs to be applied.

related to #504

### Sources:
https://minecraft.gamepedia.com/Firework_Rocket
https://gist.github.com/Postremus/7223c6a153c3b9042bb63ed510c052ea
https://gist.github.com/Postremus/bef66401490048a5a1d42fcae8f55e75

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemType.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemType.java
@@ -168,16 +168,16 @@ public class ItemType {
      */
     public enum Context {
         /**
-         * The Item can only be used in Air
+         * The item can only be used when clicking in the air
          */
         AIR,
         /**
-         * The Item can only be used in conjunction with a block
+         * The item can only be used when clicking against a block
          */
         BLOCK,
         /**
-         * The Item can be used in Air and in conjunction with a block
+         * The item can be used on any click
          */
-        AIR_BLOCK
+        ANY
     }
 }
--- a/src/main/java/net/glowstone/block/itemtype/ItemType.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemType.java
@@ -133,11 +133,11 @@ public class ItemType {
     }
 
     /**
-     * If this item can only be used without any context (essentially used in air)
-     * @return If this item can only be used without any context
+     * Get the context this item can be used in
+     * @return context of the item, default is {{@link Context#BLOCK}}
      */
-    public boolean canOnlyUseSelf() {
-        return false;
+    public Context getContext() {
+        return Context.BLOCK;
     }
 
     /**
@@ -161,5 +161,23 @@ public class ItemType {
     @Override
     public final String toString() {
         return getClass().getSimpleName() + "{" + getId() + " -> " + getMaterial() + "}";
+    }
+
+    /**
+     * Context of the Items interaction
+     */
+    public enum Context {
+        /**
+         * The Item can only be used in Air
+         */
+        AIR,
+        /**
+         * The Item can only be used in conjunction with a block
+         */
+        BLOCK,
+        /**
+         * The Item can be used in Air and in conjunction with a block
+         */
+        AIR_BLOCK
     }
 }
--- a/src/main/java/net/glowstone/entity/passive/GlowFirework.java
+++ b/src/main/java/net/glowstone/entity/passive/GlowFirework.java
@@ -1,0 +1,204 @@
+package net.glowstone.entity.passive;
+
+import com.flowpowered.network.Message;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+import lombok.Getter;
+import lombok.Setter;
+import net.glowstone.EventFactory;
+import net.glowstone.entity.GlowEntity;
+import net.glowstone.entity.meta.MetadataIndex;
+import net.glowstone.net.message.play.entity.EntityMetadataMessage;
+import net.glowstone.net.message.play.entity.SpawnObjectMessage;
+import net.glowstone.util.InventoryUtil;
+import org.bukkit.EntityEffect;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.Sound;
+import org.bukkit.SoundCategory;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.EntityType;
+import org.bukkit.entity.Firework;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
+import org.bukkit.event.entity.FireworkExplodeEvent;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.FireworkMeta;
+import org.bukkit.util.Vector;
+
+public class GlowFirework extends GlowEntity implements Firework {
+
+    @Getter
+    @Setter
+    private ItemStack item;
+    @Getter
+    @Setter
+    private UUID spawningEntity;
+    @Getter
+    private LivingEntity boostedEntity;
+    /**
+     * The number of ticks before this fireworks rocket explodes.
+     */
+    @Getter
+    @Setter
+    private int lifeTime;
+
+    public GlowFirework(Location location) {
+        super(location);
+        setSize(0.25f, 0.25f);
+    }
+
+    public GlowFirework(Location location, UUID spawningEntity, LivingEntity boostedEntity, ItemStack item) {
+        super(location);
+        this.spawningEntity = spawningEntity;
+        setBoostedEntity(boostedEntity);
+        setSize(0.25f, 0.25f);
+
+        ThreadLocalRandom random = ThreadLocalRandom.current();
+        setVelocity(new Vector(random.nextGaussian() * 0.001, 0.05, random.nextGaussian() * 0.001));
+
+        this.item = InventoryUtil.itemOrEmpty(item).clone();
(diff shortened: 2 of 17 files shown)


src/main/java/net/glowstone/net/handler/play/player/UseItemHandler.java
```diff
@@ -15,17 +17,22 @@ public void handle(GlowSession session, UseItemMessage message) {
...
+                if (type.getContext() == Context.AIR || type.getContext() == Context.AIR_BLOCK) {
                     type.rightClickAir(player, holding);
                 } else {
                     if (holding.getType() == Material.WATER_BUCKET || holding.getType() == Material.LAVA_BUCKET) {
                         holding.setType(Material.BUCKET);
                     }
                 }
             }
+
+            if (!InventoryUtil.isEmpty(holding) && holding.getAmount() <= 0) {
```
aramperes: That check doesn't make much sense, if the amount is `<= 0` then `isEmpty` will return true, thus breaking the condition.

[This](https://github.com/GlowstoneMC/Glowstone/blob/d4c3c57ce6bf9587563aad64b7d53b1bc30df56d/src/main/java/net/glowstone/entity/passive/GlowCow.java#L35-L40) is how we usually do it.

src/main/java/net/glowstone/net/handler/play/player/UseItemHandler.java
```diff
@@ -15,17 +17,22 @@ public void handle(GlowSession session, UseItemMessage message) {
         GlowPlayer player = session.getPlayer();
         ItemStack holding = player.getItemInHand();
```
aramperes: Do fireworks work with off-hand? If so, we'll need to support it here.

src/main/java/net/glowstone/entity/passive/GlowFirework.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        this.item.setItemMeta(fireworkMeta.clone());
+        this.metadata.set(MetadataIndex.FIREWORK_INFO, item.clone());
+        this.lifeTime = calculateLifeTime(fireworkMeta.getPower());
+    }
+
+    private int calculateLifeTime(int power) {
+        ThreadLocalRandom random = ThreadLocalRandom.current();
+        return ((power + 1) * 10 + random.nextInt(0, 6) + random.nextInt(0, 7));
```
aramperes: Isn't `nextInt(0, n)` the same as `nextInt(n)`?

src/main/java/net/glowstone/entity/passive/GlowFirework.java
```diff
@@ -0,0 +1,204 @@
...
+import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
+import org.bukkit.event.entity.FireworkExplodeEvent;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.FireworkMeta;
+import org.bukkit.util.Vector;
+
+public class GlowFirework extends GlowEntity implements Firework {
+
+    @Getter
+    @Setter
```
aramperes, line 34: If you are storing the item inside the metadata, no need to keep an instance here. Use the get/set methods for the metadata to store this info (prevents having duplicate instances)

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -162,4 +162,22 @@ public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face,
     public final String toString() {
         return getClass().getSimpleName() + "{" + getId() + " -> " + getMaterial() + "}";
     }
+
+    /**
+     * Context of the Items interaction
+     */
+    public enum Context {
+        /**
+         * The Item can only be used in Air
```
aramperes: `The item can only be used when clicking in the air`

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -162,4 +162,22 @@ public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face,
...
+         */
+        AIR,
+        /**
+         * The Item can only be used in conjunction with a block
```
aramperes: `The item can only be used when clicking against a block`

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -162,4 +162,22 @@ public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face,
...
+         */
+        BLOCK,
+        /**
+         * The Item can be used in Air and in conjunction with a block
```
aramperes: `The item can be used when clicking either against a block or in the air`

src/main/java/net/glowstone/entity/passive/GlowFirework.java
```diff
@@ -0,0 +1,204 @@
...
+        this.lifeTime = calculateLifeTime(fireworkMeta.getPower());
+    }
+
+    private int calculateLifeTime(int power) {
+        ThreadLocalRandom random = ThreadLocalRandom.current();
+        return ((power + 1) * 10 + random.nextInt(0, 6) + random.nextInt(0, 7));
+    }
+
+    @Override
+    public void detonate() {
```
aramperes: I think this method should just make the firework explode immediately, unless I'm reading the Javadoc wrong.

aramperes (changes requested):

src/main/java/net/glowstone/entity/passive/GlowFirework.java
Postremus: Tested this. FireworkExplodeEvent is executed 1 Tick after detonate was called.

src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -162,4 +162,22 @@ public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face,
...
+         */
+        AIR_BLOCK
```
mastercoms: Strange name, perhaps `BOTH` or `ANY` could work better.

src/main/java/net/glowstone/entity/passive/GlowFirework.java
```diff
@@ -96,18 +90,39 @@ public void setFireworkMeta(FireworkMeta fireworkMeta) {
...
         }
 
+        ItemStack item = getFireworkItem();
+        item.setItemMeta(fireworkMeta.clone());
+        setFireworkItem(item);
+        this.lifeTime = calculateLifeTime(fireworkMeta.getPower());
+    }
+
+    /**
+     * Get
```
aramperes: I think you forgot something here :P

src/main/java/net/glowstone/entity/passive/GlowFirework.java
```diff
@@ -44,6 +41,7 @@
     @Getter
     @Setter
     private int lifeTime;
+    public static final ItemStack DEFAULT_FIREWORK_ITEM = new ItemStack(Material.FIREWORK);
```
aramperes: Can this be `private`?

src/main/java/net/glowstone/entity/passive/GlowFirework.java
Postremus: :1st_place_medal:

aramperes (approved):

src/main/java/net/glowstone/net/message/play/player/UseItemMessage.java
```diff
@@ -2,9 +2,14 @@
...
 import lombok.Data;
+import org.bukkit.inventory.EquipmentSlot;
 
 @Data
 public class UseItemMessage implements Message {
 
     private final int hand;
+
+    public EquipmentSlot getEquipmentSlot() {
+        return EquipmentSlot.values()[hand];
```
aramperes: I would do this instead: https://github.com/GlowstoneMC/Glowstone/blob/ea2763eb4f8fa2dd340b131a6f7ceba9ff233860/src/main/java/net/glowstone/net/message/play/player/BlockPlacementMessage.java#L14
That way if the client sends an incorrect hand ID, it will default to the main hand.

aramperes (changes requested):
See comment I added a few minutes ago.

Postremus:
resolved this in 2f0dbf47ad4d033993f0456dba076c4bd67fdcc0

aramperes (approved):

PR #546, opened 2017-08-11
Implement lightning related stuff

Postremus:
Implement LightningStrike.Spigot

Make entities transmute when hit by lightning
pig -> pigzombie
villager -> witch
creeper -> charged creeper

related to #504

### DIFF
--- a/src/main/java/net/glowstone/entity/EntityRegistry.java
+++ b/src/main/java/net/glowstone/entity/EntityRegistry.java
@@ -52,7 +52,7 @@ public class EntityRegistry {
                     .put(IronGolem.class, GlowIronGolem.class)
                     .put(Item.class, GlowItem.class)
                     .put(ItemFrame.class, GlowItemFrame.class)
-                    .put(LeashHitch.class, GlowLeashHitch.class)                     
+                    .put(LeashHitch.class, GlowLeashHitch.class)
                     .put(LightningStrike.class, GlowLightningStrike.class)
                     .put(Llama.class, GlowLlama.class)
                     .put(MagmaCube.class, GlowMagmaCube.class)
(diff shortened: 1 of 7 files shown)


src/main/java/net/glowstone/entity/EntityRegistry.java
```diff
@@ -52,8 +52,8 @@
                     .put(IronGolem.class, GlowIronGolem.class)
                     .put(Item.class, GlowItem.class)
                     .put(ItemFrame.class, GlowItemFrame.class)
-                    .put(LeashHitch.class, GlowLeashHitch.class)
-                    //TODO: Lightning
+                    .put(LeashHitch.class, GlowLeashHitch.class)
```
aramperes: Remove extra whitespace at end of line 55

src/main/java/net/glowstone/entity/GlowLightningStrike.java
```diff
@@ -39,10 +40,25 @@
      * Whether the lightning strike is just for effect.
      */
     private boolean effect;
+    /**
+     * Whether the lightning strike is silent
+     */
+    private boolean isSilent;
+    private LightningStrike.Spigot spigot = new LightningStrike.Spigot() {
```
aramperes: I think this can be final.

src/main/java/net/glowstone/entity/passive/GlowVillager.java
```diff
@@ -99,4 +102,14 @@ protected Sound getDeathSound() {
     protected Sound getAmbientSound() {
         return Sound.ENTITY_VILLAGER_AMBIENT;
     }
+
+    @Override
+    public void damage(double amount, Entity source, DamageCause cause) {
+        if (!DamageCause.LIGHTNING.equals(cause)) {
+            super.damage(amount, source, cause);
```
aramperes, line 109: This would call the EntityDamageEvent which can be cancelled by plugins -- we should probably have an internal method that returns false when that event is cancelled, so subsequent actions (like converting) can be cancelled too.

aramperes (changes requested):

src/main/java/net/glowstone/entity/passive/GlowVillager.java
Postremus: resolved this in 47cf385

aramperes (approved):

Issue #550, opened 2017-08-18
/tp player toplayer does not work


Postremus:
Hello,

I tried to teleport myself to an test account of mine using `/tp Postremus otheraccount`.
This did not work.
As output in the chat I got `Teleported Postremus to Postremus`, and was, presumably, teleported to myself.

Expected behavior is that I can teleport to another account using this syntax.

regards,
Postremus

Foorack:
@Postremus I find it weird that you were able to teleport a player to a player, as that does not seem to be implemented in the code. :thinking:

Postremus:
Ah, most likely the other account just got ignored, and something which resembles `/tp Postremus` got executed.

notourist:
Why are there two distinct teleport commands (/tp and /teleport)? Shouldn't they be the same?

mastercoms:
https://minecraft.gamepedia.com/Commands#teleport

https://minecraft.gamepedia.com/Commands#tp

aramperes:
Fixed in #554

PR #554, opened 2017-08-26
Fix issue #550 (/tp player toplayer does not work)

notourist:
As the title says, this PR fixes issue #550.


mastercoms:
Thank you for making my day, this made me smile 😄

PR #555, opened 2017-08-26
Implement effect command

notourist:
This adds the /effect command (see #499 ).

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -858,6 +858,7 @@ public final class GlowServer implements Server {
         commandMap.register("minecraft", new ToggleDownfallCommand());
         commandMap.register("minecraft", new SetWorldSpawnCommand());
         commandMap.register("minecraft", new PlaySoundCommand());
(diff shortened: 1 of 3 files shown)


aramperes:
Would be nice to add tab-completion to this, like we did with the other commands.

notourist:
I just checked the auto completion in vanilla Minecraft and all the effect names have a "minecraft:"-prefix. Should I leave this in or not, because currently the prefix gets removed.

aramperes:
I think we should leave it in.

This is what we did for the `/give` command:

https://github.com/GlowstoneMC/Glowstone/blob/09e8432fab1f18a377560f0035221c435dd4a842/src/main/java/net/glowstone/command/minecraft/GiveCommand.java#L95-L100

However, I see that you are getting the potion IDs using the enum values (`Impl.values()`). So basically, you would add a new field to this enum (e.g. `key`) with the actual vanilla ID, with the namespace (e.g. `minecraft:regeneration`). Then you can map the potion effects using these IDs instead.

mastercoms (approved):

mastercoms:
Thank you so much for your contribution!

PR #557, opened 2017-08-31
Fix shift-clicking in containers (Fixes #548)

### LINKED ISSUE #548: Shift-clicking a stack from a container to the inventory shows wrong item count
When shift-clicking a stack from a container (like a chest) into an inventory containing another (non-full) stack of the same type, the stacks will not combine: the clicked stack will use the default shift-click behavior (last available slot in hotbar), while the client expects the items to be merged into one.

This will temporarily make the inventory un-synced with the client because it expects vanilla behavior.

Here's a video of it in action: https://momoperes.ca/videos/issue_chest.mp4

I am using a build for commit ea2763e

puigru:
This PR changes the order in which ItemStacks are filled in order to match vanilla behaviour, fixing #548.

### DIFF
--- a/src/main/java/net/glowstone/inventory/GlowInventory.java
+++ b/src/main/java/net/glowstone/inventory/GlowInventory.java
@@ -200,7 +200,7 @@ public class GlowInventory implements Inventory {
         }
         ItemStack maxStack = stack.clone();
         maxStack.setAmount(stack.getMaxStackSize());
-        Integer firstEmpty = null;
+        int firstEmpty = -1;
         for (int s = 0; s < slots.length && stack.getAmount() > 0; s += 2) {
             // Iterate through all pairs of start and end slots
             int start = slots[s];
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/inventory/GlowInventory.java
```diff
@@ -200,6 +200,7 @@ public ItemStack tryToFillSlots(ItemStack stack, int... slots) {
         }
         ItemStack maxStack = stack.clone();
         maxStack.setAmount(stack.getMaxStackSize());
+        Integer firstEmpty = null;
```
mastercoms: Could this be an `int` instead? Just use -1 for an uninitialized value.

puigru: Aren't there instances where the slot number could be -1?
EDIT: There are, but they don't apply to this case, will change.

mastercoms (approved):

mastercoms:

PR #562, opened 2017-09-09
Fix Enderman carried item metadata not being set properly

aramperes:
This fixes an issue where the Enderman carried-data would be saved as a short instead of a `BlockID` (Integer) in the metadata entry.

However, it should be noted that storing the block data should be possible as well, but is not working for some reason. The protocol documentation states that the metadata entry should be stored as `id << 4 | data`. From my testing though, the client appears to just read the full int as the block ID, thus shifting the bits would show different blocks instead. 

For now, I have decided to let it at that (only storing the ID) until I can investigate more on that issue.


mastercoms:
Perhaps it uses the block palette.

http://wiki.vg/Chunk_Format#Global_and_section_palettes

PR #569, opened 2017-10-03
Memory usage optimization with chunks

aramperes:
The goal of this PR is to lower the usage of memory by certain objects. Using a profiler, I noticed that over 75% of the memory usage was made of `GlowChunk$Key` objects, which are used to index chunks by their coordinates in multiple places in the code.

Since objects take more place than primitive types like ints, I tried caching these keys by having a simple permanent hash attributed to them (`x * 31 + z`) and storing them in a concurrent map. I noticed that not only the amount of `GlowChunk$Key` objects was reduced considerably (from 5 million to 1600 in a fresh world), but the RAM usage in bytes was lowered because most of these objects were replaced with single integers.

I have also noticed that there are many `GlowBlock` objects being created and deleted constantly, probably because this type is never cached. I didn't try much in that area yet, but there might be some more stuff to fix as well.

I am no optimization expert, so I am just analyzing graphs over time here. Please feel free to suggest anything or review my attempts if they feel inappropriate.

PS: I didn't really take the time to properly format/beautify my code, sorry in advance :)

### DIFF
--- a/src/main/java/net/glowstone/chunk/GlowChunk.java
+++ b/src/main/java/net/glowstone/chunk/GlowChunk.java
@@ -282,7 +282,7 @@ public final class GlowChunk implements Chunk {
      * @param cy   the Y coordinate of the BlockEntity
      * @param cz   the Z coordinate of the BlockEntity
      * @param type the type of BlockEntity
-     * @return     The BlockEntity that was created.
+     * @return The BlockEntity that was created.
      */
     public BlockEntity createEntity(int cx, int cy, int cz, int type) {
         Material material = Material.getMaterial(type);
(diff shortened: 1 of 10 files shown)


aramperes:
I just realized I named the branch "chunk-opt**m**imization" by accident. This is why I shouldn't be working on stuff like this at 1 in the morning :)

src/main/java/net/glowstone/chunk/GlowChunk.java
```diff
@@ -755,7 +755,30 @@ public ChunkDataMessage toMessage(boolean skylight, boolean entireChunk) {
...
+        public int hashCode() {
+            return hashCode;
+        }
     }
 
+    public static final class ChunkKeyStore {
+        private static final ConcurrentHashMap<Integer, Key> keys = new ConcurrentHashMap<>();
+
+        public static Key get(int x, int z) {
+            int id = x * 31 + z;
```
Postremus: Can we maybe remove the duplication of this hashcode generation code?

Postremus (approved):
Looks mostly good so far.

src/main/java/net/glowstone/chunk/GlowChunk.java
aramperes: Sure

Postremus:
thank you.

PR #573, opened 2017-10-15
Implemented the /testfor command

smartboyathome:
As part of this, NBT tag matching was implemented so that the command can specify some subset of an entity's tags can be specified.

### DIFF
--- a/src/main/java/net/glowstone/command/CommandUtils.java
+++ b/src/main/java/net/glowstone/command/CommandUtils.java
@@ -116,6 +116,7 @@ public class CommandUtils {
         return builder.toString();
     }
 
+    // TODO: Move this into the Server class within Glowkit, and implement it with the GlowServer class.
     private static GlowWorld getDefaultWorld() {
         return (GlowWorld) Bukkit.getServer().getWorlds().get(0);
     }
--- a/src/main/java/net/glowstone/command/CommandUtils.java
+++ b/src/main/java/net/glowstone/command/CommandUtils.java
@@ -116,15 +116,19 @@ public class CommandUtils {
         return builder.toString();
     }
 
+    private static GlowWorld getDefaultWorld() {
+        return (GlowWorld) Bukkit.getServer().getWorlds().get(0);
+    }
+
     public static GlowWorld getWorld(CommandSender sender) {
         if (sender instanceof ConsoleCommandSender) {
-            return (GlowWorld) Bukkit.getServer().getWorlds().get(0); // get the default world
+            return getDefaultWorld();
         } else if (sender instanceof Entity) {
             return (GlowWorld) ((Entity) sender).getWorld();
         } else if (sender instanceof BlockCommandSender) {
             return (GlowWorld) ((BlockCommandSender) sender).getBlock().getWorld();
         }
-        return null;
+        return getDefaultWorld();
     }
 
     public static Location getLocation(CommandSender sender) {
@@ -133,7 +137,7 @@ public class CommandUtils {
         } else if (sender instanceof BlockCommandSender) {
             return ((BlockCommandSender) sender).getBlock().getLocation();
         }
-        return null;
+        return new Location(getDefaultWorld(), 0, 0, 0);
     }
 
     public static String getName(CommandSender sender) {
--- a/src/main/java/net/glowstone/command/minecraft/TestForCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/TestForCommand.java
@@ -82,7 +82,7 @@ public class TestForCommand extends VanillaCommand {
             }
         }
 
-        // Note: When command blocks are implemented, this should be updated to output the number of matching entities.
+        // TODO: When command blocks are implemented, this should be updated to output the number of matching entities.
         return true;
     }
 }
--- a/src/main/java/net/glowstone/command/minecraft/TestForCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/TestForCommand.java
@@ -1,0 +1,88 @@
+package net.glowstone.command.minecraft;
+
+import net.glowstone.command.CommandTarget;
+import net.glowstone.command.CommandUtils;
+import net.glowstone.entity.GlowEntity;
+import net.glowstone.entity.GlowPlayer;
+import net.glowstone.io.entity.EntityStorage;
+import net.glowstone.util.mojangson.Mojangson;
+import net.glowstone.util.mojangson.ex.MojangsonParseException;
+import net.glowstone.util.nbt.CompoundTag;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.command.CommandSender;
+import org.bukkit.command.defaults.VanillaCommand;
+import org.bukkit.entity.Entity;
+
+import java.util.*;
+
+public class TestForCommand extends VanillaCommand {
+    public TestForCommand() {
+        super("testfor",
+                "Tests for a certain target in game",
+                "/testfor <player> [dataTag]",
+                Collections.emptyList());
+        setPermission("minecraft.command.testfor");
+    }
+
+    @Override
+    public boolean execute(CommandSender sender, String label, String[] args) {
+        if (!testPermission(sender)) {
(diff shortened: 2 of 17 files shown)


src/main/java/net/glowstone/command/minecraft/TestForCommand.java
```diff
@@ -0,0 +1,88 @@
...
+        if (args.length < 1) {
+            sender.sendMessage(ChatColor.RED + "Usage:" + usageMessage);
+            return false;
+        }
+
+        String name = args[0];
+        Entity[] entities;
+        if (name.startsWith("@")) {
+            CommandTarget target = new CommandTarget(sender, name);
+            entities = target.getMatched(CommandUtils.getLocation(sender));
```
aramperes, line 43: Assuming the `/testfor` command can only be executed by physical senders (no console), you should validate that before (see other commands that use target selectors)

src/main/java/net/glowstone/command/minecraft/TestForCommand.java
```diff
@@ -0,0 +1,88 @@
...
+                    }
+                }
+            }
+        } else {
+            for (Entity entity : entities) {
+                sender.sendMessage("Found " + CommandUtils.getName(entity));
+            }
+        }
+
+        // Note: When command blocks are implemented, this should be updated to output the number of matching entities.
```
aramperes: Use `// TODO: ` here so we can locate this easier later on

src/main/java/net/glowstone/command/minecraft/TestForCommand.java
smartboyathome: I did some testing with the vanilla Minecraft server, and /testfor works fine from the console. Is there anything different I have to do in that case?

aramperes: If it can be executed by the console, I assume you can't use target selectors (like `@e`) as the console? IIRC you can only test-for players using the console.

smartboyathome: After more testing, `@e` does work in the console. Just spawned a bunch of zombies into my vanilla server, and ran `/testfor @e[r=500,type=zombie]`, and it gave me *[Server: Found Zombie]*. Looks like the command uses (0, 0, 0) as the location for the command when run in the console.

aramperes: That's interesting, I assumed it would take the target string as a literal player name if it's a console. 

Did you test this with Vanilla or Spigot? Since Vanilla can only have 1 world, it makes sense to have this arbitrary location. But since Spigot (and Glowstone) support multiple worlds, this may not be possible (unless it takes the default world too?)

smartboyathome: I was testing with the Vanilla server. Just spun up a Spigot server, spawned in some zombies, and tested the command, works there too. Again, its centered on (0, 0, 0) in the default overworld.

aramperes: Okay. So that means we should return `0 0 0` here:
https://github.com/GlowstoneMC/Glowstone/blob/47db4c461f2077620731692777c07e146af4d3f1/src/main/java/net/glowstone/command/CommandUtils.java#L136

and the default world here:

https://github.com/GlowstoneMC/Glowstone/blob/47db4c461f2077620731692777c07e146af4d3f1/src/main/java/net/glowstone/command/CommandUtils.java#L127

smartboyathome: Will do. Want me to clean up the other commands that use these functions and check for null, or do you want that to be in a separate PR?

aramperes: That's up to you, and what scope you want to set for your PR :)

smartboyathome:
Requested changes made, and cleanup done.

src/main/java/net/glowstone/command/CommandUtils.java
```diff
@@ -116,15 +116,19 @@ public static String prettyPrint(String[] strings) {
         return builder.toString();
     }
 
+    private static GlowWorld getDefaultWorld() {
```
aramperes: Although not part of the scope of this PR, I think it would be better to add this method to the `Server` interface (Glowkit) and implement it in the `GlowServer` class. Can you put a `todo` here for this?

aramperes:
Code looks good, I'll test the functionality tonight hopefully.

src/main/java/net/glowstone/command/CommandUtils.java
smartboyathome: Done.

src/main/java/net/glowstone/command/minecraft/TestForCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public boolean execute(CommandSender sender, String label, String[] args) {
+        if (!testPermission(sender)) {
+            return false;
+        }
+
+        if (args.length < 1) {
+            sender.sendMessage(ChatColor.RED + "Usage:" + usageMessage);
```
aramperes: Add a space after the colon in "Usage:"

src/main/java/net/glowstone/command/minecraft/TestForCommand.java
```diff
@@ -0,0 +1,88 @@
...
+            if (player == null) {
+                sender.sendMessage(ChatColor.RED + "Entity '" + name + "' cannot be found");
+                return false;
+            } else {
+                entities = new Entity[] {player};
+            }
+        }
+
+        if (args.length >= 2) {
+            String data = String.join(" ", Arrays.copyOfRange(args, 4, args.length));
```
aramperes: The `4` here is incorrect and produces this exception if the number of arguments is under 4:

```
org.bukkit.command.CommandException: Unhandled exception executing 'testfor @e[type=pig] {Saddle:1b}' in net.glowstone.command.minecraft.TestForCommand(testfor)
        at org.bukkit.command.SimpleCommandMap.dispatch(SimpleCommandMap.java:160)
Caused by: java.lang.IllegalArgumentException: 4 > 2
        at java.util.Arrays.copyOfRange(Arrays.java:3480)
        at java.util.Arrays.copyOfRange(Arrays.java:3441)
        at net.glowstone.command.minecraft.TestForCommand.execute(TestForCommand.java:60)
```

src/main/java/net/glowstone/util/nbt/CompoundTag.java
```diff
@@ -59,6 +59,38 @@ public void remove(String key) {
         value.remove(key);
     }
 
+    /**
+     * Checks to see if this tag is a strict, deep submap of the given CompoundTag.
+     * @param other The CompoundTag that should contain our values.
+     * @return
+     */
+    public boolean matches(CompoundTag other) {
```
aramperes, line 67: After testing the following command in my world with only 1 saddled pig: `/testfor @e[type=pig] {Saddle:1b}`, it still matched with pigs that weren't saddled. Either there is a problem with the logic in this function or with the way it is called in the command.

It would be a good idea to have a simple unit test to check if two compounds match using this function.

aramperes (changes requested):

smartboyathome:
Made the requested changes and fixed the bugs.

aramperes:
Thanks, will test asap.

gdude2002:
Your changes have broken two tests. https://circleci.com/gh/GlowstoneMC/Glowstone/1210

smartboyathome:
Yeah, I realized that after I had pushed. Fixed them in my specific test, although I don't think that Tag.equals() will work correctly for ByteArrayTag and IntArrayTag, since Array.equals() will check for instance equality, while Arrays.equals() checks for subelement equality.

aramperes (approved):

aramperes:
Thank you for your contribution! :)

By the way, you don't need to squash commits together to fix something. Having them separated makes it easier to review, and we squash everything when merging anyways.

PR #577, opened 2017-10-22
Pistons

picknchew:
This PR implements Pistons. From my testing, it seems that they work fine. There are still a few things missing like checking for non-pushable blocks, block breaking and directly powering the piston (right now the piston needs to be indirectly powered).

Even with a few things missing, I think it's a great start and can be merged.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockPiston.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockPiston.java
@@ -5,6 +5,8 @@ import net.glowstone.chunk.GlowChunk;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.net.message.play.game.BlockActionMessage;
 import org.bukkit.Material;
+import org.bukkit.Sound;
+import org.bukkit.SoundCategory;
 import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
@@ -66,13 +68,6 @@ public class BlockPiston extends BlockDirectional {
         if (me.isBlockIndirectlyPowered() && !isPistonExtended(me)) {
             List<Block> blocks = new ArrayList<>();
 
-            me.getWorld().getNearbyEntities(me.getLocation(), 20, 20, 20).stream().filter(entity -> entity.getType() == EntityType.PLAYER).forEach(player -> {
-                ((GlowPlayer) player).getSession().send(message);
-            });
-
-            // extended state for piston base
-            me.setData((byte) (me.getData() | 0x08));
-
             // get all blocks to be pushed by piston
             // add 2 to push limit to compensate for i starting at 1 and also to get the block after the push limit
             for (int i = 1; i < PUSH_LIMIT + 2; i++) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -35,17 +46,142 @@ public boolean isSticky() {
...
-            // spawn some smoke because we won't be working on pistons until 1.13
-            block.getWorld().playSound(block.getLocation(), Sound.BLOCK_LAVA_EXTINGUISH, 0.3F, 1.0F);
-            for (int i = 0; i < 3; i++) {
-                block.getWorld().spawnParticle(Particle.SMOKE_LARGE, block.getLocation().clone().add(0.5, 0.8, 0.5), 0);
+
+        if (!isPistonExtended(me)) {
+            return;
+        }
+
+        me.getWorld().getNearbyEntities(me.getLocation(), 20, 20, 20).stream().filter(entity -> entity.getType() == EntityType.PLAYER).forEach(player -> {
```
aramperes: Any reason for the arbitrary `20` blocks range? I think we use this logic usually:

https://github.com/GlowstoneMC/Glowstone/blob/f1d5f6f9fc38c88a414ff2bc60250f7ce151dbd5/src/main/java/net/glowstone/block/entity/BlockEntity.java#L44-L47

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -35,17 +46,142 @@ public boolean isSticky() {
...
+                case NORTH:
+                    return PistonDirection.NORTH;
+                case SOUTH:
+                    return PistonDirection.SOUTH;
+                case WEST:
+                    return PistonDirection.WEST;
+                case EAST:
+                    return PistonDirection.EAST;
+            }
+
```
aramperes: Remove empty line.

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -35,17 +46,142 @@ public boolean isSticky() {
...
+        ((GlowChunk) world.getChunkAt(block)).setMetaData(x & 0xf, z & 0xf, y, data);
+    }
+
+    public enum PistonDirection {
+        DOWN,
+        UP,
+        NORTH,
+        SOUTH,
+        WEST,
+        EAST;
```
aramperes: Put this all in one line (`DOWN, UP, ` etc.)

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -35,17 +46,142 @@ public boolean isSticky() {
...
+
+        public static PistonDirection getByBlockFace(BlockFace face) {
```
aramperes: Can you add some documentation about this method, if it needs to be public?

aramperes (changes requested):

aramperes (approved):

aramperes:
Good work! I'm going to have fun breaking this now ;)

PR #578, opened 2017-10-22
Fix food features and update food values

kaenganxt:
This pull request fixes some food related features and updates the food values like restored hunger, saturation and the applied potion effects.

The following things are done by this PR:
- Update restored hunger and saturation according to the [Minecraft wiki](https://minecraft.gamepedia.com/Food)
- New ItemSoup class which returns an empty bowl to the player
- Added beetroot, beetroot soup and chorus fruit as food
- Update potion effect chances, durations and levels according to the Minecraft wiki and some additional investigation
- Changed the food potion effects to use `new PotionEffect(...)` because the old version uses the "duration modifiers" (I have no clue why these exist) which change the given duration depending on the potion effect resulting in wrong durations applied to the player
- Use an `equals` in the eating logic as `getItemInHand` returns a copy of the itemstack object


aramperes (approved):

aramperes:
Thanks!

PR #579, opened 2017-10-22
Implement block states for commands and /testforblock command

aramperes:
This PR adds the system to parse and compare block state data as strings instead of numerical data values. This feature was added to some commands in 1.11, and allows using `color=orange` instead of the data value `1` for wool blocks, for example.

- Implements a parsing system to read such block state strings and compare them
  - Currently only supports Wool block state as proof-of-concept, needs to be expanded to [other blocks](https://minecraft.gamepedia.com/Block_states).
- Implements the `/testforblock` command, which uses this new functionality.
  - Sample usage: `/testforblock ~ ~-1 ~ minecraft:wool color=light_blue`
- Fixed an inconsistency with tilde notation (e.g. `~ ~-1 ~`), where the relative Y value would be shifted by 0.5 blocks like the X and Z axis. Some tests needed to be fixed to pass with this correction.
- Moved the block entity states inside the `net.glowstone.block.entity.state` package to avoid confusion.

I still need to add some documentation for the block state parsing system.

Related to #499

### DIFF
--- a/src/main/java/net/glowstone/block/state/BlockStateData.java
+++ b/src/main/java/net/glowstone/block/state/BlockStateData.java
@@ -1,12 +1,53 @@
 package net.glowstone.block.state;
 
-import java.util.HashMap;
+import com.google.common.collect.Maps;
 
-public class BlockStateData extends HashMap<String, String> {
+import java.util.Map;
+
+public class BlockStateData {
+
+    private final Map<String, String> map = Maps.newHashMap();
+    private final int numericValue;
+
+    public BlockStateData(int numericValue) {
+        this.numericValue = numericValue;
+    }
+
+    public BlockStateData() {
+        this.numericValue = -1;
+    }
+
+    public boolean isEmpty() {
+        return map.isEmpty();
+    }
+
+    public void put(String key, String value) {
+        map.put(key, value);
+    }
+
+    public boolean contains(String key) {
+        return map.containsKey(key);
+    }
+
+    public String get(String key) {
+        return map.get(key);
+    }
+
+    public boolean isNumeric() {
+        return numericValue != -1;
+    }
+
+    public int getNumericValue() {
+        return numericValue;
+    }
+
     @Override
     public String toString() {
+        if (isNumeric()) {
+            return String.valueOf(numericValue);
+        }
         StringBuilder builder = new StringBuilder();
-        for (Entry<String, String> entry : entrySet()) {
+        for (Map.Entry<String, String> entry : map.entrySet()) {
             builder.append(",").append(entry.getKey()).append("=").append(entry.getValue());
         }
         return builder.length() == 0 ? builder.toString() : builder.substring(1);
--- a/src/main/java/net/glowstone/block/state/BlockStateData.java
+++ b/src/main/java/net/glowstone/block/state/BlockStateData.java
@@ -1,0 +1,14 @@
+package net.glowstone.block.state;
+
+import java.util.HashMap;
+
+public class BlockStateData extends HashMap<String, String> {
+    @Override
+    public String toString() {
+        StringBuilder builder = new StringBuilder();
+        for (Entry<String, String> entry : entrySet()) {
+            builder.append(",").append(entry.getKey()).append("=").append(entry.getValue());
+        }
+        return builder.length() == 0 ? builder.toString() : builder.substring(1);
+    }
+}
--- a/src/main/java/net/glowstone/block/state/BlockStateReader.java
+++ b/src/main/java/net/glowstone/block/state/BlockStateReader.java
@@ -3,7 +3,11 @@ package net.glowstone.block.state;
 import org.bukkit.Material;
 import org.bukkit.material.MaterialData;
 
+import java.util.Set;
+
 public abstract class BlockStateReader<T extends MaterialData> {
+
+    public abstract Set<String> getValidStates();
 
     public abstract T read(Material material, BlockStateData data) throws InvalidBlockStateException;
 
--- a/src/main/java/net/glowstone/block/state/BlockStateReader.java
+++ b/src/main/java/net/glowstone/block/state/BlockStateReader.java
@@ -1,0 +1,11 @@
+package net.glowstone.block.state;
+
+import org.bukkit.Material;
+import org.bukkit.material.MaterialData;
+
+public abstract class BlockStateReader<T extends MaterialData> {
+
+    public abstract T read(Material material, BlockStateData data) throws InvalidBlockStateException;
+
+    public abstract boolean matches(BlockStateData state, T data) throws InvalidBlockStateException;
+}
--- a/src/main/java/net/glowstone/block/state/StateSerialization.java
+++ b/src/main/java/net/glowstone/block/state/StateSerialization.java
@@ -55,7 +55,7 @@ public class StateSerialization {
     }
 
     public static BlockStateReader getReader(Material material) {
-        if (!READERS.containsKey(material)) {
+        if (material == null) {
             return null;
         }
         return READERS.get(material);
--- a/src/main/java/net/glowstone/block/state/StateSerialization.java
+++ b/src/main/java/net/glowstone/block/state/StateSerialization.java
@@ -1,0 +1,78 @@
+package net.glowstone.block.state;
+
+import net.glowstone.block.state.impl.WoolStateDataReader;
+import org.bukkit.DyeColor;
+import org.bukkit.Material;
+import org.bukkit.material.MaterialData;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class StateSerialization {
+
+    private static final Map<Material, BlockStateReader> READERS = new HashMap<>();
+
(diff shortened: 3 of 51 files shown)


src/main/java/net/glowstone/block/state/impl/WoolStateDataReader.java
```diff
@@ -0,0 +1,49 @@
...
+            wool.setColor(color);
+        } else {
+            wool.setColor(DyeColor.WHITE);
+        }
+        return wool;
+    }
+
+    @Override
+    public boolean matches(BlockStateData state, Wool data) throws InvalidBlockStateException {
+        if (state.containsKey("color")) {
```
smartboyathome: Just checking whether the BlockStateData contains a key actually leads to the wrong results. This won't catch the case where BlockStateData contains an invalid property for this material. In this case, I would recommend copying all the keys into a new ArrayList<String>, and then while looping over the iterator. This way, whenever a key is successfully matched, you can call iterator.remove() to remove that key from the array. The matches() method then can only return true when the keys list is empty, ensuring that this case is caught.

src/main/java/net/glowstone/block/state/BlockStateReader.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Set;
+
+public abstract class BlockStateReader<T extends MaterialData> {
+
+    public abstract Set<String> getValidStates();
+
+    public abstract T read(Material material, BlockStateData data) throws InvalidBlockStateException;
+
+    public abstract boolean matches(BlockStateData state, T data) throws InvalidBlockStateException;
```
smartboyathome, line 14: I feel like sticking this on the reader is a bit disjoint with how I implemented the Entity's tag matching. I'm not opposed to sticking it on the reader, but and if you want to keep it like this you should think about changing how the CompoundTag's matching works. Otherwise, I'd say you should implement the match method on the BlockStateData, similar to how CompoundTag implements its match method.

src/main/java/net/glowstone/block/state/BlockStateData.java
```diff
@@ -0,0 +1,14 @@
+package net.glowstone.block.state;
+
+import java.util.HashMap;
+
+public class BlockStateData extends HashMap<String, String> {
```
smartboyathome: Extending an implementation of Map<> seems like the wrong way to go here. In my experience, this makes it harder to refactor, especially if another implementation of map is needed, and can dilute the reason for having BlockStateData vs a regular map. It's convenient, yes, but often introduces a design smell. I usually prefer introducing the map as a field and exposing only the relevant delegate methods, making it clear what is going on.

src/main/java/net/glowstone/block/state/StateSerialization.java
```diff
@@ -0,0 +1,85 @@
...
+    public static BlockStateData parse(Material material, String state) throws InvalidBlockStateException {
+        if (state == null || state.trim().isEmpty() || state.trim().equals("*")) {
+            return new BlockStateData();
+        }
+        BlockStateReader reader = getReader(material);
+        if (material == null || reader == null) {
+            throw new InvalidBlockStateException(material, state);
+        }
+        BlockStateData data = new BlockStateData();
+        String[] split = state.trim().split(",");
```
smartboyathome, line 24: This is a case where I'd say a small Regex could make the expected syntax a bit clearer, and reduce the number of cases you need to check to throw InvalidBlockStateException. Looks like the regex you want is `"^\\s*((\\w+)\\s*=\\s*(\\w+))\\s*(,\\s*(\\w+)\\s*=\\s*(\\w+)\\s*)*$"`. Looks intimidating, yes, but that is just keeping compatibility with the trimmed code above.

To use this regex, you'd be iterating through all the groups (starting at index 1 because the group at index 0 *always* matches the entire string), but taking only the second and third group out of every three, as the first will represent the entire "key=value" match.

src/main/java/net/glowstone/command/minecraft/TestForBlockCommand.java
```diff
@@ -0,0 +1,117 @@
...
+    public boolean execute(CommandSender sender, String label, String[] args) {
+        if (!testPermission(sender)) {
+            return false;
+        }
+        if (args.length < 4) {
+            sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
+            return false;
+        }
+        String itemName = args[3];
+        if (!itemName.startsWith("minecraft:")) {
```
smartboyathome, line 43: I'm not as familiar with the plugin side of things, but if a plugin introduces a new block, will it still start with "minecraft:"? If not, then this check would seem to fail when trying to test for a "foo-plugin:bar" block.

src/main/java/net/glowstone/command/minecraft/TestForBlockCommand.java
```diff
@@ -0,0 +1,117 @@
...
+        Material type = ItemIds.getItem(itemName);
+        Location location = CommandUtils.getLocation(CommandUtils.getLocation(sender), args[0], args[1], args[2]);
+        GlowBlock block = (GlowBlock) location.getBlock();
+        if (block.getType() != type) {
+            sender.sendMessage(ChatColor.RED + "The block at " + location.getBlockX() + ", " + location.getBlockY() + ", " + location.getBlockZ() +
+                    " is " + ItemIds.getName(block.getType()) + " (expected: " + ItemIds.getName(type) + ")");
+            return false;
+        }
+        if (args.length > 4) {
+            String state = args[4];
```
smartboyathome, line 55: Since we know this same "check the data value or state argument" is used in at least three commands (clone and testforblock, and testforblocks), mind abstracting this check out into a CommandUtils method? May want to return a custom type with two properties, `Integer data` and `BlockStateData stateData`, for ease of use in these commands.

src/main/java/net/glowstone/command/minecraft/TestForBlockCommand.java
```diff
@@ -0,0 +1,117 @@
...
+                        return false;
+                    }
+                } catch (InvalidBlockStateException e) {
+                    sender.sendMessage(ChatColor.RED + e.getMessage());
+                    return false;
+                }
+            }
+        }
+        if (args.length > 5 && block.getBlockEntity() != null) {
+            String dataTag = String.join(" ", new ArrayList<>(Arrays.asList(args)).subList(5, args.length));
```
smartboyathome, line 80: Given we do this CompoundTag check in a few commands as well now, we might want to think about also abstracting this out into a CommandUtils method.

smartboyathome (changes requested):
There you go, that's my feedback.

src/main/java/net/glowstone/block/state/BlockStateData.java
aramperes: Good points, I will change this.

src/main/java/net/glowstone/block/state/StateSerialization.java
aramperes: I hadn't thought of using a regex here, I will consider it. My only concern would be readability, but that can be resolved with a comment.

src/main/java/net/glowstone/command/minecraft/TestForBlockCommand.java
aramperes: You can't register new blocks/items with Bukkit. The ids are registered in the ItemIds class and all start with the minecraft namespace. Adding this functionality is outside of the scope of the PR.

src/main/java/net/glowstone/command/minecraft/TestForBlockCommand.java
aramperes: Good point, but it's outside of the scope of the PR. I'll consider it once this is merged.

src/main/java/net/glowstone/command/minecraft/TestForBlockCommand.java
aramperes: Will do.

src/main/java/net/glowstone/block/state/impl/WoolStateDataReader.java
aramperes: Doesn't [`getValidStates()`](https://github.com/GlowstoneMC/Glowstone/blob/states/src/main/java/net/glowstone/block/state/StateSerialization.java#L35) solve this?

src/main/java/net/glowstone/block/state/StateSerialization.java
aramperes: After checking, the regular expression you provided doesn't work all the time. For the string `a=b,c=d,e=f`:

Full match | `a=b,c=d,e=f`
-- | --
Group 1 | `a=b`
Group 2 | `a`
Group 3 | `b`
Group 4 | `,e=f`
Group 5 | `e`
Group 6 | `f`

This would skip the middle section.

aramperes:
I need approval to merge this for the release of 2017.10 on Wednesday. I can address the documentation and further design changes later on.

mastercoms (approved):

aramperes:
Merged in 18a3b49adaa3d3b1553fa54951f887036d344b91.

PR #581, opened 2017-10-25
Use ThreadLocalRandom

picknchew:
This PR removes _most_ local instantiations of Random and replaces it with ThreadLocalRandom. This will improve performance and get rid of the _random_ Random instances that are scattered everywhere.

### DIFF
--- a/src/main/java/net/glowstone/Explosion.java
+++ b/src/main/java/net/glowstone/Explosion.java
@@ -1,12 +1,5 @@
 package net.glowstone;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Random;
-import java.util.Set;
-import java.util.stream.Collectors;
 import net.glowstone.block.GlowBlock;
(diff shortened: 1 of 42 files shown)


mastercoms:
This is nice, thank you for your work. However, I'm not sure about replacing the world's random instance. Are there any downsides to the world's random generation possibly not sharing the same seed?

Also, just to make sure, did you replace any random instances with a specific seed?

picknchew:
I purposely did not change the world generation side of things. Most random's that were changed did not have seeds set. However, a few of them I found was not useful.

mastercoms:
Sorry for the confusion, when I said "world random generation", I didn't mean world generation, but rather the generation of random values within the GlowWorld class, previously generated by a single Random instance. Does it matter that these cases have no guarantee of sharing the same seed?

picknchew:
From what I could tell, no. It does not matter. ThreadLocalRandom generates only 1 seed and it's shared across the thread.

aramperes (approved):
Reviewed and approved by @kamcio96

PR #582, opened 2017-10-27
Add placement of chorus flower and plant, eating of chorus fruit

kaenganxt:
This pull request implements the chorus plant items/blocks.

1. Chorus fruit
The chorus fruit is edible and teleports the player to a random location as described in the [Minecraft Wiki](https://minecraft.gamepedia.com/Chorus_Fruit). The location check currently doesn't check for the bounding box of the player to be empty, but rather centers the location and checks for two air blocks.

2. Chorus flower/plant
Implement placing restrictions and drops for the chorus flower and plant blocks.
This does not include dropping the plant if certain conditions are met.

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemChorusFruit.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemChorusFruit.java
@@ -1,5 +1,6 @@
 package net.glowstone.block.itemtype;
 
+import java.util.concurrent.ThreadLocalRandom;
 import net.glowstone.entity.GlowPlayer;
 import org.bukkit.Location;
 import org.bukkit.Sound;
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/block/itemtype/ItemChorusFruit.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public boolean eat(GlowPlayer player, ItemStack item) {
+        if (!super.eat(player, item)) {
+            return false;
+        }
+
+        for (int i = 0; i < 16; i++) { //16 attempts: +/- 8 blocks in every direction
+            Location attempt = player.getLocation();
+            double deltaX = Math.random() * 16 - 8;
```
mastercoms: Use `ThreadLocalRandom.current().nextDouble()` instead of `Math.random()`

src/main/java/net/glowstone/block/itemtype/ItemChorusFruit.java
```diff
@@ -0,0 +1,74 @@
...
+        return true;
+    }
+
+    private Location getSafeLocation(Location loc) {
+        int yCoord = loc.getBlockY();
+        World world = loc.getWorld();
+        if (yCoord > world.getHighestBlockYAt(loc)) {
+            yCoord = world.getHighestBlockYAt(loc) + 1;
+        }
+        boolean found = false;
```
mastercoms, line 49: What is this loop for? Getting the highest y block should give you a coordinate with enough space for the player, as all blocks above the highest block are air.

mastercoms (changes requested):

src/main/java/net/glowstone/block/itemtype/ItemChorusFruit.java
kaenganxt: The y coordinate is randomly generated and may be under the ground (or the player eats underground). Vanilla then selects the next possible location lying under the selected target. So the final target is never above the selected random location.

mastercoms (approved):

aramperes:

PR #585, opened 2017-11-01
Fix #584 - ConcurrentModificationException

### LINKED ISSUE #584: ConcurrentModificationException, while spawning zombie
> 2017/10/31 17:01:57 [SCHWERWIEGEND] Error occurred while pulsing world world
	at java.util.ArrayList.forEach(Unknown Source)
	at net.glowstone.entity.ai.TaskManager.pulse(TaskManager.java:66)
	at net.glowstone.entity.GlowEntity.pulse(GlowEntity.java:470)
	at net.glowstone.entity.GlowLivingEntity.pulse(GlowLivingEntity.java:184)
	at net.glowstone.entity.monster.GlowMonster.pulse(GlowMonster.java:54)
	at net.glowstone.GlowWorld.pulse(GlowWorld.java:395)
	at net.glowstone.scheduler.WorldScheduler$WorldThread.run(WorldScheduler.java:139)
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)

Glowstone build(compiled locally): 2017.10.0-SNAPSHOT.6ecd65a (MC: 1.12.2)
Glowstone build(official): 2017.10.0-SNAPSHOT.18a3b49

Steps to reproduce:
1.  Spawn a zombie with a spawnegg
2. Hit the zombie(so i tried to force him to move)
3. Wait for movements
(4. Sometimes i timed out, after placing 2 zombies)

EDIT: Would someone please edit the title to "ConcurrentModificationException caused by zombie movements" or something similiar?

Microsamp:
We should use a CopyOnWriteArrayList to prevent ConcurrentModificationExceptions, while iterating and removing/adding some elements to the list.

Ps: I tried to implement/test ReadWriteLocks, but the performance and stability(deadlocks etc.) wasn't that great.

 https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArrayList.html

### DIFF
--- a/src/main/java/net/glowstone/entity/ai/EntityTask.java
+++ b/src/main/java/net/glowstone/entity/ai/EntityTask.java
@@ -4,7 +4,7 @@ import net.glowstone.entity.GlowLivingEntity;
 
 import java.util.concurrent.ThreadLocalRandom;
 
-public abstract class EntityTask {
+public abstract class EntityTask implements Comparable<EntityTask> {
     private final String name;
     private boolean executing = false;
     private int duration = 0;
@@ -12,6 +12,11 @@ public abstract class EntityTask {
 
     public EntityTask(String name) {
         this.name = name;
+    }
+
+    @Override
+    public int compareTo(EntityTask other) {
+        return 0;
     }
 
     public final void pulse(GlowLivingEntity entity) {
--- a/src/main/java/net/glowstone/entity/ai/EntityTask.java
+++ b/src/main/java/net/glowstone/entity/ai/EntityTask.java
@@ -16,7 +16,7 @@ public abstract class EntityTask implements Comparable<EntityTask> {
 
     @Override
     public int compareTo(EntityTask other) {
-        return 0;
+        return 0; // TODO: AI task priority
     }
 
     public final void pulse(GlowLivingEntity entity) {
--- a/src/main/java/net/glowstone/entity/ai/TaskManager.java
+++ b/src/main/java/net/glowstone/entity/ai/TaskManager.java
@@ -2,17 +2,18 @@ package net.glowstone.entity.ai;
 
 import net.glowstone.entity.GlowLivingEntity;
 
-import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 public class TaskManager {
 
     private final GlowLivingEntity entity;
-    private final ArrayList<EntityTask> tasks;
+    private final List<EntityTask> tasks;
 
     public TaskManager(GlowLivingEntity entity) {
         this.entity = entity;
-        this.tasks = new ArrayList<>();
+        this.tasks = new CopyOnWriteArrayList<>();
     }
 
     public EntityTask getTask(String name) {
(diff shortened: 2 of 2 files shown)


aramperes:
Hello @Microsamp,

Thank you for your contribution. Please make sure you sign the Contributor License Agreement before we review your pull request. See the comment left by the CLA assistant bot above.

Microsamp:
Hello @momothereal,

i signed the Contributor License Agreement.

aramperes:
I'm worried about the possible performance hit this would cause. Keep in mind that each entity has a Task Manager being pulsed every tick, and this solution appears to be quite costly according to the docs.

aramperes:
What about using `ConcurrentSkipListSet`? From what I understand, it is safely iterable, but it may happen that a task won't be pulsed until the next tick after it is added.

Microsamp:
> What about using ConcurrentSkipListSet?

Mhmmm, it seems to be a better choice for performance.

> but it may happen that a task won't be pulsed until the next tick after it is added.

I think it doesn't matter if a task is executed one tick later.

Microsamp:
WIP

EDIT: The ConcurrentSkipListSet throwed an undocumented(https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentSkipListSet.html) exception:
>Caused by: java.lang.ClassCastException: net.glowstone.entity.ai.LookAtPlayerTask cannot be cast to java.lang.Comparable
at java.util.concurrent.ConcurrentSkipListMap.cpr(Unknown Source)
at java.util.concurrent.ConcurrentSkipListMap.doPut(Unknown Source)
at java.util.concurrent.ConcurrentSkipListMap.putIfAbsent(Unknown Source)
at java.util.concurrent.ConcurrentSkipListSet.add(Unknown Source)
at net.glowstone.entity.ai.TaskManager.addTask(TaskManager.java:85)`

I think if we want to use this Set, we have to implement Comparable to every EntityTask.

Ps: Sry for wrong push(to my fork). I accidentally used an old jar...

aramperes:
I guess we could use the Comparable class to prioritize tasks (ordering), but that it is probably out of the scope of this PR.

Microsamp:
The question is: Should i fix the exception with the first commit, should i prioritize all EntityTasks with 0 or should i wait for a full concept of your idea

aramperes:
If it is outside of the scope, you can put a placeholder to fix the exception (for instance, using 0 for the Comparable impl), as long as it kinda works.

Microsamp:
So: I implemented your workaround with ConcurrentSkipListSet and default priority(comparable#compareTo) of zero.

I spawned many zombies(20-30) and hitted them. The zombies moved around and died(diamond-sword) after 2 or 3 punches.
After killing all zombies the thread number(/glowstone) is now 18(normal).

src/main/java/net/glowstone/entity/ai/EntityTask.java
```diff
@@ -14,6 +14,11 @@ public EntityTask(String name) {
         this.name = name;
     }
 
+    @Override
+    public int compareTo(EntityTask other) {
+        return 0;
```
aramperes: Add a todo here (`// todo: AI task priority`)

aramperes (changes requested):

aramperes (approved):

aramperes:

PR #586, opened 2017-11-06
Removed broken chunk key optimization, use longs instead

smartboyathome:
As discussed in discord, the chunk key class implemented its caching mechanism wrong. By doing (x * 31 + z), it was confusing the chunks at (1, 0) and (0, 31) for each other. This, I observed, lead to corruption in the saved chunks. I ported over Momo's use of longs for chunk keys, which in my testing is not observedly slower than the existing code.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -11,6 +11,7 @@ import net.glowstone.chunk.ChunkManager;
 import net.glowstone.chunk.ChunkManager.ChunkLock;
 import net.glowstone.chunk.ChunkSection;
 import net.glowstone.chunk.GlowChunk;
+import net.glowstone.chunk.GlowChunk.Key;
 import net.glowstone.chunk.GlowChunkSnapshot.EmptySnapshot;
 import net.glowstone.constants.*;
 import net.glowstone.data.CommandFunction;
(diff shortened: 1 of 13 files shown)


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -251,14 +250,14 @@ public LightningStrike strikeLightningEffect(Location loc, boolean isSilent) {
      * Per-chunk spawn limits on various types of entities.
      */
     private int monsterLimit, animalLimit, waterAnimalLimit, ambientLimit;
-    private Map<Integer, GlowStructure> structures;
+    private Map<Long, GlowStructure> structures;
```
LuminousLoreBloom: Perhaps TLongObjectMap could be better here?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -251,14 +250,14 @@ public LightningStrike strikeLightningEffect(Location loc, boolean isSilent) {
...
 
     /**
      * The maximum height at which players may place blocks.
      */
     private int maxBuildHeight;
 
-    private Set<Key> activeChunksSet = new HashSet<>();
+    private Set<Long> activeChunksSet = new HashSet<>();
```
LuminousLoreBloom: or TLongSet here?

src/main/java/net/glowstone/GlowWorld.java
smartboyathome: You're talking about the trove stuff, correct? I hadn't seen it used anywhere else in the code base, and this is already pretty lean as it is. I try not to do too much premature optimization, and in this case I don't think we'll be storing millions of items in the map. In addition, while Trove is stable, it seems within the last couple years [other libraries](http://java-performance.info/hashmap-overview-jdk-fastutil-goldman-sachs-hppc-koloboke-trove-january-2015/) outperform it.

smartboyathome:
After much discussion, I scaled back the change so it only affected GlowChunk.ChunkKeyStore and GlowChunk.Key, preserving them rather than replacing them. We can observe how this affects memory/performance, and go from there.

In addition, I had to make a tweak to GlowPlayer due to some changes introduced in Glowkit that prevented Glowstone from compiling.

mastercoms (approved):

PR #587, opened 2017-11-22
Item functions

mastercoms:
This PR will remove the item and block hierarchy, which has many problems with conflicting item and block behaviors and creates a bad system for modding support, which we may want to pursue in the future.

It uses functions to provide default item and block functionality, and provides a simple system mods and plugins can use to extend this functionality and add their own items/blocks with unique functionality in the future.


Pr0methean:
I'll review this tonight or tomorrow.

Pr0methean:
This looks to me a little too much like Aspect-Oriented Programming. How much use are stack traces going to be under this new regime?

mastercoms:
Do you think this could be better implemented in Java 9's [Flow API](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html)?

Also, I feel like it would make more sense to migrate to 1.13 first before doing such a massive change to our item/block type system.

mastercoms:
Closed because we're waiting for 1.13 and this branch will probably be rewritten by then

mastercoms:
Reopening this branch since it has become relevant again.

PR #588, opened 2017-11-23
Enforce Javadocs on all public methods.

aramperes:
I suggest we enforce having documentation for **public, non-override** methods. Many methods and classes are declared as public and do not have any description about what they do ~and some of them are kinda cryptic~. Some of these methods could also be private/protected instead of documented.

As of writing this there are 960 violations of this rule. The checkstyle rule excludes private/protected/override methods, as well as constructors.

Maybe? Yes/no?

### DIFF
--- a/etc/checkstyle.xml
+++ b/etc/checkstyle.xml
@@ -127,6 +127,7 @@
         <module name="JavadocMethod">
             <property name="scope" value="public"/>
             <property name="tokens" value="METHOD_DEF"/>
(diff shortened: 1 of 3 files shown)


mastercoms:
I'd like to suggest looking at the Google checkstyle that I have introduced in another branch. https://github.com/GlowstoneMC/Glowstone/commit/a657f994ac6154cdd98744862f06f6507586a5a3#diff-23868c392fcee3be938534ee2acec4c8R166

mastercoms (approved):

PR #589, opened 2017-11-24
Fix fire not getting updated. Fixes #435

### LINKED ISSUE #435: Fire is not affected by rain
Title: Fire is not affected by rain

Glowstone build: 8a57f8a

After setting Grass and Tallgrass on fire they stayed lit (while tunderstorm was raging).
http://puu.sh/umjzl/732b4553b8.png

Might be related to #130

kumpelblase2:
After causing a fire through flint and steel, fire did not behave properly: it just sat there doing nothing. The reason for this behavior was due to setting the type to the material the player was currently holding. Since it was flint and steel, it was set to that instead of a fire block. Running an update on that obviously does nothing.

Setting the block type to fire on the other hand does the right thing. Updates get propagated, it will spread and at some point stop.

The fire spreading seems to be a little off as well, but I left that for another day.

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
@@ -15,6 +15,10 @@ import org.bukkit.util.Vector;
 
 public class ItemFlintAndSteel extends ItemTool {
 
+    public ItemFlintAndSteel() {
+        this.setPlaceAs(Material.FIRE);
+    }
+
     @Override
     public boolean onToolRightClick(GlowPlayer player, GlowBlock target, BlockFace face, ItemStack holding, Vector clickedLoc, EquipmentSlot hand) {
         if (target.getType() == Material.OBSIDIAN) {


mastercoms (approved):

mastercoms (commented):
I'm not sure that is this is the right fix, since fire ticking was working before a few tick system updates. I'd rather try to find the root of the issue rather than workaround it.

However, maybe this is a good fix, at least for now. I'll have to investigate it further though.

Regardless, thank you for your contribution!

aramperes (approved):

kumpelblase2:
The reason it currently didn't run the update is the following. After right clicking, the block location got scheduled for receiving ticks and the type of the block was set to what the player was holding (which is wrong in this case). Once the next tick comes in, we reach `GlowWorld#pulseTickMap`, spicifically https://github.com/GlowstoneMC/Glowstone/blob/master/src/main/java/net/glowstone/GlowWorld.java#L1951 . `getMaterial` here returns "Flint & Steel", which is wrong (and this PR fixes that). Because it returns that instead of "Fire", the `itemTable.getBlock` returns null as "Flint & Steel" has no block type (which is correct). Since there's no block type for this location the system just thinks there's nothing here and stops progressing ticks for that location.

I didn't run this with fire created by the environment, which might still work without this fix. But fire created by flint and steel certainly didn't.

So I do acutally believe the system behaved correctly, given that we somewhat fed it wrong info. That it worked before seems to be more of a coincidence in this case. And therefor I wouldn't consider this a workaround but rather the correct fix here.

mastercoms:
If that's the issue, I think the fix should be setting flint and steel's placeAs to fire.

aramperes:
Can confirm that overriding ItemFlintAndSteel's `placeAs` to BlockFire resolves this problem.

aramperes (changes requested):
See comment.

kumpelblase2:
Alright, that does sound like a better solution. Updated PR accordingly.

aramperes (approved):

mastercoms:
Thank you very much for your contribution to Glowstone!

PR #590, opened 2017-11-26
First pass at implementing the /clone command

smartboyathome:
So, this is a big one that requires a lot of manual testing (tried writing some unit tests, but the mocks just weren't that easy to create). This implementation shouldn't require too much memory when executing, due to the dynamic direction of the iteration in order to avoid any overlap. By making it dynamic for each axis, I was able to clone without staging copies of each block in memory first. As mentioned, I did as much manual testing as I could, but doubtless I have missed several scenarios that may break this. More eyes are needed on this change before it should be merged.


src/main/java/net/glowstone/command/minecraft/CloneCommand.java
```diff
@@ -0,0 +1,338 @@
...
+        public BackwardsAxisIterator(int length) {
+            this.current = length;
+        }
+
+        @Override
+        protected Integer computeNext() {
+            if (current >= 0) {
+                // Could use a post-decrement operator here, but this is a bit more clear in getting the meaning across.
+                int retval = current;
+                current -= 1;
```
aramperes: I understand why you'd prefer not using `int retval = current--;`, but maybe replace the `current -= 1` line with `current--;`

src/main/java/net/glowstone/command/minecraft/CloneCommand.java
```diff
@@ -0,0 +1,338 @@
...
+            return false;
+        }
+
+        if (args.length < 9) {
+            sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
+            return false;
+        }
+
+        if (!CommandUtils.isPhysical(sender)) {
+            sender.sendMessage("Cannot access blocks outside of the world");
```
aramperes: I think the error message and the if-statement don't match here. (e.g. should be `This command may only be executed by physical objects` or something. I think other commands have this message somewhere.

src/main/java/net/glowstone/command/minecraft/CloneCommand.java
```diff
@@ -0,0 +1,338 @@
...
+                if (args.length >= 12) {
+                    Material blockType = ItemIds.getItem(args[11]);
+                    if (args.length >= 13) {
+                        Byte data = null;
+                        try {
+                            data = Byte.parseByte(args[12]);
+                        } catch (NumberFormatException ignored) {
+                        }
+                        if (data == null || 0 > data || data > 15) {
+                            sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
```
aramperes: Give a more precise error here (e.g. `Invalid data: <data>`)

src/main/java/net/glowstone/command/minecraft/CloneCommand.java
```diff
@@ -0,0 +1,338 @@
...
+                        blockFilter = new FilteredBlockFilter(blockType);
+                    }
+                } else {
+                    sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
+                    return false;
+                }
+                break;
+
+            default:
+                sender.sendMessage(ChatColor.RED + "Usage: " + usageMessage);
```
aramperes, line 167: Also here, say something like `Unknown block filter: <filter>`

src/main/java/net/glowstone/command/minecraft/CloneCommand.java
smartboyathome: Yeah, my biggest problem with the post- and pre- increment/decrement operators is that, in variable assignments, its not obvious which returns the old value and which returns the new value. I can definitely change that line to be current-- though.

src/main/java/net/glowstone/command/minecraft/CloneCommand.java
```diff
@@ -0,0 +1,338 @@
...
+        }
+    }
+
+    public CloneCommand() {
+        super(
+                "clone",
+                "Clones a section of the world.", "/clone <x1> <y1> <z1>  <x2> <y2> <z2>  <x> <y> <z> [maskMode] [cloneMode] [tileName] [tileData]",
+                Collections.emptyList()
+        );
+        setPermission("minecraft.command.seed");
```
aramperes: Wrong permission :^)

src/main/java/net/glowstone/command/minecraft/CloneCommand.java
```diff
@@ -0,0 +1,338 @@
...
+                            fromBlock.setType(Material.AIR, false);
+                        }
+
+                        blocksCloned++;
+                    }
+                }
+            }
+        }
+
+        sender.sendMessage("Cloned " + blocksCloned + " blocks.");
```
aramperes: If `blocksCloned` ends up being 0, should another message pop up here? (e.g. No blocks were cloned)

mastercoms (approved):

PR #597, opened 2017-12-08
Improve hanging entities

kaenganxt:
This pull request fixes several bugs and implements some events regarding item frames and paintings.

The following bugs/features are fixed/implemented:
- Call `HangingBreakByEntityEvent` when player breaks the item frame/painting
- Don't drop item frame if player is in creative
- Call `EntityDamageByEntityEvent` when item is removed from item frame (equals spigot behaviour)
- Only drop item frame content if player is in creative
- Don't reset rotation when item is removed or placed (equals vanilla behaviour)
- Remove useless/wrong check for air behind a painting, since the call to `isObstructed()` is executed before that and performs a wider check

### DIFF
--- a/src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
+++ b/src/main/java/net/glowstone/entity/objects/GlowItemFrame.java
@@ -22,6 +22,9 @@ import org.bukkit.Rotation;
 import org.bukkit.block.BlockFace;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.ItemFrame;
+import org.bukkit.event.entity.EntityDamageByEntityEvent;
+import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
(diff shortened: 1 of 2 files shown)


mastercoms:
Shouldn't the `pulse()` method check `!isObstructed()` to see if the painting can't survive at its current location? Or are the docs wrong for it? `@return true if the painting can survive, false otherwise`

kaenganxt:
The method returns true if the painting should drop, so the docs are wrong.

mastercoms:
And vanilla drops paintings if any block behind them is air? How do secret painting doors exist?

kaenganxt:
Yes, vanilla drops paintings if there's at least one block of air (I justed tested it). For secret doors, there's a ladder or a sign behind the painting.

mastercoms:
Ah ok, thank you for your contribution.

PR #598, opened 2017-12-09
Implement more methods in GlowPlayer.Spigot

hibo98:
This pull request:
- implements most of the unimplemented methods in `GlowPlayer.Spigot`
- moves the `isDead()` check to the normal `respawn()` method
I thought about a case where you don't need a check for dead when respawning a player but I can't think of a case where you don't want that
- doesn't implement `getPing()`, because I didn't find where it is stored, if already


aramperes (approved):

mastercoms:

PR #603, opened 2017-12-12
Material based ItemTable

mastercoms:
This converts the `ItemTable` and its usage to be based on `Material` types, rather than item IDs. This partially future proofs Glowstone and also prevents autoboxing `Integer`s in the type map.

To support custom materials, a new mapping from namespaced keys to `ItemType`s has been added.


aramperes:
I have a feeling Spigot will deprecate the Material enum for a more mutable alternative. Although I think this is better (using Material instead of IDs), I don't know how future-proof this is.

aramperes (approved):
I think this is good for the moment.

PR #604, opened 2017-12-12
Clean up and fix several issues regarding player profiles

kaenganxt:
This fixes several problems and inconsistencies with player profiles, offline players and player data fetcher.

#### Removed name constructor in `GlowOfflinePlayer`
Calling `PlayerProfile.getProfile(name)` may return null, which initializes the `GlowOfflinePlayer` with a null profile. This is not allowed as many methods would throw a `NullPointerException` (including the loadData method called in constructor).
To get an `OfflinePlayer` by name `GlowServer.getOfflinePlayer(name)` should be used instead, as this method assigns the correct offline player uuid if no online uuid could be retrieved.

#### Simplified `GlowOfflinePlayer.getPlayer()`
UUID can never be null and getting player by name would cause a loop as `getName()` calls `getPlayer()`.

#### Simplified `GlowServer.getOfflinePlayer(name)` and `GlowServer.getOfflinePlayer(uuid)`
Calling getPlayer() two times is unnecessary

#### `PlayerDataFetcher.getProfile(uuid)` now returns a `PlayerProfile` with a null name in case of an error
Returning null in case of an error can cause `NullPointerExceptions` at several places (e.g. the profile in `GlowOfflinePlayer` could again be null)

#### Removed name lookup in `PlayerProfile` constructor
This could cause a loop in case of a lookup error: `ProfileCache.getProfile()` -> `PlayerDataFetcher.getProfile()` -> creates `PlayerProfile` with null name -> Lookup process starts again

#### Changes in `PlayerProfile.getProfile(name)`
At first, the name is looked up with `getPlayerExact()` instead of `getPlayer()` for obvious reasons.
Then, a uuid is only requested from Mojang when the server is in online mode.

#### Removed `PlayerDataService.lookupUUID()`
This method was not used anywhere and introduced another uuid cache like the one in `ProfileCache`. Further it performed the same actions as `GlowServer.getOfflinePlayer(name)` if the uuid was not found, but iterated over all OfflinePlayers before that, which is very inefficient.

#### Removed `UuidListFiles.getUUIDs()`
The use of this function can and should be replaced by `getProfiles()` as that method fills the name with a fallback version from the corresponding list file, if the uuid lookup failed (e.g. it allows to autocomplete in deop command if the uuid of an opped player does not exist)

I did my best to fix all uses of these functions wherever the change would break something, but please review the changes as I might have missed some cases.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -29,7 +29,11 @@ import java.util.Map;
 import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -1588,6 +1592,15 @@ public final class GlowServer implements Server {
 
     @Override
     public OfflinePlayer[] getOfflinePlayers() {
+        return getOfflinePlayersAsync().join();
+    }
+
+    /**
+     * Gets every player that has ever played on this server.
+     *
+     * @return An OfflinePlayer[] future.
+     */
+    public CompletableFuture<OfflinePlayer[]> getOfflinePlayersAsync() {
         Set<OfflinePlayer> result = new HashSet<>();
         Set<UUID> uuids = new HashSet<>();
 
@@ -1599,13 +1612,11 @@ public final class GlowServer implements Server {
             }
         }
 
-        // add all offline players that aren't already online
-        getPlayerDataService().getOfflinePlayers().stream().filter(offline -> !uuids.contains(offline.getUniqueId())).forEach(offline -> {
-            result.add(offline);
-            uuids.add(offline.getUniqueId());
-        });
-
-        return result.toArray(new OfflinePlayer[result.size()]);
+        return getPlayerDataService().getOfflinePlayers().thenAcceptAsync(offlinePlayers -> offlinePlayers.stream()
+                .filter(offline -> !uuids.contains(offline.getUniqueId())).forEach(offline -> {
+                    result.add(offline);
+                    uuids.add(offline.getUniqueId());
+                })).thenApply((v) -> result.toArray(new OfflinePlayer[result.size()]));
     }
 
     @Override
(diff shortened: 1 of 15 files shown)


mastercoms:
Could we have the blocking methods return futures?

kaenganxt:
I never worked with futures. How would they be implemented?
And how would we present them in API methods?

kamcio96:
Unfortunately you can't use Future with bukkit API. Will review and test tomorrow ;D

May I should back to Glowstone development :thinking:

mastercoms:
We would use a blocking `Future#get` call to provide values for API methods.

kaenganxt:
So the ProfileCache.getUUID(name) would look like this?:
```java
public static CompletableFuture<UUID> getUUID(String playerName) {
    if (uuidCache.containsKey(playerName)) {
        return CompletableFuture.completedFuture(uuidCache.get(playerName));
    }
    CompletableFuture<UUID> uuidFuture = CompletableFuture
            .supplyAsync(() -> PlayerDataFetcher.getUUID(playerName));
    uuidFuture.thenAcceptAsync(uid -> uuidCache.put(playerName, uid));
    return uuidFuture;
}
```

I will look more into this tomorrow.

kaenganxt:
I now changed the methods to return Futures. I decided to start this on the ProfileCache layer and to keep the PlayerDataFetcher methods blocking.
I further added non-blocking methods to GlowServer for `getOfflinePlayer(name)`, `getOfflinePlayer(uuid)` and `getOfflinePlayers()`.

The default api methods for `getOfflinePlayer(...)` currently have a timeout of five seconds. I don't know if we need such timeouts, but a similar mechanism was already used before in the old uuid lookup. I was also wondering if `getOfflinePlayers()` would need a timeout, because the time it takes to complete depends on the amount of offline players.

Everything is still untested, but please give feedback on what you think about the changes.

All Glowstone methods which have used the changed methods currently use blocking calls. Some cases could be changed to a real asynchronous behaviour at a later date (e.g. the whitelist command).

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1675,25 +1686,66 @@ public OfflinePlayer getOfflinePlayer(PlayerProfile profile) {
...
-        PlayerProfile profile = PlayerProfile.getProfile(name);
-        if (profile == null) {
-            return getOfflinePlayer(new PlayerProfile(name, UUID.nameUUIDFromBytes(("OfflinePlayer:" + name).getBytes())));
-        } else {
-            return getOfflinePlayer(profile);
+        try {
+            // probably blocking, limited to five seconds
+            return getOfflinePlayerAsync(name).get(5, TimeUnit.SECONDS);
+        } catch (InterruptedException | ExecutionException ex) {
+            Logger.getLogger(GlowServer.class.getName()).log(Level.SEVERE, "UUID lookup interrupted: ", ex);
```
aramperes: You can use `GlowServer.logger.log(Level.X, "message", exception)`

kamcio96 (approved):

kaenganxt:
So, what do you think about the timeouts in the GlowServer methods?

aramperes:
I think the timeout is fine, although it would be nice to make it configurable.

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1701,7 +1705,12 @@ public OfflinePlayer getOfflinePlayer(String name) {
     @Override
     public OfflinePlayer getOfflinePlayer(UUID uuid) {
         try {
-            return getOfflinePlayerAsync(uuid).get(5, TimeUnit.SECONDS); // todo: should we timeout? If yes, how long?
+            // probably blocking, timeout depending on config setting
+            if (config.getInt(Key.PROFILE_LOOKUP_TIMEOUT) <= 0) {
```
aramperes: Add a method in GlowServer to get this value (e.g. `getProfileLookupTimeout()`), with documentation specifying it is in seconds.

aramperes (approved):

mastercoms (approved):

PR #607, opened 2017-12-17
Implement GlowMapRenderer.render()

Pr0methean:
This implementation is basic -- it only considers the material value of the highest block at each pixel -- but it implements the limited vision range and is configurable in YAML. Someone may still need to implement persistence for the GlowMapCanvas.


src/main/resources/builtin/materialValues.yml
```diff
@@ -4,21 +4,28 @@ default:
   lightOpacity: 255
   flameResistance: -1
   fireResistance: -1
+  baseMapColor: 44
 values:
+  # TODO: Define baseMapColor for all materials
```
mastercoms, line 9: The material values file is a WIP for all properties, so I feel like this comment is confusing.

mastercoms (commented):
Looks good on my first pass through!

Pr0methean:
Reopened directly from the feature branch as #608

PR #612, opened 2017-12-17
Add a slf4j back-end to pom.xml

### LINKED ISSUE #606: SLF4J Warning on startup
Glowstone build: 89f3731 / `1407`

I'm getting this warning during server startup. This happens right after the configs are created and before binding the server.

```
[WARNING] SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
[WARNING] SLF4J: Defaulting to no-operation (NOP) logger implementation
[WARNING] SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Pr0methean:
Fixes #606
Used only by a library, not by Glowstone itself.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <version>1.7.25</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
       <groupId>net.glowstone</groupId>
       <artifactId>glowkit</artifactId>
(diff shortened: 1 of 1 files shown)


aramperes:
Since Glowstone doesn't use it, I think it should be added to the library manager and downloaded at runtime instead. Thoughts @mastercoms?

Pr0methean:
That will only fix it if the library that uses slf4j hasn't already invoked LoggerFactory.

mastercoms:
This is fine for now. I'll leave the issue open until we get a new library management system that does a proper fix to it.

mastercoms (approved):

PR #613, opened 2017-12-17
Fix a ConcurrentModificationException

### LINKED ISSUE #609: Concurrency issue when player spawns in world
Glowstone build: `1407`

Sometimes, there will be a concurrency exception when the player joins the world as the player itself is being pulsed:

```
        at java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)
        at java.util.HashMap$KeyIterator.next(HashMap.java:1461)
        at net.glowstone.entity.GlowPlayer.pulse(GlowPlayer.java:810)
        [ ... ]
        at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:418)
        at net.glowstone.GlowWorld.pulsePlayers(GlowWorld.java:581)
        at net.glowstone.GlowWorld.pulse(GlowWorld.java:461)
```

This doesn't happen all the time.

Pr0methean:
Fixes #609

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -18,7 +18,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -806,17 +806,23 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
 
         // update or remove entities
         List<Integer> destroyIds = new LinkedList<>();
-        for (Iterator<GlowEntity> it = knownEntities.iterator(); it.hasNext(); ) {
-            GlowEntity entity = it.next();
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -806,17 +805,23 @@ public void pulse() {
...
-                it.remove();
+                destroyEntities.add(entity);
             } else {
                 entity.createUpdateMessage(session).forEach(session::send);
             }
         }
         if (!destroyIds.isEmpty()) {
             session.send(new DestroyEntitiesMessage(destroyIds));
+            for (Entity entity : destroyEntities) {
+                knownEntitiesCopy.remove(entity);
```
kamcio96: `removeAll` ?

smartboyathome:
Do note that this won't completely eliminate the possibility of a ConcurrentModificationException, it'll just reduce the window during which it can happen.

aramperes:
I'm worried about this solution because while it does fix the exception for the majority of cases, it doesn't address the base problem

- The creation of the new list can still cause a concurrency exception
- This is creating an extra List instance every tick, per player
- There may be a risk of broken functionality that isn't addressed here. This exception can happen when a player switches from one world from another. Entity IDs are only unique per-world, so it may happen that an entity is wrongly destroyed on the client-side in the new world.

Pr0methean:
Does shifting to a new world trigger a new `pulse` call and restore wrongly-deleted entities? If so, a solution may be to have `world` and `knownEntities` guarded by the same lock.

Issue #505, opened 2017-06-24
[#396] Entities: Projectiles [$50 awarded]


gdude2002:
[![Bountysource bounty](https://api.bountysource.com/badge/issue?issue_id=46548051)](https://www.bountysource.com/issues/46548051-396-entities-projectiles)

This issue is part of [this story](https://github.com/GlowstoneMC/Glowstone/issues/396), *Completing entities in Glowstone* (#396). To make the work easier and bounties fairer, we've split it into multiple issues.

Anyone can participate in the completion of this issue, which will be a major turning point in the development and the possible use-cases for Glowstone as a Minecraft server alternative.

---

**Before getting started!** If you are unfamiliar with Glowstone, please remember that Glowstone is a complete rewrite of the Minecraft vanilla server, aimed to be a better alternative to other software like Spigot. You are by no means allowed to push code to this repository based on Mojang's own platform (whether it be Forge, MCP, decompiled binaries, etc.) Doing so will cause in the rejection of your Pull Request. You are allowed to use other resources like the numerous wikis that exist on the Internet (Minecraft Wiki, wiki.vg, etc.) as well as trial-and-error.

---

Projectiles
=======

Projectiles are not implemented, but Momo has worked on them before in a separate [branch](https://github.com/GlowstoneMC/Glowstone/tree/projectiles). The implementation can be based on that work, but the physics system has been reworked a lot (so it could be severely outdated). The following projectiles need to be implemented:

* Arrow (including shooting them with a bow, and the functionality behind it -- e.g. pulling back a bow), and Potion arrows
* Eggs (including throwing them and spawning chickens randomly)
* Ender Pearl (including throwing them and teleporting the player)
* Fireball
* Snowball (including throwing them and hitting other entities)
* Splash potions (and possibly implementing effects at the same time!)
* Experience bottles (including implementing the Experience Orbs at the same time!)
* Wither Skull (including targeting, related to AI)

PR #614, opened 2017-12-17
Projectiles

Pr0methean:
Also implements arrow consumption for ItemBow, updates the entity IDs for projectiles, and stubs out and registers missing projectile classes.


Pr0methean:
This is a partial fix of #505

src/main/java/net/glowstone/entity/EntityRegistry.java
```diff
@@ -68,74 +68,18 @@
...
-import org.bukkit.entity.Vindicator;
-import org.bukkit.entity.Weather;
-import org.bukkit.entity.Witch;
-import org.bukkit.entity.Wither;
-import org.bukkit.entity.WitherSkeleton;
-import org.bukkit.entity.Wolf;
-import org.bukkit.entity.Zombie;
-import org.bukkit.entity.ZombieHorse;
-import org.bukkit.entity.ZombieVillager;
+import org.bukkit.entity.*;
```
hibo98: Please remove the wildcard import
See [Code Style](https://github.com/GlowstoneMC/Glowstone/wiki/Code-Style)

mastercoms:
FYI, we'll be waiting to merge this after January 1st, due to the current soft feature freeze.

Pr0methean:
Thanks for the warning. I'll check again for merge conflicts shortly after Christmas.

Pr0methean:
Should an AreaEffectCloud created by a lingering potion always be the same color as the potion?

Pr0methean:
This is ready for review.

src/main/java/net/glowstone/block/entity/state/GlowDispenser.java
```diff
@@ -92,12 +98,24 @@ public static void register() {
                     InventoryUtil.copyPotionDataToTippedArrow(tippedArrow, itemStack);
                     return tippedArrow;
                 })));
-        registry.putBehavior(Material.FIREWORK_CHARGE, new ProjectileDispenseBehavior(location -> {
```
aramperes: Was this removal intentional?

src/main/java/net/glowstone/block/blocktype/BlockTnt.java
```diff
@@ -19,7 +20,7 @@
      * @param tntBlock The block to ignite.
      * @param ignitedByExplosion True if another explosion caused this ignition.
      */
-    public static void igniteBlock(GlowBlock tntBlock, boolean ignitedByExplosion) {
+    public static void igniteBlock(Block tntBlock, boolean ignitedByExplosion) {
```
aramperes, line 23: I think this is due to a merge mistake -- is there a reason for this change?

src/main/java/net/glowstone/block/itemtype/ItemBow.java
```diff
@@ -0,0 +1,101 @@
...
+            if (ThreadLocalRandom.current().nextDouble() < 0.2) {
+                launchedArrow.setCritical(true);
+            }
+        }
+        InventoryUtil.damageItem(player, item);
+        player.setUsageItem(null);
+        player.setUsageTime(0);
+        currentArrow = null;
+    }
+
```
aramperes: Remove whitespace.

src/main/java/net/glowstone/block/entity/state/GlowDispenser.java
Pr0methean: Yes: should have been FIREBALL.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -207,6 +208,11 @@
      * Whether gravity applies to the entity.
      */
     private boolean gravity = true;
+    /**
+     * Whether friction applies to the entity.
+     */
+    @Setter
+    private boolean friction = true;
```
aramperes, line 236: Why only lombokify the Setter?

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,133 @@
...
+    @Setter
+    private int knockbackStrength;
+    /**
+     * How long an entity burns after being shot with a burning arrow.
+     */
+    private static final int TARGET_BURN_TICKS = 100;
+
+    public GlowArrow(Location location) {
+        super(location);
+        setGravityAccel(new Vector(0,-0.05,0));
```
aramperes: Spaces after commas (`0, -0.05, 0`)

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public void setOnGround(boolean onGround) {
+        super.setOnGround(onGround);
+        if (onGround) {
+            setVelocity(new Vector(0, 0, 0));
+        }
+    }
+
```
aramperes: Remove whitespace.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public PickupRule getPickupRule() {
+        return null;
+    }
+
+    @Override
+    public void setPickupRule(PickupRule pickupRule) {
+
```
aramperes: Can you implement this? It shouldn't be too hard to just add a field with lombok.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,133 @@
...
+    public boolean isCritical() {
+        return metadata.getBit(MetadataIndex.ARROW_CRITICAL, 0x1);
+    }
+
+    @Override
+    public void setCritical(boolean critical) {
+        metadata.setBit(MetadataIndex.ARROW_CRITICAL, 0x1, critical);
+    }
+
+    @Override public PickupStatus getPickupStatus() {
```
aramperes: Annotations on different line

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,133 @@
...
+        PickupStatus customPickupStatus = this.customPickupStatus;
+        return customPickupStatus != null ? customPickupStatus :
+                getShooter() instanceof Monster ? PickupStatus.DISALLOWED :
+                PickupStatus.ALLOWED;
+    }
+
+    @Override public void setPickupStatus(PickupStatus pickupStatus) {
```
aramperes: Annotations on different line

src/main/java/net/glowstone/entity/projectile/GlowLingeringPotion.java
```diff
@@ -0,0 +1,55 @@
...
+                    effect = new PotionEffect(effect.getType(), effect.getDuration() >> 2,
+                            effect.getAmplifier());
+                }
+                // TODO: else effect is 1/2 the usual
+                cloud.addCustomEffect(effect, true);
+            }
+        }
+        remove();
+    }
+
```
aramperes: Remove whitespace.

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
```diff
@@ -0,0 +1,30 @@
...
+class AreaEffectCloudStore extends EntityStore<GlowAreaEffectCloud> {
+
+    public AreaEffectCloudStore() {
+        super(GlowAreaEffectCloud.class, EntityType.AREA_EFFECT_CLOUD);
+    }
+
+    @Override
+    public void load(GlowAreaEffectCloud entity, CompoundTag tag) {
+        super.load(entity, tag);
+        // TODO
```
aramperes: Can you add what NBT fields need to be added in the TODO?

src/test/java/net/glowstone/block/itemtype/ItemBowTest.java
```diff
@@ -0,0 +1,67 @@
...
+        // Finish shooting
+        bow.endUse(player, bowItemStack);
+        verify(player, times(1)).launchProjectile(Arrow.class);
+        verify(player, times(1)).setUsageItem(null);
+        verify(player, times(1)).setUsageTime(0);
+        assertEquals(1, bowItemStack.getDurability()); // shooting should damage the bow
+        // FIXME: assertEquals(1, arrows.getAmount()); // one arrow should be consumed
+    }
+
+}
```
aramperes: Add blank line at end (LF)

src/test/java/net/glowstone/net/PlayProtocolTest.java
```diff
@@ -84,6 +82,9 @@
 import org.bukkit.util.BlockVector;
 import org.bukkit.util.Vector;
 
+import java.util.Arrays;
+import java.util.UUID;
```
aramperes: Probably due to a merge mistake

aramperes (changes requested):

src/main/java/net/glowstone/block/entity/state/GlowDispenser.java
aramperes: Do fireworks no longer get let off by dispensers? Or is `firework_charge` something else?

Pr0methean: I'm pretty sure FIREWORK_CHARGE is a firework *star*. The dispense behavior for firework *rockets* will not be a ProjectileDispenseBehavior either, because GlowFirework doesn't implement Projectile.

src/main/java/net/glowstone/block/blocktype/BlockTnt.java
Pr0methean: The block doesn't actually have to be a GlowBlock for the method body to work, and widening the types will help simplify calling code and maximize plugin compatibility.

src/main/java/net/glowstone/block/itemtype/ItemBow.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final double MAX_SPEED = 53;
+    private ItemStack currentArrow;
+
+    @Override
+    public void startUse(GlowPlayer player, ItemStack item) {
+        findArrow: for (ItemStack itemStack : player.getInventory().getContents()) {
+            switch (itemStack.getType()) {
+                case ARROW:
+                    if (currentArrow == null) {
+                        currentArrow = itemStack;
```
aramperes: I don't think this is how it should be implemented. The arrow lookup should be done in the `endUse`: the startUse should just check if there *is* an arrow somewhere.

Suppose that the player has two stacks of arrows in their inventory. This function will determine that the first stack is to be used, and will be stored in `currentArrow`. If the stack is dropped or removed during the bow pulling (by a plugin or whatever), the `endUse` function won't be able to find the stack with this current system. This search should be done when by the `endUse`.

The `endUse` function should find the arrow and get the slot index for the stack. Then, it will summon the arrow and remove one/delete and update the slot (`inventory.setItem(slot, arrowStack)`)

aramperes (dismissed):

src/main/java/net/glowstone/block/entity/state/GlowDispenser.java
aramperes: Ah okay, I thought the firework entities were considered projectiles.

src/main/java/net/glowstone/entity/GlowEntity.java
Pr0methean: Because a lombokified Getter would be called "isFriction" where we'd want "hasFriction". And the getter isn't used anyway, so for now there's none.

src/main/java/net/glowstone/block/entity/state/GlowDispenser.java
Pr0methean: Adding a TODO in the next commit.

src/main/java/net/glowstone/block/itemtype/ItemBow.java
Pr0methean: Done in the next commit.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Done in the next commit.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Done in the next commit.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Done in the next commit.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Done in the next commit.

src/main/java/net/glowstone/entity/GlowEntity.java
aramperes: Ah okay :)

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Switching to interface defaults instead (stub generation must have been using a different version of Glowkit).

src/main/java/net/glowstone/entity/projectile/GlowLingeringPotion.java
Pr0methean: Done in the next commit.

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
Pr0methean: Done in the next commit.

src/test/java/net/glowstone/block/itemtype/ItemBowTest.java
Pr0methean: Done in the next commit.

src/test/java/net/glowstone/net/PlayProtocolTest.java
Pr0methean: Reverted in the next commit.

src/main/java/net/glowstone/block/itemtype/ItemBow.java
Pr0methean: Done in cb38717fadd2b1bdbaa3f0108b47cff6680c9c9d

src/main/java/net/glowstone/block/itemtype/ItemBow.java
```diff
@@ -0,0 +1,115 @@
...
+            player.setUsageItem(item);
+            player.setUsageTime(TICKS_TO_FULLY_CHARGE);
+        }
+    }
+
+    @Override
+    public void endUse(GlowPlayer player, ItemStack item) {
+        ItemStack currentArrow = null;
+        // Check arrows again, since plugins can change the inventory while a bow is drawn
+        findArrow: for (ItemStack itemStack : player.getInventory().getContents()) {
```
aramperes: Merge both for-loops into a method, returning an `Optional<ItemStack>`.

aramperes (dismissed):

src/main/java/net/glowstone/block/itemtype/ItemBow.java
Pr0methean: Done in 4a56ae374ea310c61123396fc796d71a0692da34

src/main/java/net/glowstone/block/itemtype/ItemProjectile.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public abstract class ItemProjectile extends ItemType {
+
+    protected final EntityType entityType;
+
+    public ItemProjectile(EntityType entityType) {
+        this.entityType = entityType;
+    }
+
+    public abstract Projectile use(GlowPlayer player, ItemStack holding);
```
aramperes: Needs documentation

src/main/java/net/glowstone/util/InventoryUtil.java
```diff
@@ -67,4 +76,46 @@ public static int getRandomSlot(Random random, Inventory inventory, boolean igno
...
         return notEmpty.get(random.nextInt(notEmpty.size()));
     }
+
+    /**
+     * Inflicts damage to an item. Unbreaking enchantment is applied if present.
+     *
+     * @param player The player holding the item, or null if held by a dispenser
+     * @param holding The item
+     */
+    public static void damageItem(GlowPlayer player, ItemStack holding) {
```
aramperes: What I would do, is make a method with just the `ItemStack`, that returns an `ItemStack`. The ItemStack doesn't get updated in the player's inventory until you use `inventory.setItem(slot, stack)`; and since the slot information isn't inside of the ItemStack object, this current implementation won't work.

Also, in that case, return `InventoryUtil.createEmptyStack()` if the item is to be removed (max damage)

aramperes (changes requested):

aramperes:
Also, please consider stubbing/implementing the `WitherSkull` entity, which extends `Fireball`

src/main/java/net/glowstone/block/itemtype/ItemProjectile.java
Pr0methean: Documented in a314f6a8d42113df74766f2043f9dbc9bb23a91d

src/main/java/net/glowstone/util/InventoryUtil.java
Pr0methean: Done in 87f6098857730d7a152df87c2d8ed49c15535b69

Pr0methean:
Added GlowWitherSkull in f394fab

src/main/java/net/glowstone/entity/projectile/GlowWitherSkull.java
```diff
@@ -0,0 +1,22 @@
...
+    @Getter
+    @Setter
+    private boolean charged;
+
+    @Override
+    protected int getObjectId() {
+        return NETWORK_ID;
+    }
+
+    public GlowWitherSkull(Location location) {
```
aramperes: Put the constructor above other methods.

aramperes (dismissed):

src/main/java/net/glowstone/entity/projectile/GlowWitherSkull.java
Pr0methean: Done in f97a894

src/main/java/net/glowstone/block/itemtype/ItemBow.java
```diff
@@ -0,0 +1,116 @@
...
+import org.bukkit.inventory.meta.PotionMeta;
+
+public class ItemBow extends ItemTimedUsage {
+    private static final long TICKS_TO_FULLY_CHARGE = 20;
+    private static final double MAX_BASE_DAMAGE = 9;
+    private static final double MAX_SPEED = 53;
+
+    @Override
+    public void startUse(GlowPlayer player, ItemStack item) {
+        if (findArrow(player).isPresent()) {
```
aramperes: In vanilla, you can shoot a bow even if you don't have arrows in your inventory. By default, it will shoot plain arrows, and if there are tipped arrows, it will use those instead (without consumption).

Here, the startUse would do nothing if the player is in creative and has no arrows, thus not shooting anything in the end.

Pr0methean: Fixed: now shoots a plain arrow if you're in Creative and have no arrows.

src/main/java/net/glowstone/entity/projectile/GlowLingeringPotion.java
```diff
@@ -0,0 +1,54 @@
...
+        ItemMeta meta = getItem().getItemMeta();
+        if (meta instanceof PotionMeta) {
+            PotionMeta potionMeta = (PotionMeta) meta;
+            cloud.setColor(potionMeta.getColor());
+            cloud.setBasePotionData(potionMeta.getBasePotionData());
+            for (PotionEffect effect : getEffects()) {
+                // Cloud effects have only 1/4 the usual duration
+                PotionEffectType type = effect.getType();
+                if (!type.isInstant()) {
+                    effect = new PotionEffect(effect.getType(), effect.getDuration() >> 2,
```
Postremus: Could we use a simple "/ 4" here instead?
Reading bit shifts is not not natural to all of us, and requires a few extra seconds to understand.

Pr0methean: Done in bd87257c067ff07dde11c181fff35b4a7ce36355

src/main/java/net/glowstone/entity/projectile/GlowProjectile.java
```diff
@@ -0,0 +1,76 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import net.glowstone.entity.GlowEntity;
+import net.glowstone.net.message.play.entity.EntityMetadataMessage;
+import net.glowstone.net.message.play.entity.EntityVelocityMessage;
+import net.glowstone.net.message.play.entity.SpawnObjectMessage;
+import net.glowstone.util.Position;
+import org.bukkit.Location;
+import org.bukkit.block.Block;
+import org.bukkit.block.BlockFace;
```
mastercoms: Unused import.

src/main/java/net/glowstone/PotionDataHolder.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * <p>Plugins should check that hasCustomEffects() returns true before calling
+     * this method.
+     *
+     * @return the immutable list of custom potion effects
+     */
+    List<PotionEffect> getCustomEffects();
+
+    /**
+     * Adds a custom potion effect to this arrow.
```
mastercoms: Why do these docs refer to arrows only?

src/main/java/net/glowstone/block/ItemTable.java
```diff
@@ -391,7 +395,6 @@ private void registerBuiltins() {
                 .YELLOW_GLAZED_TERRACOTTA));
         reg(Material.CHORUS_FLOWER, new BlockChorusFlower());
         reg(Material.CHORUS_PLANT, new BlockChorusPlant());
-
```
mastercoms: Don't remove the separation here between blocks and items.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -957,28 +969,50 @@ protected void pulsePhysics() {
...
-                    velocity.setX(velocity.getX() * 0.91);
-                    velocity.setZ(velocity.getZ() * 0.91);
+                    velocity.setY(airDrag * (velocity.getY() + getGravityAccel().getY()));
+                    if (isOnGround()) {
+                        velocity.setX(velocity.getX() * slipMultiplier);
+                        velocity.setZ(velocity.getZ() * slipMultiplier);
+                    } else {
+                        horizontalAirDrag = 0.91;
+                        velocity.setX(velocity.getX() * horizontalAirDrag);
+                        velocity.setZ(velocity.getZ() * 0.91);
```
mastercoms: Use `horizontalAirDrag` here as well.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -957,28 +969,50 @@ protected void pulsePhysics() {
...
+                    }
+                }
+            } else if (hasGravity() && !isOnGround()) {
+                switch (location.getBlock().getType()) {
+                    case WATER:
+                    case LAVA:
+                        velocity.setY(velocity.getY() + getGravityAccel().getY() / 4d);
+                        break;
+                    default:
+                        velocity.setY(velocity.getY() + getGravityAccel().getY() / 4d);
```
mastercoms, line 1005: Gravity should not be divided by 4 here.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -663,11 +668,60 @@ public boolean isWithinSolidBlock() {
...
 
     @Override
-    public <T extends Projectile> T launchProjectile(Class<? extends T> projectile,
-        Vector velocity) {
-        T entity = world.spawn(getEyeLocation(), projectile);
-        entity.setVelocity(velocity);
-        return entity;
+    public <T extends Projectile> T launchProjectile(Class<? extends T> clazz, Vector vector) {
+        float offset = 0.0F;
+        float velocity = 1.5F;
```
mastercoms: Rename velocity to speed here, and any where else velocity is being defined as a magnitude. Then rename vector to velocity.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -663,11 +668,60 @@ public boolean isWithinSolidBlock() {
...
+        if (vector == null) {
+            vector = getVelocity();
+        }
+
+        T projectile = throwProjectile(clazz, getEyeLocation().clone(), vector, offset, velocity);
+        projectile.setShooter(this);
+        return projectile;
+    }
+
+    protected <T extends Projectile> T throwProjectile(Class<? extends T> type, Location location,
```
mastercoms: Why does the method name change from launchProjectile to throwProjectile?

Likewise, why does clazz get changed to type?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -663,11 +668,60 @@ public boolean isWithinSolidBlock() {
...
+            Vector originalVector, float offset, float velocity) {
+        double k = Math.toRadians(-1);
```
mastercoms: Could you rename these k variables to something more descriptive?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -663,11 +668,60 @@ public boolean isWithinSolidBlock() {
...
+        double x = cos(k * location.getPitch()) * sin(k * location.getYaw());
```
mastercoms: Could you explain your vector math here and in the next method?

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    public void collide(Block block) {
+        setFireTicks(0); // Arrows stop burning when they land, and ignite only TNT
+        if (block.getType() == Material.TNT) {
+            BlockTnt.igniteBlock(block, false);
+        }
+        setOnGround(true);
+        // Don't despawn
```
mastercoms: Arrows should be despawned eventually, perhaps with a configurable value like in Spigot/Paper.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
```diff
@@ -0,0 +1,130 @@
...
+            entity.setFireTicks(TARGET_BURN_TICKS);
+        }
+        // TODO: knockback
+        entity.setArrowsStuck(entity.getArrowsStuck() + 1);
+        remove();
+    }
+
+    @Override
+    protected int getObjectId() {
+        return 60;
```
mastercoms: Get rid of this object ID stuff and define it here: https://github.com/GlowstoneMC/Glowstone/blob/97fc5a6017b566af4e86b30cf6def6ebc4f0b9d4/src/main/java/net/glowstone/net/message/play/entity/SpawnObjectMessage.java#L14

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
```diff
@@ -0,0 +1,83 @@
+package net.glowstone.entity.projectile;
+
+import static net.glowstone.net.message.play.entity.SpawnObjectMessage.THROWN_ENDERPEARL;
```
mastercoms: I would prefer using `SpawnObjectMessage.THROWN_ENDERPEARL` in line.

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
```diff
@@ -0,0 +1,83 @@
...
+import org.bukkit.entity.EnderPearl;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.event.entity.EntityDamageEvent;
+import org.bukkit.event.player.PlayerTeleportEvent;
+import org.bukkit.projectiles.ProjectileSource;
+import org.bukkit.util.Vector;
+
+public class GlowEnderPearl extends GlowProjectile implements EnderPearl {
+    private static final double ENDER_PEARL_DAMAGE = 5.0;
```
mastercoms: Add a line between the class declaration and this field.

src/main/java/net/glowstone/entity/meta/MetadataIndex.java
```diff
@@ -80,16 +82,15 @@
...
-    //TODO 1.9-1.10 - Support this
-    //AREAEFFECTCLOUD_RADIUS(6, FLOAT, Entity.class),
-    //AREAEFFECTCLOUD_COLOR(7, INT, Entity.class),
-    //AREAEFFECTCLOUD_UNKNOWN(8, BOOLEAN, Entity.class),
-    //AREAEFFECTCLOUD_PARTICLEID(9, INT, Entity.class),
-    //AREAEFFECTCLOUD_PARTICLE_PARAM1(10, INT, Entity.class),
-    //AREAEFFECTCLOUD_PARTICLE_PARAM2(11, INT, Entity.class),
+    AREAEFFECTCLOUD_RADIUS(6, FLOAT, AreaEffectCloud.class),
+    AREAEFFECTCLOUD_COLOR(7, INT, AreaEffectCloud.class),
+    AREAEFFECTCLOUD_UNKNOWN(8, BOOLEAN, AreaEffectCloud.class),
```
mastercoms: Rename this to `AREAEFFECTCLOUD_POINT`.

src/main/java/net/glowstone/net/handler/play/player/UseItemHandler.java
```diff
@@ -90,6 +92,16 @@ static void handleRightClickAir(GlowPlayer player, ItemStack holding, EquipmentS
             if (type != null) {
                 if (type.getContext().isAirApplicable()) {
                     type.rightClickAir(player, holding);
+                } else {
+                    if ((holding.getType() == Material.WATER_BUCKET
```
mastercoms: Seems like this was incorrectly reverted from the ender pearl PR.

src/main/java/net/glowstone/scheduler/GlowScheduler.java
```diff
@@ -74,17 +76,33 @@
...
+     * The session registry used to pulse all players.
+     */
+    private final SessionRegistry sessionRegistry;
 
     /**
      * Creates a new task scheduler.
      *
      * @param server The server that will use this scheduler.
-     * @param worlds The {@link WorldScheduler} this scheduler will use for ticking the server's
+     * @param worlds The {@link WorldScheduler} this scheduler will use for ticking the servers
```
mastercoms: The previous doc was correct: server's worlds

src/test/java/net/glowstone/block/itemtype/ItemProjectileTest.java
```diff
@@ -0,0 +1,42 @@
...
+        Projectile projectile = Mockito.mock(projectileClass);
+        when(player.launchProjectile(projectileClass)).thenReturn((T) projectile);
+        item.use(player, itemStack);
+        checkProjectile((T) projectile);
+    }
+
+    protected void checkProjectile(T projectile) {
+        verify(player, times(1)).launchProjectile(projectileClass);
+    }
+}
```
mastercoms: Newline

mastercoms (dismissed):

src/main/java/net/glowstone/entity/GlowLivingEntity.java
aramperes: I believe I wrote this math originally in the projectiles branch. Here's my explanation (I learned trigonometry and vector math in another language, sorry if I'm not using the terms properly).

**Pitch/Yaw**
Pitch is defined as the angle of rotation on the horizontal axis, ranging from 90 degrees to -90 degrees.
Yaw is defined as the angle of rotation on the vertical axis ranging from 180 to -180.

**X component**
The first part consists of using the pitch (horizontal-axis rotation angle) to find out at which horizontal angle (downwards/upwards) the player is shooting the projectile. This angle helps as a “multiplier” for the X component, because shooting directly upwards or downwards should nullify the component. Because pitch is ranged negatively (negative is below the origin), we have to negate the conversion:
image
We must then take into consideration the vertical-axis rotation angle (yaw). A yaw of -90deg represents a full X component (+X), while 90deg represents the opposite (-X). 180 and 0 degrees represent a null X component. This conversion can be achieved by using:
image
Because the pitch acts as a multiplier to the vertical-axis rotation:
image

**Z component**
This conversion is similar to the X component: it uses the pitch as a multiplier too. What changes is the use of the yaw range, which can be calculated using sin instead of cosine.
image

**Y component**
This is similar to the operation used to find the multiplier to the X/Z components, but here we need the upwards and downwards directions to represent a full Y component (using sin) instead of nullifying it (using cosine).
image

src/main/java/net/glowstone/entity/GlowLivingEntity.java
aramperes: I made this variable to avoid using `Math.toRadians` for both pitch and yaw. In retrospect, there's a lot of stuff that can be simplified with this math (considering `cos(-x) = cos(x)`). I would definitely clean this up.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -663,11 +668,60 @@ public boolean isWithinSolidBlock() {
...
+        double y = sin(k * (location.getPitch() - offset));
+        double z = cos(location.getPitch() * k) * cos(location.getYaw() * k);
+        T projectile = throwProjectile(type, location, x, y, z, velocity);
+        projectile.getVelocity().add(originalVector);
+        return projectile;
+    }
+
+    private <T extends Projectile> T throwProjectile(Class<? extends T> clazz, Location location,
+            double x, double y, double z, float velocity) {
+        double k = Math.sqrt(x * x + y * y + z * z);
```
aramperes: `k` here is the magnitude of the unit represented by `x, y, z`, previously calculated. I believe the math below multiplies the vector by the given `velocity` magnitude (using component operations) and then adds these results to the base components, in order to give a certain offset. (this is in response to mastercoms)

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Added TODO for this, while I find out the best way to schedule the despawn event.

src/main/java/net/glowstone/entity/projectile/GlowArrow.java
Pr0methean: Done for the return values of all overrides of getObjectId()

src/main/java/net/glowstone/scheduler/GlowScheduler.java
Pr0methean: Fixed on both constructors.

src/main/java/net/glowstone/PotionDataHolder.java
Pr0methean: Fixed.

src/main/java/net/glowstone/block/ItemTable.java
Pr0methean: Reverted and added comments.

src/main/java/net/glowstone/entity/GlowEntity.java
Pr0methean: Done. Also removed spurious setter on line 988.

src/main/java/net/glowstone/entity/GlowEntity.java
Pr0methean: That's the existing behavior (see existing lines 964 and 967 below what's now 973).

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Done in all throwProjectile overloads.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Renamed to `launchProjectile` and `type` throughout.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Done: Calling this one `negDegreesPerRadian` and the other one `magnitude`.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Added a comment to explain this briefly.

src/main/java/net/glowstone/entity/meta/MetadataIndex.java
Pr0methean: Done.

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
Pr0methean: Done.

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
Pr0methean: Done.

src/main/java/net/glowstone/net/handler/play/player/UseItemHandler.java
Pr0methean: Fixed.

src/test/java/net/glowstone/block/itemtype/ItemProjectileTest.java
Pr0methean: Done.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Now called `magnitude` and checked for zero (so if x, y and z are all zero, we don't end up at coordinates of infinity or NaN).

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -654,20 +659,101 @@ public boolean isWithinSolidBlock() {
...
+     * @param vector the direction to shoot in
+     * @param pitchOffset degrees to subtract from the pitch angle while calculating the y component
+     *         of the initial direction
+     * @param speed the speed for the first flight tick
+     * @param <T> the projectile class
+     * @return the launched projectile
+     */
+    protected <T extends Projectile> T launchProjectile(Class<? extends T> type, Location location,
+            Vector originalVector, float pitchOffset, float velocity) {
+        double negDegreesPerRadian = Math.toRadians(-1);
```
aramperes: I think these calculations can be simplified like this:

```java
double pitchRadians = Math.toRadians(location.getPitch());
double yawRadians = Math.toRadians(location.getYaw());

double verticalMultiplier = cos(pitchRadians);
double x = verticalMultiplier * sin(-yawRadians);
double z = verticalMultiplier * cos(yawRadians);
double y = sin(-(Math.toRadians(location.getPitch() - pitchOffset));
```

src/main/java/net/glowstone/net/handler/play/player/UseItemHandler.java
```diff
@@ -90,6 +92,10 @@ static void handleRightClickAir(GlowPlayer player, ItemStack holding, EquipmentS
             if (type != null) {
                 if (type.getContext().isAirApplicable()) {
                     type.rightClickAir(player, holding);
+                } else if (type instanceof ItemTimedUsage) {
+                    ((ItemTimedUsage) type).startUse(player, holding);
```
kaenganxt: This call is already in the `rightClickAir` handler of `ItemTimedUsage`. You should make ItemProjectile similar to ItemTimedUsage (returning `Context.ANY` for `getContext()` and calling the `use` method in both `rightClickAir` and `rightClickBlock`). I haven't tested it, but I think currently you are not able to shoot a projectile when aiming at a block.

kaenganxt (dismissed):

src/main/java/net/glowstone/net/handler/play/player/UseItemHandler.java
Pr0methean: Done in 985f213b34b6752c1058d5d2f37d824a9eacaed3

src/main/java/net/glowstone/EventFactory.java
```diff
@@ -49,15 +52,16 @@ private EventFactory() {
      * @return the called event
      */
     public static <T extends Event> T callEvent(T event) {
-        GlowServer server = (GlowServer) Bukkit.getServer();
+        Server server = Bukkit.getServer();
```
aramperes, line 56: Why was this change necessary? `Bukkit.getServer()` is guaranteed to be a `GlowServer` instance. The only difference I see here are the additional imports.

Or was this done because it broke tests somehow?

Pr0methean: Yes it did (ServerShim doesn't extend GlowServer).

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
```diff
@@ -52,8 +52,8 @@ public void load(GlowAreaEffectCloud entity, CompoundTag tag) {
             try {
                 entity.setParticle(Particle.valueOf(particle));
             } catch (IllegalArgumentException e) {
-                Bukkit.getServer().getLogger().warning("Ignoring invalid particle type "
-                        + particle);
+                Bukkit.getServer().getLogger().warning(String.format(
```
mastercoms: Could you use `.info(Level.WARNING, ...)`

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
```diff
@@ -0,0 +1,111 @@
...
+        handleFloatIfPresent(tag, RADIUS, entity::setRadius);
+        handleFloatIfPresent(tag, RADIUS_ON_USE, entity::setRadiusOnUse);
+        handleFloatIfPresent(tag, RADIUS_PER_TICK, entity::setRadiusPerTick);
+        if (tag.isString(PARTICLE)) {
+            final String particle = tag.getString(PARTICLE);
+
+            try {
+                entity.setParticle(Particle.valueOf(particle));
+            } catch (IllegalArgumentException e) {
+                Bukkit.getServer().getLogger().warning(String.format(
```
mastercoms: Could you use `log(Level.WARNING, ...)`

src/main/java/net/glowstone/io/entity/FireballStore.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Arrays;
+import java.util.List;
+import net.glowstone.entity.projectile.GlowFireball;
+import net.glowstone.util.nbt.CompoundTag;
+import net.glowstone.util.nbt.TagType;
+import org.bukkit.util.Vector;
+
+public class FireballStore<T extends GlowFireball> extends ProjectileStore<T> {
+
+    public static final String IS_INCENDIARY = "X-Glowstone-IsIncendiary";
```
mastercoms: Could you make these private like the other ones? Also, I feel like `glowstone:IsIncendiary` or if that's not valid, `glowstoneIsIncendiary` would be better.

mastercoms (dismissed):

src/main/java/net/glowstone/io/entity/FireballStore.java
Pr0methean: Made private and changed to "glowstone:IsIncendiary" and "glowstone:ExplosionPowerFloat" in aaadd0e2356f861f5a5090c2d63ee61f501c23f2.

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
Pr0methean: Done in aaadd0e2356f861f5a5090c2d63ee61f501c23f2.

src/main/java/net/glowstone/block/itemtype/ItemBow.java
```diff
@@ -0,0 +1,127 @@
...
+                launchedTippedArrow.copyFrom((PotionMeta) arrow.getItemMeta());
+                break;
+            case SPECTRAL_ARROW:
+                launchedArrow = player.launchProjectile(SpectralArrow.class);
+                break;
+            case AIR:
+                // Not in creative mode and have no arrow
+                break;
+            default:
+                player.getServer().getLogger()
```
mastercoms: `GlowServer.logger` (or is this for testing) and `log(Level, Supplier)`

Pr0methean: Done in f656da22b66794f0e340518025296d49d53a43a8 (also migrated to the `Supplier<String>` logging function).

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
```diff
@@ -0,0 +1,113 @@
+package net.glowstone.io.entity;
+
+import java.util.UUID;
+import java.util.logging.Level;
```
mastercoms: unused import

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.stream.Collectors;
+import net.glowstone.GlowServer;
+import net.glowstone.entity.GlowAreaEffectCloud;
+import net.glowstone.inventory.GlowMetaPotion;
+import net.glowstone.util.nbt.CompoundTag;
+import org.bukkit.Bukkit;
```
mastercoms: unused import

Pr0methean: Fixed in b599a16b37602fe0c22e8a8ca6a53d49bb3cff91

src/main/java/net/glowstone/io/entity/AreaEffectCloudStore.java
Pr0methean: Fixed in b599a16b37602fe0c22e8a8ca6a53d49bb3cff91

Issue #576, opened 2017-10-21
Java 9 support for downloaded libraries


TuxCoding:
Title: Java 9 support for downloaded libraries
Glowstone build: 4534b4a5

Java 9 changed the system class loader and therefore we can no longer call addURL to dynamically load the library files. [Paperclip](https://github.com/PaperMC/Paperclip) had the same issue and they solved it [by using a Java agent](https://github.com/PaperMC/Paperclip/blob/d0395f6304db49e4472dab9dbf247e994a1122b8/java9/src/main/java/com/destroystokyo/paperclip/Agent.java#L17).

```
[WARNING] java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
[WARNING] 	at net.glowstone.util.LibraryManager$LibraryDownloader.run(LibraryManager.java:101)
[WARNING] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
[WARNING] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
[WARNING] 	at java.base/java.lang.Thread.run(Thread.java:844)
```

gdude2002:
Thanks for creating the ticket - we do actually know about this one. @mastercoms has some plans for it, I think.

Pr0methean:
A temporary solution may be to use `ClassLoaders.AppClassLoader.appendToClassPathForInstrumentation`, although this will probably raise an illegal-reflection warning and be specific to Oracle and OpenJDK.

mastercoms:
Fixed in https://github.com/GlowstoneMC/Glowstone/commit/536305cb88faad3d1b36c34b1b9ae145cc6c14e2.

PR #615, opened 2017-12-17
Skip library loading when system ClassLoader isn't a URLClassLoader

Pr0methean:
Allows graceful degradation for #576

### DIFF
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -74,6 +74,10 @@ public final class LibraryManager {
 
         @Override
         public void run() {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -74,6 +74,10 @@ public void run() {
 
         @Override
         public void run() {
+            ClassLoader sysLoader = ClassLoader.getSystemClassLoader();
+            if (!(sysLoader instanceof URLClassLoader)) {
```
mastercoms: Add a severe logging message here describing the issue to server owners (and maybe tell them they will need to downgrade for now).

mastercoms (approved):

PR #622, opened 2017-12-20
Refactor 4 Message classes to take Location constructor param

Pr0methean:
New constructor overloads accept a Location parameter, eliminating lots of duplicate code to extract x,y,z,pitch,yaw. Also, methods returning only one instance of one or both classes now use Collections.singletonList.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowCreature.java
+++ b/src/main/java/net/glowstone/entity/GlowCreature.java
@@ -48,17 +48,11 @@ public class GlowCreature extends GlowLivingEntity implements Creature {
         List<Message> result = new LinkedList<>();
 
         // spawn mob
-        double x = location.getX();
-        double y = location.getY();
-        double z = location.getZ();
-        int yaw = Position.getIntYaw(location);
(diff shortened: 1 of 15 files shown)


Pr0methean:
This also removes unused imports in net.glowstone.entity, even if they are pre-existing.

mastercoms:
The collection returned by `Collections#singletonList` is immutable. I don't think we want that behavior.

Pr0methean:
Many of the existing invocations return Arrays.asList(), which can't be resized, anyway.

mastercoms (approved):

mastercoms:
Thank you for your one of many contributions to Glowstone! I've invited you to our organization, as per our [contributing policy](https://github.com/GlowstoneMC/Glowstone/blob/dev/.github/CONTRIBUTING.md).

Pr0methean:
@mastercoms Thanks! For now I'll still wait a while for reviews when I'm making significant functional changes, but not for the remaining CheckStyle fixes (except where the former and the latter turn out to overlap, e.g. #633 ).

PR #626, opened 2017-12-20
Lombokify GlowMinecart

Pr0methean:
Getter-setter pairs that were previously stubs now have a field behind them.

### DIFF
--- a/src/main/java/net/glowstone/entity/objects/GlowMinecart.java
+++ b/src/main/java/net/glowstone/entity/objects/GlowMinecart.java
@@ -29,6 +29,7 @@ import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.MaterialData;
 import org.bukkit.util.Vector;
 
+// TODO: Minecarts stop when there's an arrow in the way.
 public abstract class GlowMinecart extends GlowEntity implements Minecart {
 
     @Getter @Setter private volatile double damage;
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/objects/GlowMinecart.java
```diff
@@ -27,8 +29,17 @@
 import org.bukkit.material.MaterialData;
 import org.bukkit.util.Vector;
 
+// TODO: Minecarts stop when there's an arrow in the way.
 public abstract class GlowMinecart extends GlowEntity implements Minecart {
 
+    @Getter @Setter private volatile double damage;
```
mastercoms: Could the annotations be on separate lines please? Also, why are these volatile?

src/main/java/net/glowstone/entity/objects/GlowMinecart.java
```diff
@@ -27,8 +29,17 @@
 import org.bukkit.material.MaterialData;
 import org.bukkit.util.Vector;
 
+// TODO: Minecarts stop when there's an arrow in the way.
```
aramperes: Since Minecart behavior is essentially not implemented, I think this todo is too specific. Movement needs to be implemented, then collision checks with other entities (not just arrows) as well.

src/main/java/net/glowstone/entity/objects/GlowMinecart.java
Pr0methean: They're volatile because we can't guarantee they'll be accessed from only one thread.

mastercoms (approved):

aramperes (approved):

PR #627, opened 2017-12-20
Replace deprecated MapView.Scale.getValue() call

Pr0methean:
with ordinal()


kaenganxt:
`MapView.Scale.getValue()` is deprecated because it returns a magic value. In Bukkit/Spigot this generally does not mean that a method is going to be removed in the future. `ordinal()` does not solve this problem, as the integer value is still "magic". As there's no real alternative for the scale shift I think the call to `getValue()` is more clear and should be kept like it is.

PR #629, opened 2017-12-20
Add log4j configuration

Pr0methean:
Necessary to fix a warning at startup.


src/main/resources/log4j.xml
```diff
@@ -0,0 +1,17 @@
...
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
+    <appender name="console" class="org.apache.log4j.ConsoleAppender">
+        <param name="Target" value="System.out"/>
+        <layout class="org.apache.log4j.PatternLayout">
+            <param name="ConversionPattern" value="%-5p %c{1} - %m%n"/>
+        </layout>
+    </appender>
+
+    <root>
+        <priority value ="debug" />
```
smartboyathome: Since this is the "release" config, we probably should have the priority set to info.

smartboyathome (changes requested):

smartboyathome:
Also, in general, I think it would be more useful if log4j output were logged to a file rather than to the console.

src/main/resources/log4j.xml
```diff
@@ -0,0 +1,17 @@
...
+        <appender-ref ref="console" />
+    </root>
+
+</log4j:configuration>
```
aramperes: Add a blank line at the end (LF) for both files.

aramperes (changes requested):

Pr0methean:
I think it's an important feature of Minecraft that errors and warnings are available on the console, even with the more stable Mojang code, so I've set the production log4j.xml to handle messages like so:

- Warning or error: file and console
- Info: file only
- Debug: ignore

smartboyathome (approved):

aramperes (approved):

mastercoms (approved):

PR #632, opened 2017-12-22
Fix existing CheckStyle issues (B..BlockLava)

Pr0methean:


src/main/java/net/glowstone/block/blocktype/BlockCrops.java
```diff
@@ -68,12 +68,12 @@ public boolean canTickRandomly() {
     public void updateBlock(GlowBlock block) {
         GlowBlockState state = block.getState();
         int cropState = block.getData();
-        // we check light level on the above block, meaning crops needs at least one free block above it
-        // in order to grow naturally (vanilla behavior)
+        // we check light level on the above block, meaning crop needs at least one free block above
```
mastercoms: I think it should read "the crops need" and then "them" instead of it

src/main/java/net/glowstone/block/blocktype/BlockDirectDrops.java
```diff
@@ -16,29 +16,63 @@
...
     private final MaterialMatcher neededTool;
 
+    /**
+     * Creates an instance.
+     * @param dropType the type of item to drop when broken
+     * @param data the damage or block-data value of the dropped item
+     * @param amount the amount to drop
+     * @param neededTool the tool(s) that can break this block
+     */
     public BlockDirectDrops(Material dropType, int data, int amount, MaterialMatcher neededTool) {
```
mastercoms: I think the description here should be "Creates a block type that drops an item directly when broken", or something along those lines. Adapt the others to say something similar as well.

src/main/java/net/glowstone/block/blocktype/BlockCrops.java
Pr0methean: Done.

src/main/java/net/glowstone/block/blocktype/BlockDirectDrops.java
Pr0methean: Done.

PR #633, opened 2017-12-22
Refactor BossBar management

Pr0methean:
Was previously exposing BossBarManager's own list field to external modification.

### DIFF
--- a/src/main/java/net/glowstone/boss/BossBarManager.java
+++ b/src/main/java/net/glowstone/boss/BossBarManager.java
@@ -13,6 +13,10 @@ public class BossBarManager {
     private final List<GlowBossBar> bossBars = new ArrayList<>();
     private final GlowServer server;
 
+    /**
+     * Create an instance for the given server.
+     * @param server the server
+     */
     public BossBarManager(GlowServer server) {
         if (instance != null) {
             throw new RuntimeException("BossBar Manager has already been initialized.");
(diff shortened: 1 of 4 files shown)


Pr0methean:
Totally rewritten following discussion with kamcio96 on the Discord.

src/main/java/net/glowstone/boss/BossBarManager.java
```diff
@@ -1,83 +1,49 @@
...
-        }
-        this.server = server;
-        instance = this;
-    }
-
-    public static BossBarManager getInstance() {
-        return instance;
-    }
-
+public enum BossBarManager {
```
mastercoms: Please make this a `static class`.

src/main/java/net/glowstone/boss/BossBarManager.java
```diff
@@ -1,83 +1,49 @@
...
+    ;
```
mastercoms: One lonely semicolon, who needs to be destroyed.

mastercoms (dismissed):

src/main/java/net/glowstone/boss/BossBarManager.java
Pr0methean: Done; some authors recommend that utility classes that aren't going to be instantiated be zero-instance enums.

PR #641, opened 2017-12-23
Eliminate BossBarManager

Pr0methean:

PR #642, opened 2017-12-23
Boss bars for Wither & Ender Dragon

Pr0methean:
Merge #641 first. Should this wait for after 2018.0.0?


src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -854,11 +854,9 @@ public void damage(double amount, Entity source, DamageCause cause) {
         }
 
         // play sounds, handle death
-        if (health > 0) {
-            Sound hurtSound = getHurtSound();
-            if (hurtSound != null && !isSilent()) {
-                world.playSound(location, hurtSound, getSoundVolume(), getSoundPitch());
-            }
+        Sound hurtSound = health > 0 ? getHurtSound() : getDeathSound();
```
aramperes: What is the reason for this change? Death sound is already played here: 

https://github.com/Pr0methean/Glowstone/blob/eae90f1453fc76a33a59317a3877b3776e6a247e/src/main/java/net/glowstone/entity/GlowLivingEntity.java#L724-L729

src/main/java/net/glowstone/entity/monster/GlowBoss.java
```diff
@@ -0,0 +1,62 @@
...
+    public GlowBoss(Location loc, EntityType type, double maxHealth, BarColor color, BarStyle style,
+            BarFlag... barFlags) {
+        this(loc, type, maxHealth, type.getName(), color, style, barFlags);
+    }
+
+    @Override
+    public void setHealth(double health) {
+        super.setHealth(health);
+        if (health <= 0) {
+            bar.removeAll();
```
aramperes: I think this should be moved to the `remove()` method instead. It is called when the entity dies, as well as if it is removed by other factors like despawning or plugins.

Pr0methean: Done.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Reverted.

aramperes (approved):

kamcio96 (approved):

mastercoms:
I'm fine with merging this for this release because it isn't huge and doesn't have breaking changes.

PR #643, opened 2017-12-25
Implement ItemSpawnEvent and ItemDespawnEvent

aramperes:
This PR implements ItemSpawnEvent and ItemDespawnEvent, and replaces GlowItem constructor usage by the `GlowWorld#dropItem` method.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -82,6 +82,7 @@ import org.bukkit.entity.Arrow;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.FallingBlock;
+import org.bukkit.entity.Item;
 import org.bukkit.entity.LightningStrike;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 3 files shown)


kaenganxt (approved):

aramperes:
Do not merge, I found an issue with this but can't fix it for a moment

mastercoms (approved):

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1417,18 +1419,18 @@ public GlowEntity spawn(Location location, Class<? extends GlowEntity> clazz, Sp
...
                     spawnEvent = EventFactory.callEvent(new EntitySpawnEvent(entity));
                 }
-                if (!spawnEvent.isCancelled()) {
-                    List<Message> spawnMessage = entity.createSpawnMessage();
-                    getRawPlayers().stream().filter(player -> player.canSeeEntity(impl)).forEach(player -> player.getSession().sendAll(spawnMessage.toArray(new Message[spawnMessage.size()])));
-                } else {
+                if (spawnEvent != null && spawnEvent.isCancelled()) {
                     // TODO: separate spawning and construction for better event cancellation
                     entity.remove();
+                } else {
```
mastercoms, line 1431: Shouldn't this branch also require non null? Wrap the whole if else in a null check.

aramperes: No, if the event is null, it means that the event was fired elsewhere. The spawn message is still necessary.

mastercoms: Oh whoops, I don't know how, but I misread that null check as one for `entity`, not `spawnEvent`.

mastercoms (approved):

PR #644, opened 2017-12-28
Fix player interact issues

kaenganxt:
This PR fixes several issues with the `PlayerInteractEvent` and the `rightClickBlock()` and `rightClickAir()` methods of `ItemType`.

The following bugs are fixed:
- Block clicks fired an additional interact event with left click air (this was checked in `PlayerSwingArmHandler` and failed because `GlowLivingEntity#getTargetBlock` returned wrong values)
- Right clicking on a block with an unplaceable item (e.g. a nether star) fired an additional right click air event because the client sends the place block and use item packets in this case
- When trying to place a placeable block and it fails (e.g. tall grass on stone), the client only sends a use item packet, which only fired a right click air interact. There's a right click block now fired instead (The block and block face values are computed by the `UseItemHandler`). This includes handling all internal right click actions.
- Water behind a wall could be pickup up with a bucket
- In some cases, interacting with blocks was possible even when sneaking and a block in the hand. (Solved by requiring both hands to be empty)

Other changes:
- Emptying a bucket in survival mode now changes the item type to `BUCKET`
- Implemented rightClickBlock handlers for all ItemTypes which previously depended on rightClickAir even if a block was clicked
- Cleaned up boat placement

There still is an issue with this implementation:
When clicking next to an interactable block (e.g. a lever), the lever is still flipped because the current raytrace (BlockIterator) ignores the size of the blocks. Also, the `clickedAt` vector in the `ItemType#rightClickBlock` method doesn't have correct values in it because of this problem.
So a proper raytracing method has to be implemented in the future, which would be useful also for other things (like filling a bucket).

Everything regarding the `PlayerInteractEvent` has been tested in comparison to Spigot behaviour. The results were not equal in all cases, but that's because Spigot isn't handling everything correctly itself ;) (Some interacts even seem quite random).

One difference I noticed was that for right and left click air events, Spigot always used `BlockFace.SOUTH` whereas we are setting the value to null. So there may be plugins that don't expect a null value here. Should we also set the face to `SOUTH` (which doesn't really make sense), should we use something more general like `BlockFace.SELF` or should we still use null?

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemBoat.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemBoat.java
@@ -1,8 +1,7 @@
 package net.glowstone.block.itemtype;
 
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
+import java.util.Set;
+import net.glowstone.block.GlowBlock;
 import net.glowstone.entity.GlowPlayer;
 import org.bukkit.Location;
 import org.bukkit.Material;
@@ -10,7 +9,9 @@ import org.bukkit.TreeSpecies;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.entity.Boat;
+import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.util.Vector;
 
 public class ItemBoat extends ItemType {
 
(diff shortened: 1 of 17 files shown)


src/main/java/net/glowstone/block/itemtype/ItemBoat.java
```diff
@@ -22,24 +23,31 @@ public ItemBoat(TreeSpecies woodType) {
...
-        if (first.isPresent()) {
-            Block block = first.get();
-            Location location = block.getRelative(BlockFace.UP).getLocation();
+        // Executed when the player clicks on water that doesn't have a block beneath
+        placeBoat(player);
+    }
+
+    @Override
+    public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face, ItemStack holding, Vector clickedLoc, EquipmentSlot hand) {
+        // Two cases are handled here: Either the player clicked on a block on the land or beneath water
```
mastercoms, line 32: I can't place boats under water in vanilla.

src/main/java/net/glowstone/block/itemtype/ItemKnowledgeBook.java
```diff
@@ -19,6 +41,8 @@ public void rightClickAir(GlowPlayer player, ItemStack holding) {
                         .getRecipeByKey(recipe), true);
                 }
             }
+
+            holding.setAmount(0);
```
mastercoms: It should also set the type to air.

mastercoms (changes requested):

mastercoms:
We should be giving an air block with a face of SELF in the event.

src/main/java/net/glowstone/block/itemtype/ItemKnowledgeBook.java
kaenganxt: Both handlers create an empty stack if necessary ([example](https://github.com/GlowstoneMC/Glowstone/pull/644/files#diff-7544280850f3676c540cf78ab99c1db1R97)) after executing `rightClickAir` or `rightClickBlock`.

mastercoms: Then why did knowledge books stay in my inventory after use?

src/main/java/net/glowstone/block/itemtype/ItemBoat.java
kaenganxt: I fixed boats being placeable underwater. The placement check is not perfect as vanilla seems to do a bounding box check, but I think it's good enough for now.

aramperes (approved):

mastercoms (approved):

PR #645, opened 2017-12-29
Update boss bars when necessary

Pr0methean:
I.e. when a player or boss enters or leaves a world.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -714,12 +714,6 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
         super.remove();
     }
 
-    private void clearBossBars() {
-        for (BossBar bar : bossBars) {
-            removeBossBar(bar);
-        }
-    }
-
     public void remove(boolean async) {
         knownChunks.clear();
         chunkLock.clear();
@@ -1068,6 +1062,12 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
             }
         } finally {
             worldLock.readLock().unlock();
+        }
+    }
+
+    private void clearBossBars() {
+        for (BossBar bar : bossBars) {
+            removeBossBar(bar);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/entity/monster/GlowBoss.java
```diff
@@ -10,11 +12,34 @@
 
 /**
  * A monster with a boss bar.
+ *
+ * @author Chris Hennick
```
aramperes: While you did create that class, we don't use the `@author` tag in this project.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -707,9 +710,16 @@ public void remove() {
             scoreboard.unsubscribe(this);
             scoreboard = null;
         }
+        clearBossBars();
         super.remove();
     }
 
+    private void clearBossBars() {
```
aramperes: This method should go after the `remove(boolean async)` method.

aramperes (dismissed):

src/main/java/net/glowstone/entity/GlowPlayer.java
Pr0methean: Done.

src/main/java/net/glowstone/entity/monster/GlowBoss.java
mastercoms: Well there still are classes with that tag, but only for historical reasons.

Pr0methean: Reverted.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -707,6 +710,7 @@ public void remove() {
             scoreboard.unsubscribe(this);
             scoreboard = null;
         }
+        clearBossBars();
```
mastercoms, line 713: Why are boss bars being updated for a player that was removed from the server?

Pr0methean: To avoid memory leaks resulting from the (logged-out) player still holding a reference to the boss bar, or vice-versa. (Although if you can find a way to prevent this with WeakReference&lt;?> and/or WeakHashMap&lt;?,?> instead, then I'm all ears.)

mastercoms: I guess this is an acceptable solution for now, but later on we may want to investigate using weaker references to prevent the problem.

mastercoms (approved):

PR #647, opened 2017-12-29
Fix existing CheckStyle issues (GlowEntityEquipment--GlowMapView)

Pr0methean:


src/main/java/net/glowstone/chunk/GlowChunkSnapshot.java
```diff
@@ -11,17 +12,31 @@
  */
 public class GlowChunkSnapshot implements ChunkSnapshot {
 
-    private final int x, z;
+    @Getter private final int x;
```
mastercoms: Separate lines.

mastercoms (dismissed):

src/main/java/net/glowstone/chunk/GlowChunkSnapshot.java
Pr0methean: Done; also corrected z.

mastercoms (approved):

PR #648, opened 2017-12-29
Refactor GlowMeta*

Pr0methean:
List getters and setters now consistently make defensive copies.
Non-list getters and setters are lombokified (including a few that were stubs). Copy constructors now
handle *any* implementations of their respective interfaces, improving
plugin-friendliness. GlowMetaShield now extends GlowMetaBanner.
GlowMetaSkull.setOwningPlayer now checks the ProfileCache but still
doesn't wait for a profile to be downloaded. Missing Javadocs are added.

### DIFF
--- a/src/main/java/net/glowstone/inventory/GlowMetaSkull.java
+++ b/src/main/java/net/glowstone/inventory/GlowMetaSkull.java
@@ -43,6 +43,13 @@ public class GlowMetaSkull extends GlowMetaItem implements SkullMeta {
     ////////////////////////////////////////////////////////////////////////////
     // Internal stuff
 
+    /**
+     * Deserializes an instance as specified in {@link
+     * org.bukkit.configuration.serialization.ConfigurationSerializable}.
+     *
+     * @param data a serialized instance
+     * @return the instance as a GlowMetaSkull
+     */
     public static GlowMetaSkull deserialize(Map<String, Object> data) {
         GlowMetaSkull result = new GlowMetaSkull(null);
         if (data.containsKey("owner")) {
--- a/src/main/java/net/glowstone/inventory/GlowMetaSkull.java
+++ b/src/main/java/net/glowstone/inventory/GlowMetaSkull.java
@@ -6,6 +6,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicReference;
 import net.glowstone.GlowOfflinePlayer;
 import net.glowstone.GlowServer;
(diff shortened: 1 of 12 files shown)


smartboyathome:
Rather than defensive copies, you should think about wrapping the lists in Collections.unmodifiableList() where possible.

Pr0methean:
@smartboyathome If I wrap a list in Collections.unmodifiableList while there are still other references to it, from which it can be modified, then it won't be thread-safe. Plus, even if we returned a thread-safe read-only view, we'd still run into trouble when people were assuming they were getting a snapshot.

aramperes (dismissed):

src/main/java/net/glowstone/inventory/GlowMetaSkull.java
```diff
@@ -93,24 +120,44 @@ public boolean setOwner(String name) {
...
+        return setOwningPlayerInternal(owningPlayer);
+    }
+
+    private boolean setOwningPlayerInternal(OfflinePlayer owningPlayer) {
         if (owningPlayer instanceof GlowOfflinePlayer) {
             GlowOfflinePlayer impl = (GlowOfflinePlayer) owningPlayer;
-            this.owner = impl.getProfile();
+            this.owner.set(impl.getProfile());
+            return true;
+        } else {
```
kaenganxt: `owningPlayer` could also be a `GlowPlayer` (which does not extend `GlowOfflinePlayer`).

kaenganxt (changes requested):

src/main/java/net/glowstone/inventory/GlowMetaSkull.java
Pr0methean: Handled that case.

kaenganxt (approved):

src/main/java/net/glowstone/inventory/GlowMetaSkull.java
```diff
@@ -147,6 +148,8 @@ private boolean setOwningPlayerInternal(OfflinePlayer owningPlayer) {
             GlowOfflinePlayer impl = (GlowOfflinePlayer) owningPlayer;
             this.owner.set(impl.getProfile());
             return true;
+        } else if (owningPlayer instanceof GlowPlayer) {
+            this.owner.set(((GlowPlayer) owningPlayer).getProfile());
```
aramperes: Missing a return statement, then it should be good to go.

aramperes (approved):

PR #649, opened 2017-12-29
Introducing Java 9 support! (Fixes #576)

### LINKED ISSUE #576: Java 9 support for downloaded libraries

smartboyathome:
- Glowstone has a new GlowBootstrap main() method, in charge of downloading any depndencies before starting the application.
- GlowMain now contains the contents of the old GlowServer.main() method.
- LibraryManager was rewritten to use the Jar Class Loader library, rather than relying on URLClassLoader.addURL().
- Still works on Java 8!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,6 +103,12 @@
       <artifactId>jocl-main</artifactId>
       <version>2.3.2</version>
     </dependency>
+    <!-- JCL: Jar Class Loader -->
+    <dependency>
+      <groupId>org.xeustechnologies</groupId>
+      <artifactId>jcl-core</artifactId>
+      <version>2.8</version>
+    </dependency>
     <!-- Tests -->
     <dependency>
       <groupId>junit</groupId>
--- a/src/main/java/net/glowstone/GlowBootstrap.java
+++ b/src/main/java/net/glowstone/GlowBootstrap.java
@@ -1,0 +1,27 @@
+package net.glowstone;
+
+import net.glowstone.util.LibraryManager;
+import org.xeustechnologies.jcl.JarClassLoader;
+import org.xeustechnologies.jcl.JclObjectFactory;
+
+public class GlowBootstrap {
+    /**
+     * Creates a new server on TCP port 25565 and starts listening for connections.
+     *
+     * @param args The command-line arguments.
+     */
+    public static void main(String... args) {
+        LibraryManager libraryManager = new LibraryManager();
+        JarClassLoader libraryClassLoader = libraryManager.run();
+
+        if (libraryClassLoader != null) {
+            JclObjectFactory objectFactory = JclObjectFactory.getInstance();
+            Object main = objectFactory.create(libraryClassLoader, "net.glowstone.GlowMain", (Object) args);
+            Thread mainThread = new Thread((Runnable) main);
+            mainThread.setContextClassLoader(libraryClassLoader);
+            mainThread.start();
+        }
+    }
+
+
+}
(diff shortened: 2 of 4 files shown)


Pr0methean:
NB: Until https://github.com/rzwitserloot/lombok/issues/985 can at least be worked around, we'll still have to *build* on JDK 8 and can't do a multirelease jar.

hibo98:
Hey, I yesterday also tried to figure out a fix for this problem.
But I think my fix is one of the quick and dirty ones.
Also because logging is not correctly initialized in the agent.
https://github.com/hibo98/Glowstone/commit/d04866cf8cab15d491c1e77b63789d232d8e38b9

So that looks quiet good

mastercoms (approved):

smartboyathome:
Good note. It's true that it won't build on Java 9, but I was more focused on getting it to run. The key to this was minimizing the classes referenced in the bootstrap phase. That's one reason I created GlowMain and reference it only as a runnable, trying hard to avoid triggering any static variables from being loaded.

hibo98:
I test this pull-request on my server and it did not work!
When the server trys to connect the database then it throws a SQLException.
```
java.sql.SQLException: No suitable driver found for jdbc:mysql://*******
        at java.sql/java.sql.DriverManager.getConnection(Unknown Source)
        at java.sql/java.sql.DriverManager.getConnection(Unknown Source)
...
```
So the mysql driver is not loaded!
Verison is `2018.0.0-SNAPSHOT.3f96747`

--- EDITED ---

This Problem occurs under Java 8 too.

```
java.sql.SQLException: No suitable driver found for jdbc:mysql://*******
        at java.sql.DriverManager.getConnection(DriverManager.java:689)
        at java.sql.DriverManager.getConnection(DriverManager.java:247)
...
```

mastercoms:
I have reverted it. @smartboyathome please remake the PR.

smartboyathome:
@hibo98 I cannot reproduce this. I ran it on my Win10 machine and it starts up just fine, both under Java 8 and Java 9. What OS are you using? Any specific configuration I should be aware of?

hibo98:
Like you I am using Windows 10, with installed Oracle Java 8 or 9. And the start up also fine.
The Problem occurs, when my plugin try to connect to the database.
So the loading of the classes probably don't work.
You have to test for example with a plugin that uses a mysql connection.

smartboyathome:
Thanks. I'll play around some tonight, but https://stackoverflow.com/questions/42052856/java-9-classpath-and-library-path-extension worries me, seems like plugins that rely on JDBC just may not work with how Glowstone downloads its libraries anymore.

PR #652, opened 2017-12-29
Fix library managment for Java 9

hibo98:
The ClassPathAgent starts before the GlowServer and saves the Instrumentation for future use. (This Works only as of Java 9) The LibraryManager uses as of Java 9 the ClassPathAgent to add jars to the classpath.

Test with Oracle and OpenJDK 8+9

### DIFF
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -97,7 +97,7 @@ public final class LibraryManager {
 
             // hack it onto the classpath
             try {
-                String[] javaVersion = System.getProperty("java.version").split("\\.");
+                String[] javaVersion = System.getProperty("java.version").split("-")[0].split("\\.");
                 if (Integer.parseInt(javaVersion[0]) >= 9) {
                     ClassPathAgent.addJarFile(new JarFile(file));
                 } else {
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -113,10 +97,15 @@ public void run() {
 
             // hack it onto the classpath
             try {
-                Method method = URLClassLoader.class.getDeclaredMethod("addURL", URL.class);
-                method.setAccessible(true);
-                method.invoke(sysLoader, file.toURI().toURL());
-            } catch (ReflectiveOperationException | MalformedURLException e) {
+                String[] javaVersion = System.getProperty("java.version").split("\\.");
+                if (Integer.parseInt(javaVersion[0]) >= 9) {
+                    ClassPathAgent.addJarFile(new JarFile(file));
```
smartboyathome: Looks like the specific instrumentation method that ClassPathAgent uses can be used with Java 8 and below. Why not just use that there?

hibo98: But I have the problem that Laucher-Agents don't work with Java 8

smartboyathome: You're right, its not explicitly called out anywhere that I can see, but I can see Launcher-Agent-Class included in the doc page for [Java 9](https://docs.oracle.com/javase/9/docs/api/java/lang/instrument/package-summary.html) but not for [Java 8](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/package-summary.html).

kaenganxt (approved):

Pr0methean:
I've tested this in combined staging with my own pending PRs, and it seems to work fine.

PR #656, opened 2017-12-31
Extracts generate(world,random,x,y,z) as an interface

Pr0methean:
De-duplicates the Javadoc for methods with that signature. May also help
de-duplicate code in the future, especially between BlockPopulator
subclasses.


Pr0methean:
Additionally refactors some generate() implementations for style and to reduce excessive nesting.

mastercoms:
Perhaps the interface should be called TerrainObject so there is more consistency with the package name.

aramperes (approved):

PR #658, opened 2017-12-31
Added testforblocks command, along with unit tests

smartboyathome:
Needed to refactor the GlowChunk object out of BlockType.createBlockEntity() in order to make it easier to mock the GlowBlocks without touching GlowChunk.

### DIFF
--- a/src/test/java/net/glowstone/testutils/InMemoryBlockStorage.java
+++ b/src/test/java/net/glowstone/testutils/InMemoryBlockStorage.java
@@ -1,5 +1,6 @@
 package net.glowstone.testutils;
 
+import static org.mockito.Matchers.anyInt;
 import static org.mockito.Mockito.when;
 
 import java.util.HashMap;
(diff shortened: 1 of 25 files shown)


src/test/java/net/glowstone/testutils/InMemoryBlockStorage.java
```diff
@@ -0,0 +1,52 @@
...
+import net.glowstone.block.entity.BlockEntity;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.powermock.api.mockito.PowerMockito;
+
+public class InMemoryBlockStorage {
+    private Map<Location, GlowBlock> mockBlocks = new HashMap<>();
+
+    private GlowBlock createMockBlock(Location location, Material type, byte data) {
+        GlowBlock mockBlock = PowerMockito.mock(GlowBlock.class);
```
Pr0methean: I think you can use plain Mockito here, since the class is no longer final.

mastercoms (changes requested):
You're doing a getBlock call even if there is no block entity to be created.

smartboyathome:
I didn't see anything wrong with this since GlowBlock is little more than a thin wrapper that points back to GlowChunk, and this only happens when a block's type is changed. Don't think lambda is going to be much better here, though, so I'll revert the changes to blocktypes though and just mock the getBlock function in GlowChunk.

mastercoms (approved):

PR #662, opened 2018-01-02
Fix CheckStyle issues in net/GlowNetworkServer.java

Pr0methean:


src/main/java/net/glowstone/net/GlowNetworkServer.java
```diff
@@ -8,13 +8,20 @@
 /**
  * Represents a network server.
  *
- * Modified implementation of {@link com.flowpowered.network.NetworkServer}.
+ * <p>Modified implementation of {@link com.flowpowered.network.NetworkServer}.</p>
```
mastercoms: Don't use a closing p tag.

Pr0methean: Fixed. I'll go remove them all on the master branch as well.

mastercoms: Sorry for missing those.

Pr0methean: No, I mean in the branch my other style-fix PRs will be split from. I'm not sure there are any pre-existing closing &lt;/p>'s.

PR #677, opened 2018-01-02
Fix CheckStyle issues in io/anvil/RegionFile.java

Pr0methean:


src/main/java/net/glowstone/io/anvil/RegionFile.java
```diff
@@ -38,53 +38,54 @@
...
+import lombok.Getter;
 import net.glowstone.GlowServer;
 import net.glowstone.util.config.ServerConfig.Key;
 import org.bukkit.Bukkit;
 
 /**
  * Interfaces with region files on the disk
  *
- * Region File Format
+ * <h2>Region File Format</h2>
```
mastercoms: Remove the h2.

Pr0methean: Done.

PR #734, opened 2018-01-02
Fix CheckStyle issues in generator/objects/OreVein.java

Pr0methean:


src/main/java/net/glowstone/generator/objects/OreVein.java
```diff
@@ -34,25 +40,25 @@ public boolean generate(World world, Random random, int sourceX, int sourceY, in
...
-            double vRadius = (Math.sin(i * (float) Math.PI / amount) + 1 * q + 1) / 2.0D;
-            for (int x = (int) (originX - hRadius); x <= (int) (originX - hRadius); x++) {
-                double pX = (x + 0.5D - originX) / hRadius;
-                pX *= pX;
-                if (pX >= 1) {
+            double radiusH = (Math.sin(i * (float) Math.PI / amount) + 1 * q + 1) / 2.0D;
+            double radiusV = (Math.sin(i * (float) Math.PI / amount) + 1 * q + 1) / 2.0D;
+            for (int x = (int) (originX - radiusH); x <= (int) (originX - radiusH); x++) {
+                double px = (x + 0.5D - originX) / radiusH;
+                px *= px;
```
mastercoms: I'd really like to decipher the name of this variable.

Pr0methean: These are transformed coordinates, with the coordinate space defined to turn the ore vein (which is a spheroid having one radius on the xz plane and another on the y axis) into the unit sphere. They are then squared to check whether the point at the center of the block is inside the spheroid. Hopefully my latest commit clears things up.

PR #742, opened 2018-01-02
Fix CheckStyle issues in inventory/GlowPlayerInventory.java

Pr0methean:


src/main/java/net/glowstone/inventory/GlowPlayerInventory.java
```diff
@@ -378,17 +393,24 @@ public Location getLocation() {
...
+        if (type == Material.AIR) {
+            return 0;
+        }
         int numCleared = 0;
         for (int i = 0; i < getSize(); ++i) {
             ItemStack stack = getItem(i);
             if (stack != null && (type == null || stack.getType() == type) && (data == null || stack
                 .getData().equals(data))) {
                 setItem(i, InventoryUtil.createEmptyStack());
-                if (!InventoryUtil.isEmpty(stack)) {
```
mastercoms: This check still needs to be here in the case of type being null.

Pr0methean: Reverted.

PR #744, opened 2018-01-02
Fix CheckStyle issues in net.glowstone.block.itemtype

Pr0methean:


src/main/java/net/glowstone/block/itemtype/ItemType.java
```diff
@@ -172,7 +173,7 @@ public Context getContext() {
      */
     public void rightClickBlock(GlowPlayer player, GlowBlock target, BlockFace face,
         ItemStack holding, Vector clickedLoc, EquipmentSlot hand) {
-        if (placeAs != null && (placeAs.getContext() == Context.ANY || placeAs.getContext() == Context.BLOCK)) {
+        if (placeAs != null) {
```
mastercoms: There should still be a check here for if the context is not AIR, or we can keep this more verbose check in case more contexts come up in the future.

Pr0methean: Fixed.

PR #753, opened 2018-01-02
Fix CheckStyle issues in inventory/WindowClickLogic.java

Pr0methean:


src/main/java/net/glowstone/inventory/WindowClickLogic.java
```diff
@@ -76,17 +77,18 @@ public static ClickType getClickType(int mode, int button, int slot) {
                 }
                 break;
 
-            case 5: // drag
-                break;
-
             case 6:
                 return ClickType.DOUBLE_CLICK;
+            case 5: // drag
```
kaenganxt: Is this correct or is it just not implemented yet?

Pr0methean: Marked as TODO to be safe.

PR #776, opened 2018-01-03
Fix CheckStyle issues in entity/passive/GlowSheep.java

Pr0methean:


src/main/java/net/glowstone/entity/passive/GlowSheep.java
```diff
@@ -22,19 +22,24 @@
...
+        } else if (8684 > colorpc) {
             setColor(DyeColor.BLACK);
-        } else if (colorpc >= 8684 && 9184 > colorpc) {
+        } else if (9184 > colorpc) {
             setColor(DyeColor.SILVER);
-        } else if (colorpc >= 9184 && 9684 > colorpc) {
+        } else if (9684 > colorpc) {
             setColor(DyeColor.GRAY);
-        } else if (colorpc >= 9684 && 9984 > colorpc) {
+        } else if (9984 > colorpc) {
```
liach, line 42: Why switching these?

Pr0methean: The deleted part is always true when we get to it. IntelliJ raises a warning about that.

PR #779, opened 2018-01-03
Fix CheckStyle issues in util/nbt/NbtOutputStream.java

Pr0methean:


mastercoms:
It seems like this does not rename from the original class.

PR #785, opened 2018-01-03
Fix CheckStyle issues in util/nbt/NbtReadLimiter.java

Pr0methean:


mastercoms:
It seems like this does not rename from the original class.

Pr0methean:
Thanks for catching this; I'll do the full rename as part of the big leftovers PR.

PR #789, opened 2018-01-03
Fix CheckStyle issues in util/noise/PerlinNoise.java

Pr0methean:


src/main/java/net/glowstone/util/noise/PerlinNoise.java
```diff
@@ -31,10 +35,26 @@ public PerlinNoise(Random rand) {
     }
 
     public static int floor(double x) {
-        int iX = (int) x;
-        return x < iX ? iX - 1 : iX;
+        int ix = (int) x;
```
mastercoms: Rename to floored.

Pr0methean: Done.

PR #799, opened 2018-01-03
some base code for a path finding algorithm (using A*)

Salil999:
Some base code for having a path finding algorithm running. I wasn't sure which class to use to compare "locations" to, so I used Bukkit's Vector class.

I referenced these two links:
- https://www.redblobgames.com/pathfinding/a-star/introduction.html#astar
- https://dzenanhamzic.com/2016/12/16/a-star-a-algorithm-implementation-in-java/

## Note
All the failed checks are really odd formatting issues. When I load it up in IntelliJ, it works perfectly fine. If anyone can suggest, that'd be great!

### DIFF
--- a/src/main/java/net/glowstone/util/PathFinder.java
+++ b/src/main/java/net/glowstone/util/PathFinder.java
@@ -1,8 +1,17 @@
 package net.glowstone.util;
 
-import org.bukkit.util.Vector;
+import java.util.Queue;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.HashMap;
+import java.util.PriorityQueue;
+import java.util.Collections;
+import java.util.Arrays;
 
-import java.util.*;
+import org.bukkit.util.Vector;
 
 /**
  * General class for path finding.
@@ -66,18 +75,18 @@ public class PathFinder {
      * Tries to find a path between the start and end points. This method currently
      * does not handle cases where a path could not potentially exist.
      */
-    public void AStarSearch() {
+    public void aStarSearch() {
         Queue<Vector> frontier = new PriorityQueue<>();
         this.fromPath = new HashMap<>();
         Map<Vector, Double> costSoFar = new HashMap<>();
-//      HashSet<Vector> visitedNodes = new HashSet<>();
+        // HashSet<Vector> visitedNodes = new HashSet<>();
 
         // Add the starting node to our frontier
         // TODO: Might have to make a wrapper class for Vectors and computed distances to compare in the priority queue
         frontier.add(startNode);
 
         // Add the starting node to our hash map
-        fromPath.put(startNode, null);
+        this.fromPath.put(startNode, null);
 
         // Cost from the start is 0
         costSoFar.put(startNode, 0.0);
@@ -103,7 +112,7 @@ public class PathFinder {
                 if (!costSoFar.containsKey(nextNode) || newCost < costSoFar.get(nextNode)) {
                     costSoFar.put(nextNode, newCost);
                     frontier.add(nextNode);
-                    fromPath.put(nextNode, current);
+                    this.fromPath.put(nextNode, current);
                 }
 
             }
--- a/src/main/java/net/glowstone/util/PathFinder.java
+++ b/src/main/java/net/glowstone/util/PathFinder.java
@@ -11,221 +11,224 @@ import java.util.*;
 
 public class PathFinder {
 
-	private final Vector startNode;
-	private final Vector endNode;
-	private List<Vector> path;
-	private Map<Vector, Vector> fromPath;
+    private final Vector startNode;
+    private final Vector endNode;
+    private List<Vector> path;
+    private Map<Vector, Vector> fromPath;
 
 
-	/**
-	 * Creates a new PathFinder object.
-	 *
-	 * @param startNode Starting position
-	 * @param endNode Ending position
-	 */
-	public PathFinder(final Vector startNode, final Vector endNode) {
-		// Save our start and end points
-		this.startNode = startNode;
-		this.endNode = endNode;
-		path = new ArrayList<Vector>();
-	}
+    /**
+     * Creates a new PathFinder object.
+     *
+     * @param startNode Starting position
+     * @param endNode Ending position
+     */
+    public PathFinder(final Vector startNode, final Vector endNode) {
+        // Save our start and end points
+        this.startNode = startNode;
+        this.endNode = endNode;
+        this.path = new ArrayList<Vector>();
+    }
 
 
-	/**
-	 * Gets the start (initial) node
-	 * @return Vector object
-	 */
-	public Vector getStartNode() {
-		return startNode;
-	}
+    /**
+     * Gets the start (initial) node.
+     *
+     * @return Vector object
+     */
+    public Vector getStartNode() {
+        return startNode;
+    }
 
 
-	/**
-	 * Gets the end (goal) node.
-	 * @return Vector object
-	 */
-	public Vector getEndNode() {
-		return endNode;
-	}
-
-	/**
-	 * Gets the path once it has been constructed.
-	 *
-	 * @note This method should only be called AFTER AStarSearch has been called
-	 * @return List of Vector objects in order from start to end
-	 */
-	public List<Vector> getPath() {
-		return path;
-	}
+    /**
+     * Gets the end (goal) node.
+     *
+     * @return Vector object
+     */
+    public Vector getEndNode() {
+        return endNode;
+    }
 
 
-	/**
-	 * Tries to find a path between the start and end points. This method currently
-	 * DOES NOT handle cases where a path could not potentially exist.
-	 */
-	public void AStarSearch() {
-		Queue<Vector> frontier = new PriorityQueue<>();
-		fromPath = new HashMap<>();
-		Map<Vector, Double> costSoFar = new HashMap<>();
-//		HashSet<Vector> visitedNodes = new HashSet<>();
-
-		// Add the starting node to our frontier
-		// TODO: Might have to make a wrapper class for Vectors and computed distances to compare in the priority queue
-		frontier.add(startNode);
-
-		// Add the starting node to our hash map
-		fromPath.put(startNode, null);
-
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.*;
+
+/**
+ * General class for path finding.
+ * TODO: Implement other path finding algorithms (BFS, DFS, D*, potentially more)
+ */
+
+public class PathFinder {
+
+	private final Vector startNode;
```
aramperes: Use 4 spaces instead of tab characters.

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+	 * Creates a new PathFinder object.
+	 *
+	 * @param startNode Starting position
+	 * @param endNode Ending position
+	 */
+	public PathFinder(final Vector startNode, final Vector endNode) {
+		// Save our start and end points
+		this.startNode = startNode;
+		this.endNode = endNode;
+		path = new ArrayList<Vector>();
```
aramperes: Change to `this.path = new ArrayList<>();`

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+
+	/**
+	 * Gets the start (initial) node
```
aramperes: Add period at end, and blank line before `@return` line.

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+	 * @return List of Vector objects in order from start to end
+	 */
+	public List<Vector> getPath() {
+		return path;
+	}
+
+
+	/**
+	 * Tries to find a path between the start and end points. This method currently
+	 * DOES NOT handle cases where a path could not potentially exist.
```
aramperes: Un-capitalize 'Does not'

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+
+
+	/**
+	 * Gets the neighbors that are touching the corners of the current vector.
+	 *
+	 * @param v The vector in which all diagnol neighbors will be generated with respect to
+	 * @return Set of Vector objects
+	 */
+	private Set<Vector> cornerNeighbors(final Vector v) {
+		Vector[] diagnolVectors = new Vector[] {
```
aramperes: `diagonal`

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+			if (current == endNode) {
+				break;
+			}
+
+			// Get all the neighbors
+			Set<Vector> neighbors = neighbors(current);
+
+			// Compute the distance for all the neighbors.
+			for (final Vector nextNode : neighbors) {
+				Double newCost = costSoFar.get(current) + current.distanceSquared(nextNode);
```
aramperes: `double` instead of `Double`

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+
+	/**
+	 * Tries to find a path between the start and end points. This method currently
+	 * DOES NOT handle cases where a path could not potentially exist.
+	 */
+	public void AStarSearch() {
+		Queue<Vector> frontier = new PriorityQueue<>();
+		fromPath = new HashMap<>();
```
aramperes: Add `this` before.

aramperes (changes requested):
This is my code review, I'll do a review of the actual algorithm later.

src/main/java/net/glowstone/util/PathFinder.java
Salil999: Do you want me to do this for all member variables?

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,243 @@
...
+            new Vector(v.getBlockX() - 1, v.getBlockY() + 1, v.getBlockZ() + 1),
+            new Vector(v.getBlockX() - 1, v.getBlockY() + 1, v.getBlockZ() - 1),
+
+            // Corners one level BELOW the current vector (in y-direction)
+            new Vector(v.getBlockX() + 1, v.getBlockY() - 1, v.getBlockZ() + 1),
+            new Vector(v.getBlockX() + 1, v.getBlockY() - 1, v.getBlockZ() - 1),
+            new Vector(v.getBlockX() - 1, v.getBlockY() - 1, v.getBlockZ() + 1),
+            new Vector(v.getBlockX() - 1, v.getBlockY() - 1, v.getBlockZ() - 1)
+        };
+        return new HashSet<>(Arrays.asList(cornerVectors));
```
aramperes, line 241: Use `Sets.newHashSet(cornerVectors)`

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,243 @@
...
+     * Creates a new PathFinder object.
+     *
+     * @param startNode Starting position
+     * @param endNode Ending position
+     */
+    public PathFinder(final Vector startNode, final Vector endNode) {
+        // Save our start and end points
+        this.startNode = startNode;
+        this.endNode = endNode;
+        this.path = new ArrayList<Vector>();
```
liach, line 39: java8, infer generic types plz

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,243 @@
...
+        return path;
+    }
+
+
+    /**
+     * Tries to find a path between the start and end points. This method currently
+     * does not handle cases where a path could not potentially exist.
+     */
+    public void findPathAStar() {
+        Queue<Vector> frontier = new PriorityQueue<>();
```
liach, line 79: Hey we need a correct comparator!

src/main/java/net/glowstone/util/PathFinder.java
```diff
@@ -0,0 +1,243 @@
...
+ * General class for path finding.
+ * TODO: Implement other path finding algorithms (BFS, DFS, D*, potentially more)
+ */
+
+public class PathFinder {
+
+    private final Vector startNode;
+    private final Vector endNode;
+    private List<Vector> path;
+    private Map<Vector, Vector> fromPath;
```
liach, line 26: Can we map the vectors into longs so that we can be slightly more efficient (with trove)?

liach:
@Salil999 By the way, in order to update the rank of the elements in a priority queue, you need to make a custom implementation (java util priorityqueue cannot update values in log time)

src/main/java/net/glowstone/util/PathFinder.java
liach: (The correct comparator should compare the current distances from the shortest path tree)

liach:
And your implementation looks like dijkstra algorithm more, since you don't really have heuristics (It's like jumping two blocks, etc.)

By the way, vanilla uses some brute force that works like dijkstra in therory.

src/main/java/net/glowstone/util/PathFinder.java
Salil999: I think I'll need to store the Vectors into a wrapper object that contains information on both the Vector and the distance and create the Queue based on that class.

src/main/java/net/glowstone/util/PathFinder.java
Salil999: Thanks, did not know about that method!

src/main/java/net/glowstone/util/PathFinder.java
Salil999: Fixed - thanks!

src/main/java/net/glowstone/util/PathFinder.java
liach: Not really. If the distance is updated, the element cannot be updated in the priority queue

mastercoms:
@Salil999 Are you interested in continuing work on this PR?

Pr0methean:
@Salil999 If you want to run our style checks in IntelliJ, get the plugin at https://plugins.jetbrains.com/plugin/1065-checkstyle-idea. You'll have to configure it to use `etc/checkstyle.xml`.

src/main/java/net/glowstone/util/PathFinder.java
Pr0methean: If you move the initialization of `this.path` up to the declaration, then you can annotate the class `@RequiredArgsConstructor` and Lombok will generate the boilerplate constructor. If you do this, add Javadocs to the `startNode` and `endNode` fields.

Salil999:
@mastercoms Yes, but school has recently started for me (senior in college). I'll try to continue with it when I get time, but I'm OK if other people contribute as well (also follows the nature of OSS)

src/main/java/net/glowstone/util/PathFinder.java
XuZhen86: I have an idea of solving this. I have used this technique in programming competitions.
So we can't update the queue in realtime when the content of the node changes. How about we have a wrapper class for `Vector` that stores `Vector` and a `timeStamp`. When we need to update a `Vector`, we just create a new wrapper with new `Vector` and the newer `timeStamp` then add it to the queue. We also keep track of the newest `timeStamp` of all `Vector` in an array. In the `compareTo()` of the wrapper, we compare distance as usual. In the A* algorithm, whenever we take out a wrapper, we check if the `timeStamp` matches the `timeStamp` in the array. We discard wrappers with outdated `timeStamp`. Two must be equal for the wrapper to be accepted. 
You can see that we will have duplicated `Vector` in the queue, but each one of them will have different `timeStamp`. Only the one with the newest `timeStamp` will be accepted.
This solves the update problem without creating a custom priority queue. I made `timeStamp`an `int` start with 0. The A* keeps a master `timeStamp` and increases whenever an enqueue happens.

mastercoms:
Closed because of more active PR at #924

Issue #543, opened 2017-08-11
Add checksum to library manager


yangm97:
Glowstone build: bf69637286e428c15955887c8683d661696f3050

Sometimes, library files might get corrupt (either because of an unfinished download, bit rot, god hating you, etc). Currently the library manager doesn’t check the validity of these files, which can lead to unexpected behaviour. I suggest adding checksum verification to these files, but there could be cheaper ways of achieving the same results (like checking the file size). Or a combination of both: checksum after downloading, checking size on server startup, maybe adding a runtime flag to force checksum on startup for the paranoids around (o/).

aramperes:
On a technical perspective, it is possible to obtain the original SHA1 and MD5 hashes of a JAR directly from the repository (for example, https://repo.glowstone.net/service/local/repositories/central/content/org/apache/logging/log4j/log4j-core/2.8.2/log4j-core-2.8.2.jar.sha1)

mastercoms:
But that requires downloading a file :)

aramperes:
I guess we're going to have to hard-code the checksum too then.

Foorack:
I don't think it is necessary to hard-code the checksum. It is extremely unlikely for the SHA1 to match if either of the files is corrupted.

aramperes:
@Foorack Yes, but when it comes to file verification against threats like man-in-the-middle attacks, having the SHA downloaded from the internet defeats the purpose

Foorack:
@momothereal I was more thinking of file corruption. If you are worrying about MITM attack then it should rather enforce HTTPS at download.

mastercoms:
Fixed in https://github.com/GlowstoneMC/Glowstone/commit/825e5113f902fb5a3ebf1c6fbad3a32ad1f18411.

PR #802, opened 2018-01-05
Add checksum validation to library manager

aramperes:
This is my take to improve the library manager by adding checksum validation and allowing configuration of repository/lib directory.

#543 

I'm open to suggestions or rejection :)

### DIFF
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -197,6 +197,9 @@ public final class LibraryManager {
 
         /**
          * Computes and validates the checksum of a file.
+         * <p>If the file does not exist, the checksum will be automatically invalidated.
+         * <p>If the reference checksum or the algorithm are empty or null,
+         * the checksum will be automatically validated.
          *
          * @param file the file.
          * @param checksum the reference checksum to validate the file's digest with.
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -153,8 +153,6 @@ public final class LibraryManager {
                             GlowServer.logger.info("Downloaded " + library + ' ' + version + '.');
                         }
 
-                        Hashing.crc32c();
-
                         if (validateChecksum && algorithm != null && checksum != null
                                 && !checksum(file, checksum, algorithm)) {
                             GlowServer.logger.severe("The checksum for the library '" + getLibrary()
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -207,8 +207,13 @@ public final class LibraryManager {
     }
 
     public enum HashAlgorithm {
-
+        /**
+         * The SHA-1 hash algorithm.
+         */
         SHA1(Hashing.sha1(), "sha1"),
+        /**
+         * The MD5 hash algorithm.
+         */
         MD5(Hashing.md5(), "md5");
 
         /**
@@ -224,6 +229,12 @@ public final class LibraryManager {
 
         private static final Map<String, HashAlgorithm> BY_NAME = Maps.newHashMap();
 
+        /**
+         * Represents a hash algorithm.
+         *
+         * @param function the {@link HashFunction} used to calculate the hash
+         * @param name the name of the algorithm
+         */
         HashAlgorithm(HashFunction function, String name) {
             checkNotNull(function);
             checkNotNull(name);
@@ -245,7 +256,7 @@ public final class LibraryManager {
         }
 
         static {
-            // add the algorithm to the map
+            // add the algorithms to the map
             for (HashAlgorithm algorithm : values()) {
                 BY_NAME.put(algorithm.getName(), algorithm);
             }
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -185,7 +185,7 @@ public final class LibraryManager {
                     method.setAccessible(true);
                     method.invoke(ClassLoader.getSystemClassLoader(), file.toURI().toURL());
                 }
-            } catch (ReflectiveOperationException | IOException e) {
+            } catch (Exception e) {
                 GlowServer.logger.log(Level.WARNING,
                         "Failed to add to classpath: " + library + " " + version, e);
             }
--- a/src/main/java/net/glowstone/util/LibraryManager.java
+++ b/src/main/java/net/glowstone/util/LibraryManager.java
@@ -30,27 +30,6 @@ import net.glowstone.util.config.ServerConfig;
  */
 public final class LibraryManager {
     /**
-     * The default hash function used to compute the digests.
-     */
-    private static final HashFunction DEFAULT_HASH_FUNCTION = Hashing.sha1();
-    /**
-     * The alternative hash function used to compute the digests.
-     */
-    private static final HashFunction ALTERNATIVE_HASH_FUNCTION = Hashing.md5();
-    /**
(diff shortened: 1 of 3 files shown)


Pr0methean (commented):
Why not signature checking while we're at it?

aramperes:
@Pr0methean I am not familiar with Java's SSL API. Do you know if/how the signature is exposed in the `HttpsURLConnection` API? I'm not sure where to look.

mastercoms:
I feel like it would be better to have the checksums hard coded in, since this is being used for file download integrity.

src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -133,9 +195,81 @@ String getLibrary() {
...
+                GlowServer.logger.log(Level.SEVERE, "Failed to compute digest for '" + file.getName() + "'", ex);
+                return false;
+            }
+            return digest.equals(checksum);
+        }
+    }
+
+    public enum HashAlgorithm {
+
+        SHA1(DEFAULT_HASH_FUNCTION, DEFAULT_HASH_FUNCTION_NAME),
```
smartboyathome: One comment here, I don't think DEFAULT_HASH_FUNCTION_NAME and ALTERNATIVE_HASH_FUNCTION_NAME should be static variables, since they're already properties of the enum. In fact, I would argue that ALTERNATIVE_HASH_FUNCTION should be gotten rid of as a static variable as well, since it'll make for some awkward code once a third hash is added.

src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -52,13 +88,20 @@ public void run() {
         }
 
         downloaderService.execute(new LibraryDownloader(
-                "org.xerial", "sqlite-jdbc", "3.21.0", ""));
+                "org.xerial", "sqlite-jdbc", "3.21.0",
+                "347e4d1d3e1dff66d389354af8f0021e62344584", DEFAULT_HASH_ALGORITHM));
```
smartboyathome: I think using DEFAULT_HASH_ALGORITHM here actually makes the code a bit less clear. With that, I have to look up that the default hash algorithm is SHA1, and if we ever need to change that then all these hashes will have to change. If you  used HashAlgorithm.SHA1, it is immediately clear what type of hash this is and allows us to switch between hashes without changing all the entries at once.

src/main/java/net/glowstone/util/config/ServerConfig.java
```diff
@@ -412,6 +414,9 @@ private boolean migrate() {
         REGION_CACHE_SIZE("advanced.region-file.cache-size", 256),
         REGION_COMPRESSION("advanced.region-file.compression", true),
         PROFILE_LOOKUP_TIMEOUT("advanced.profile-lookup-timeout", 5),
+        LIBRARY_CHECKSUM_VALIDATION("advanced.library-checksum-validation", true),
+        LIBRARY_REPOSITORY_URL("advanced.library-repository-url",
+                "https://repo.glowstone.net/service/local/repositories/central/content/"),
```
smartboyathome, line 419: More of a question than "please change this", but do we want the default library url to really be our repo? Even though it proxies back to Maven Central, thats still using our bandwidth whenever anyone builds. This question becomes more important when custom libraries are introduced, but I wanted to call it out now.

smartboyathome (changes requested):

src/main/java/net/glowstone/util/config/ServerConfig.java
mastercoms: The reason why we use our own repo is because Minecraft Forge was asked to stop using Maven Central when they added library downloading in their installer, so we are trying to proactively avoid that issue by using our own repo.

src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -104,12 +143,29 @@ public void run() {
...
+                        file.delete();
+                        return;
+                    }
                 } catch (IOException e) {
                     GlowServer.logger.log(Level.WARNING,
                             "Failed to download: " + library + ' ' + version, e);
                     file.delete();
                     return;
                 }
+            } else if (validateChecksum && algorithm != null && checksum != null
```
mastercoms: Is this case for when someone manually puts their own lib in the folder?

src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -104,12 +143,29 @@ public void run() {
...
                         GlowServer.logger.info("Downloaded " + library + ' ' + version + '.');
                     }
+
+                    if (validateChecksum && algorithm != null && checksum != null
+                            && !checksum(file, checksum, algorithm)) {
+                        GlowServer.logger.severe("The checksum for the library '" + getLibrary()
+                                + "' does not match. "
+                                + "Restart the server to download it again.");
+                        file.delete();
+                        return;
```
mastercoms: Could we have a retry system that gives up after a second or third try?

src/main/java/net/glowstone/util/LibraryManager.java
```diff
@@ -123,19 +179,110 @@ public void run() {
...
+         */
         String getLibrary() {
             return library + '-' + version + ".jar";
         }
 
-        boolean checksum(File file, String checksum) {
-            // TODO: actually check checksum
-            return file.exists();
+        /**
+         * Computes and validates the checksum of a file.
```
mastercoms, line 222: Add newlines between each paragraph.

smartboyathome (approved):

src/main/java/net/glowstone/util/config/ServerConfig.java
smartboyathome: Ah, didn't know about that. Makes sense then.

src/main/java/net/glowstone/util/LibraryManager.java
aramperes: Yes, if they overwrite the library (or change its contents) for some reason.

src/main/java/net/glowstone/util/LibraryManager.java
aramperes: Done, looks like this:

```
13:52:20 [INFO] Downloading sqlite-jdbc 3.21.0...
13:52:20 [INFO] Downloading mysql-connector-java 5.1.44...
13:52:20 [INFO] Downloading log4j-api 2.8.1...
13:52:20 [INFO] Downloading log4j-core 2.8.1...
13:52:20 [INFO] Downloading commons-lang3 3.5...
13:52:21 [INFO] Downloaded log4j-api 2.8.1.
13:52:21 [INFO] Downloaded commons-lang3 3.5.
13:52:22 [INFO] Downloaded mysql-connector-java 5.1.44.
13:52:23 [INFO] Downloaded log4j-core 2.8.1.
13:52:27 [INFO] Downloaded sqlite-jdbc 3.21.0.
13:52:27 [SEVERE] The checksum for the library 'sqlite-jdbc-3.21.0.jar' does not match. Attempting download again (2/2)
13:52:27 [INFO] Downloading sqlite-jdbc 3.21.0...
13:52:32 [INFO] Downloaded sqlite-jdbc 3.21.0.
13:52:32 [SEVERE] The checksum for the library 'sqlite-jdbc-3.21.0.jar' does not match. Restart the server to download it again.
```

src/main/java/net/glowstone/util/LibraryManager.java
aramperes: Done

src/main/java/net/glowstone/util/LibraryManager.java
mastercoms: Perhaps to be more clear, specify that restarting the server only attempts to download it again and that server owners should check to see what the problem is or wait out connection issues before trying again.

mastercoms (approved):

PR #806, opened 2018-01-07
Update setup.sh

josephworks:
your old script was BAD...

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -8,4 +8,4 @@ echo verifying environment
 java -version
 mvn -version
 echo starting
-mvn -T 1C -B package clean install
+mvn -T 1C -B -X package clean install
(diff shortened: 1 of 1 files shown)


josephworks:
installs dependencies and runs commands to check environment before building a clean install package

mastercoms:
Hi thanks for your contribution, but this PR doesn't further the goals of the `setup.sh` script. System dependency management is going to be something that will be very platform specific so a script managing this must be much more robust to support a wider variety of systems. As such, the script cannot be accepted in this state, even in a new file.

Also, I feel like the specifications for such a script would be too complex to support many systems reliably, and thus it should be up to the user to find and install the dependencies on their machine, rather than having a script to do this for them.

PR #811, opened 2018-01-16
Fix an ArrayIndexOutOfBoundsException bug

Pr0methean:
Was caused by concurrent access to a non-thread-safe list. Using a ConcurrentLinkedDeque instead.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -142,7 +142,6 @@ import org.bukkit.advancement.Advancement;
 import org.bukkit.advancement.AdvancementProgress;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
-import org.bukkit.block.BlockState;
 import org.bukkit.boss.BossBar;
 import org.bukkit.configuration.serialization.DelegateDeserialization;
 import org.bukkit.conversations.Conversation;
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -140,6 +142,7 @@
 import org.bukkit.advancement.AdvancementProgress;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.block.BlockState;
```
aramperes: Unused import

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -901,39 +904,37 @@ protected void jump() {
...
-                        Map<BlockVector, BlockChangeMessage> map = chunks
-                                .computeIfAbsent(key, k -> new HashMap<>());
-                        map.put(new BlockVector(message.getX(), message.getY(), message
-                                .getZ()), message);
-                    }
-                }
+        // separate messages by chunk
+        // inner map is used to only send one entry for same coordinates
+        Map<Key, Map<BlockVector, BlockChangeMessage>> chunks = new HashMap<>();
+        while (true) {
```
aramperes, line 909: `while (!blockChanges.isEmpty())` ?

Removes the need for the null check afterwards.

aramperes (dismissed):

src/main/java/net/glowstone/entity/GlowPlayer.java
Pr0methean: That creates a race condition if multiple threads call processBlockChanges.

src/main/java/net/glowstone/entity/GlowPlayer.java
Pr0methean: Removed in 73dbfbe

mastercoms (approved):

aramperes (approved):

PR #817, opened 2018-01-19
API Update (1.12.2-R1.0), PlayerProfile API

aramperes:
This PR implements some new methods from upstream (Paper, Spigot) as well as Paper's `PlayerProfile` API. The `Villager.Career` class was added to upstream, so updates were necessary there too.

~~Build is currently failing because Glowkit is currently unable to be deployed (due to an authentication issue, to investigate...)~~


Pr0methean (approved):

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1659,6 +1660,23 @@ public boolean suggestPlayerNamesWhenNullTabCompletions() {
...
+    @Override
+    public GlowPlayerProfile createProfile(UUID id) {
+        checkNotNull(id);
+        return createProfile(id, null);
+    }
+
+    @Override
+    public GlowPlayerProfile createProfile(String name) {
+        checkNotNull(name);
+        return createProfile(null, name);
```
kaenganxt: A null uuid is filtered out in the [GlowPlayerProfile constructor](https://github.com/GlowstoneMC/Glowstone/pull/817/files#diff-bd59b1c2be78b091d2e887be7f414fe2R58).

src/main/java/net/glowstone/entity/meta/profile/GlowPlayerProfile.java
```diff
@@ -158,19 +162,91 @@ public CompoundTag toNbt() {
...
         }
         if (!propertiesTag.isEmpty()) { // Only add properties if not empty
             profileTag.putCompound("Properties", propertiesTag);
         }
         return profileTag;
     }
 
+    @Override
+    public String getName() {
+        return null;
```
kaenganxt: `return this.name;`

kaenganxt (changes requested):

mastercoms (approved):

src/main/java/net/glowstone/GlowServer.java
kaenganxt: The null preconditions here were fine. I was talking about a problem with the uuid. This method creates a `GlowPlayerProfile` with `uuid=null` and `name=name`, which is currently not possible because of the [precondition](https://github.com/GlowstoneMC/Glowstone/pull/817/files#diff-bd59b1c2be78b091d2e887be7f414fe2R60) in the GlowPlayerProfile constructor. So the `createProfile` method with just a name would never work.

The `GlowPlayerProfile` is currently not intended to work with a null uuid. It would throw a null pointer in the `toNbt()` method and I'm not sure if some other classes can handle it if `getUniqueId()` returns null.

kaenganxt (approved):

PR #821, opened 2018-01-21
[WIP] Annotations to replace constant entity properties

aramperes:
The goal of this PR is to replace constant entity properties (for example, `isUndead` and `isArthropod`) with annotations (`@EntityProperties`) instead of allowing these methods to be overriden.

I also added a `Sounds` annotation for hurt/death/ambient sounds instead of overriding the getters. The getters are not final for the moment because I only implemented it for certain entities.

I'm thinking of adding other properties like the bounding box dimensions (`setSize()`). Other technical annotations like entity metadata are planned but not part of the scope of this PR.

<details><summary>Example (click to expand)</summary>

For example, `GlowZombie.java`:
```java
public class GlowZombie extends GlowMonster implements Zombie {

    /* ... */

    @Override
    protected Sound getHurtSound() {
        return Sound.ENTITY_ZOMBIE_HURT;
    }

    @Override
    protected Sound getDeathSound() {
        return Sound.ENTITY_ZOMBIE_DEATH;
    }

    @Override
    protected Sound getAmbientSound() {
        return Sound.ENTITY_ZOMBIE_AMBIENT;
    }

    @Override
    public boolean isUndead() {
        return true;
    }
}
```

Becomes:

```java
@EntityProperties(undead = true)
@Sounds(hurt = Sound.ENTITY_ZOMBIE_HURT,
        death = Sound.ENTITY_ZOMBIE_DEATH,
        ambient = Sound.ENTITY_ZOMBIE_AMBIENT)
public class GlowZombie extends GlowMonster implements Zombie {

    /* ... */

}
```
</details>


src/main/java/net/glowstone/util/AnnotationUtil.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Gets the annotation instance for the given class.
+     *
+     * @param annotation the annotation class
+     * @param clazz      the target class
+     * @param <T>        the type of the annotation
+     * @return the annotation instance, or null if the annotation is not present
+     */
+    public static <T extends Annotation> T getClassAnnotation(Class<T> annotation,
```
kashike: This method isn't needed.
https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getAnnotation-java.lang.Class-

kashike (changes requested):

aramperes:
Closing this because other developers expressed their preference for a Builder pattern over Annotations. I'll keep the branch up for reference.

PR #823, opened 2018-01-25
Fix mob always drop exp when killed

XuZhen86:
This pull request tries to fix the bug that mob always drop exp when killed.

Observed behavior:
Mobs always drop exp when killed, even drown in water without player touching them.

Expected behavior:
> A mob will not drop experience unless it dies within five seconds (100 game ticks) of an attack registered as a player hit (including tamed wolves and TNT).
[Experience - Official Minecraft Wiki](https://minecraft.gamepedia.com/Experience)

Solution:
1. Add a countdown to track the 100 game ticks.
2. Enable countdown when a valid source hit the mob.
3. Set the killer when it kills the mob.
4. Add ```AnimalTamer ``` to ```GlowPlayer ```. Set the killer to the owner of the wolf when wolf kills the mob.
5. Mob only drops exp when player is in survival mode.

Possible issues:
1. The bow and arrow is still under development, the killer setting of arrow is not implemented.
2. It gives me ```java.lang.IllegalArgumentException: Compound does not contain ByteTag "CollarColor"``` whenever I place a wolf using wolf egg?
3. The check style seems to crashed it self, I had to disable it to continue the compile.


src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -92,6 +93,21 @@
...
+    @Getter
+    @Setter
+    private boolean isHitByPlayer;
+
+    /**
+     * Player hit countdown. 100 game ticks by default
+     */
+    @Getter
+    @Setter
+    private int hitByPlayerCountdown;
```
aramperes: Wouldn't it be simpler to have a field like `playerDamageTick` that is set to the current `ticksLived` when the entity is hit, then check if `ticksLived - playerDamageTick <= 100` when killed?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -806,6 +829,40 @@ public void damage(double amount, Entity source, DamageCause cause) {
         // apply damage
         amount = event.getFinalDamage();
         lastDamage = amount;
+
+        // If damaged directly by a player
+        // or damaged by a tnt ignited by a player in survival mode
+        // or damaged by a tamed wolf
+        if (
+                (source instanceof GlowPlayer
+                && ((GlowPlayer)source).getGameMode() == GameMode.SURVIVAL)
```
aramperes: Should adventure mode also be checked?

src/main/java/net/glowstone/entity/GlowTntPrimed.java
```diff
@@ -24,6 +24,10 @@
     private int fuseTicks;
     private Entity source;
 
+    @Getter
+    @Setter
+    private GlowPlayer player;
```
aramperes: Add Javadoc for this

aramperes (changes requested):

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -730,7 +739,8 @@ public void setHealth(double health) {
                 if (world.getGameRuleMap().getBoolean("doMobLoot")) {
                     LootData data = LootingManager.generate(this);
                     deathEvent.getDrops().addAll(data.getItems());
-                    if (data.getExperience() > 0) {
+                    // Only drop experience when hit by a player within 5 seconds (100 game ticks)
+                    if (ticksLived - playerDamageTick <= 100 && data.getExperience() > 0) {
```
aramperes, line 829: You'll probably want to have `playerDamageTick` default to -1 (in order to know if the entity wasn't hit yet), and add a check here for it being positive.

```java
if (playerDamageTick >= 0 && ticksLived - playerDamageTick <= 100 && data.getExperience() > 0)
```

src/main/java/net/glowstone/block/blocktype/BlockTnt.java
```diff
@@ -14,20 +14,33 @@
 public class BlockTnt extends BlockType {
 
     /**
-     * Convert a TNT block into a primed TNT entity.
+     * Convert a TNT block into a primed TNT entity with the player who ignited the tnt.
```
aramperes: `tnt` -> `TNT` (in the `@param` as well)

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -806,6 +816,15 @@ public void damage(double amount, Entity source, DamageCause cause) {
         // apply damage
         amount = event.getFinalDamage();
         lastDamage = amount;
+
+        if (isPlayerHit(source)) {
+            playerDamageTick = ticksLived;
+        }
+
+        if (health - amount <= 0) {
```
aramperes: Should this if-statement be inside of the `isPlayerHit` if-block above?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -830,6 +849,47 @@ public void damage(double amount, Entity source, DamageCause cause) {
         setLastDamager(source);
     }
 
+    /**
+     * Checks if the incoming source of damage is a "Player Hit".
+     * @param source The incoming source of damage
```
aramperes:
```java
/**
 * Checks if the source of damage was caused by a player.
 *
 * @param source The source of damage
 * @return true if the source of damage was caused by a player, false otherwise.
 */
```

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -830,6 +849,47 @@ public void damage(double amount, Entity source, DamageCause cause) {
...
+     */
+    private boolean isPlayerHit(Entity source) {
+        // If directly damaged by a player
+        if (source instanceof GlowPlayer) {
+            return true;
+        // If damaged by a TNT ignited by a player
+        } else if (source instanceof GlowTntPrimed) {
+            return
+                ((GlowTntPrimed)source).getPlayer() != null
```
aramperes: Casts should have a space after the closing parentheses. There's a couple to fix here.

Ex.: `((GlowTntPrimed) source).getPlayer() != null`

Also you can store the GlowTntPrimed instance before returning because you are casting it multiple times:

```java
GlowTntPrimed tnt = (GlowTntPrimed) source;
return tnt.getPlayer() != null [...]
```

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -830,6 +849,47 @@ public void damage(double amount, Entity source, DamageCause cause) {
...
+     * Determines the player who did the damage from source of damage.
+     * @param source The incoming source of damage
+     */
+    private Player determinePlayer(Entity source) {
+        // If been killed by an ignited tnt
+        if (source instanceof GlowTntPrimed) {
+            return (Player)((GlowTntPrimed)source).getPlayer();
+        }
+        // If been killed by a player
+        else if (source instanceof GlowPlayer) {
```
aramperes: Since you are returning inside the if-blocks, the `else`s are redundant and can be omitted.

src/main/java/net/glowstone/entity/GlowTntPrimed.java
```diff
@@ -24,6 +24,14 @@
     private int fuseTicks;
     private Entity source;
 
+    /**
+     * The player who ignited the TNT.
+     * null if not ignited by player.
+     */
+    @Getter
+    @Setter
+    private GlowPlayer player;
```
aramperes: Why are this variable and getter/setter necessary? The `getSource()` method should return the player who ignited the TNT (if any), according to docs.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -830,6 +849,47 @@ public void damage(double amount, Entity source, DamageCause cause) {
...
+    /**
+     * Checks if the incoming source of damage is a "Player Hit".
+     * @param source The incoming source of damage
+     */
+    private boolean isPlayerHit(Entity source) {
+        // If directly damaged by a player
+        if (source instanceof GlowPlayer) {
+            return true;
+        // If damaged by a TNT ignited by a player
+        } else if (source instanceof GlowTntPrimed) {
```
aramperes: The `else`s are redundant and can be omitted (return statement inside of the if-blocks).

aramperes (changes requested):

aramperes:
The exception caused by spawning a Wolf using an egg is now fixed in 12df4a7.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
XuZhen86: Strictly saying it doesn't matter, they both determine the source. But put it inside does make more sense.

Pr0methean: Put it inside. It'll save a few instructions when `!isPlayerHit(source)`

src/main/java/net/glowstone/entity/GlowLivingEntity.java
Pr0methean: Should we maybe use a VicariousKillerEntity or MeansOfKilling interface to simplify all of this?

src/main/java/net/glowstone/entity/GlowTntPrimed.java
Pr0methean: IOW:
```
public GlowPlayer getPlayer() {
    Entity source = getSource();
    return (source instanceof GlowPlayer) ? (GlowPlayer) source : null;
}
public void setPlayer(Player player) {
    source = player;
}
```

src/main/java/net/glowstone/entity/GlowTntPrimed.java
```diff
@@ -98,7 +106,11 @@ private void explode() {
 
     @Override
     public final Entity getSource() {
-        return source.isValid() ? source : null;
+        if (source == null || !source.isValid()) {
+            return null;
+        } else {
+            return source;
+        }
```
Pr0methean: `return (source != null && source.isValid()) ? source : null;`

src/main/java/net/glowstone/entity/GlowLivingEntity.java
XuZhen86: I don't think that's necessary. There are only several cases that require checking vicarious killing.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
XuZhen86: I made it ```private int playerDamageTick = -101;```, it solves the problem without extra code.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -830,6 +849,64 @@ public void damage(double amount, Entity source, DamageCause cause) {
...
+        if (source instanceof GlowWolf) {
+            return (Player)((GlowWolf) source).getOwner();
+        }
+
+        // All other cases
+        return null;
+    }
+
+
+
```
aramperes: Remove whitespace (keep only 1 line between methods)

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -195,7 +196,7 @@
  * @author Graham Edgecombe
  */
 @DelegateDeserialization(GlowOfflinePlayer.class)
-public class GlowPlayer extends GlowHumanEntity implements Player {
+public class GlowPlayer extends GlowHumanEntity implements Player, AnimalTamer {
```
aramperes: I think this is redundant because `GlowPlayer` extends `GlowHumanEntity`, which implements `HumanEntity`, which extends `AnimalTamer`. (alternatively, `GlowPlayer -> Player -> HumanEntity -> AnimalTamer`)

aramperes (changes requested):

src/main/java/net/glowstone/entity/GlowTntPrimed.java
```diff
@@ -22,8 +23,28 @@
...
+        Entity source = getSource();
+        return (source instanceof GlowPlayer) ? (GlowPlayer) source : null;
+    }
+
+    /**
+     * Set the player that ignited the TNT.
+     *
+     * @param player Player that ignited the TNT
+     */
+    public void setPlayer(Player player) {
```
aramperes: Can you remove this setter and just use `setSource` with the player? I don't think having a separate method here is necessary.

src/main/java/net/glowstone/block/blocktype/BlockTnt.java
```diff
@@ -7,28 +7,40 @@
...
-     * Convert a TNT block into a primed TNT entity.
+     * Convert a TNT block into a primed TNT entity with the player who ignited the TNT.
      *
      * @param tntBlock The block to ignite.
      * @param ignitedByExplosion True if another explosion caused this ignition.
+     * @param player The player who ignited the TNT.
      */
-    public static void igniteBlock(Block tntBlock, boolean ignitedByExplosion) {
+    public static void igniteBlock(
+        GlowBlock tntBlock, boolean ignitedByExplosion, GlowPlayer player) {
```
Pr0methean: Why do you have to narrow tntBlock from Block to GlowBlock? Things like that make unit testing more complicated.

XuZhen86: It's more consistent within the class. All other methods use Glowblock.

mastercoms: I agree with Pr0methean, it should be kept like `Block`, unless there is a specific reason for doing it otherwise. The other usages of `GlowBlock` as a method parameter are not reasons for this one to use it.

mastercoms (approved):

aramperes (approved):

Issue #547, opened 2017-08-13
Issue with AuthMe and proxy-support


yangm97:
Looking at the logs, both Bungeecord and Glowstone are receiving the correct player IP:

```
06:28:30 [INFO] [/187.180.212.96:39420] <-> InitialHandler has connected
06:28:30 [INFO] [TheticalFlyer97] <-> ServerConnector [lobby] has connected
```

```
06:28:30 [INFO] UUID of player TheticalFlyer97 is a6462498-08d5-37c4-bfbb-158cac0008f2
06:28:30 [INFO] TheticalFlyer97 [/187.180.212.96:43947] connected, UUID: a6462498-08d5-37c4-bfbb-158cac0008f2
06:28:30 [INFO] TheticalFlyer97 joined the game
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: FML
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: AuthMeBridge
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: FML|HS
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: FORGE
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: BTLP/Bridge
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: BungeeCord
06:28:30 [INFO] TheticalFlyer97[/187.180.212.96:43947] registered channel: BAuthMeBridge
06:28:42 [INFO] TheticalFlyer97 issued command: /register (censored)
```

But looking at the AuthMe database, you can see that AuthMe has, somehow, received the Bungeecord IP from Glowstone and is using that instead of the player IP:

![screen shot 2017-08-13 at 03 55 16](https://user-images.githubusercontent.com/4412002/29246869-60ee637c-7fdb-11e7-9d87-807093f5de8e.png)

Another issue with AuthMe is that it’s not censoring commands with passwords (like /register and /login). AuthMe works perfectly fine otherwise.

mastercoms:
It won't be able to censor commands unless it adds support for Glowstone. It uses a different logging interface than what Glowstone uses. The IP problem is strange, though. Also, you may want to censor your IP :).

yangm97:
My IP is dynamic, so it has probably changed itself a number of times already but thanks :)

[...]

PR #824, opened 2018-01-28
Fix implementation of Player.getVirtualHost()/getProtocolVersion()

stephan-gh:
Right now, `Player.getVirtualHost()` returns the client's IP address. However, it is actually supposed to return the hostname/port the player used to connect to the server.

Correct this and also make use of the version information in GlowSession instead of just returning the constant.

For reference: API was added in https://github.com/PaperMC/Paper/pull/914


mastercoms:
Does this fix #547?

aramperes (approved):

stephan-gh:
@mastercoms Not sure, it's not too likely that many public plugins are already using this API (considering it is Paper only) - so probably not.

Pr0methean (approved):

PR #826, opened 2018-01-28
Track Additional Movement Statistics

hugokerstens:
This PR adds the tracking of the movement statistics:
FALL_ONE_CM
MINECART_ONE_CM
BOAT_ONE_CM
AVIATE_ONE_CM (Elytra)
HORSE_ONE_CM
DIVE_ONE_CM
SWIM_ONE_CM
FLY_ONE_CM
Some of the mechanics for these statistics (for example minecart) are not yet implemented, and thus do not work yet, but I added them so that if they would be added in the feature, they would be tracked automatically.
I inspected the NMS code for how exactly these statistics are tracked, and thus they should be tracked the same as in vanilla.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -1479,6 +1479,16 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
         return false;
     }
 
+    @Override
+    public void setOnGround(boolean onGround) {
+        float fallDistance = getFallDistance();
+        if (onGround && fallDistance > 2.0F) {
+            int fallDistanceStatistic = (int) Math.round(fallDistance * 100.0D);
+            this.incrementStatistic(Statistic.FALL_ONE_CM, fallDistanceStatistic);
+        }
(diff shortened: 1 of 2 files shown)


aramperes:
Hello, thank you for your interest in contributing to Glowstone.

As per the project's [Contributor License Agreement](https://glowstone.net/cla) (s. 4), it is forbidden to submit work that originates from the study of Mojang source code (decompiled or otherwise). One of the primary objectives of this project is to be a clean-room rewrite of the Vanilla server software.

hugokerstens:
@momothereal Thank you for your quick response. I understand your reasoning, and it is clear I didn't read the CLA thoroughly. 

However, I still would like to implement these statistics. How would I do this know? Is it allowed if I just implement it how I think it should be implemented? Maybe this could still be biased since I saw the original logic, I don't know if that would comply with the CLA.

aramperes:
Your safest bet is probably to not work on that part if you are afraid of bias.

However, we prefer code quality and efficiency over 100% vanilla replication. If you're able to figure out a way to do it that seems logical and "close enough" to vanilla, while finding your own solution, go for it! :)

PR #828, opened 2018-01-29
Remove reflection from AgeableStore

Pr0methean:
AgeableStore and its subclasses will no longer need to obtain a Constructor<T> reflectively; instead, their own constructors will take an extra parameter that's a Function<Location,T>. Should improve performance while loading chunks that contain entities, especially at startup.


Pr0methean:
NB: This *does* build, but Circle-CI is giving me 401 errors, ever since I deleted my fork repo and created a new one.

kamcio96 (approved):

mastercoms (approved):

mastercoms:
NB fixed in https://github.com/GlowstoneMC/Glowstone/commit/546731dfd3c5533b6c028f536052be3f70df2749.

PR #832, opened 2018-01-30
Fix cactus dropping invalid item

XuZhen86:
This pull request mainly tries to fix a bug that the cactus will drop purple block when the internal data is not 0. It also fixes a typo and improves Javadoc.

Observed behavior:
Cactus drops purple block randomly when break, as seen in the screen shot below.
The purple blocks also groups together based on the data value.
This bug occurs more often when the cactus block is on top of a growing cactus.

Expected behavior:
Cactus should always drop cactus item when break.

Other known issues:
1. The check style seems to fail on codes related to arrow?
1. When breaking a 3-blocks tall cactus or sugar cane, the items will drop but disappears right after they touch the floor.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockCactus.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockCactus.java
@@ -128,16 +128,9 @@ public class BlockCactus extends BlockType {
         }
     }
 
-    /**
-     * Get the items that will be dropped by digging the block.
-     * Specially overriten for cactus to remove data from the dropped item.
-     *
-     * @param me The cactus block it self.
-     * @param tool The tool used or {@code null} if fists or no tool was used.
-     * @return The drops that should be returned.
-     */
     @Override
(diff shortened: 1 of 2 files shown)


XuZhen86:
The title gone wild lol
screen shot 2018-01-30 at 11 12 56 am

src/main/java/net/glowstone/block/blocktype/BlockCactus.java
```diff
@@ -123,4 +127,17 @@ private boolean canPlaceNear(Material type) {
                 return true;
         }
     }
+
+    /**
+     * Get the items that will be dropped by digging the block.
+     * Specially overriten for cactus to remove data from the dropped item.
```
mastercoms: Javadocs are not required for overridden methods, I think a comment saying `Overridden for cactus to remove data from the dropped item` within the method would be sufficient. Same for sugar cane.

mastercoms (approved):

mastercoms:
Merged. Thank you for your contribution to Glowstone!

PR #833, opened 2018-01-30
Fix dead bush drops nothing

XuZhen86:
This pull request tries to fix the bug that dead bush drops nothing when breaks.

Observed behavior:
Dead bush drops nothing when breaks.

Expected behavior:
> Dead bushes drop 0–2 sticks when broken without shears. When shears are used, a dead bush will drop itself. If the block below the dead bush is removed, the bush will simply disappear without dropping anything. [Dead Bush](https://minecraft.gamepedia.com/Dead_Bush)

Solution:
Rewrite the getDrops() in BlockDeadBush.
Returns nothing when ```tool == null```
Returns the dead bush when tool is a shear, otherwise return 0-2 sticks.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockDeadBush.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockDeadBush.java
@@ -4,6 +4,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.concurrent.ThreadLocalRandom;
+
+import net.glowstone.GlowServer;
 import net.glowstone.block.GlowBlock;
 import org.bukkit.Material;
 import org.bukkit.block.BlockFace;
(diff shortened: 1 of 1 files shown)


XuZhen86:
Should line 32 also check ```tool != null```? Because I see BlockOre does that:
```Java
ItemStack stack = new ItemStack(dropType, count, (short) data);
if (tool == null) {
    return Collections.unmodifiableList(Arrays.asList(stack));
}
Collection<ItemStack> drops = super.getDrops(block, tool);
```

mastercoms (approved):

mastercoms:
Thanks for your contribution to Glowstone!

mastercoms:
Sorry, I missed your question about a tool null check. Items should be checked if they are empty, using InventoryUtil.

Issue #830, opened 2018-01-29
Issue with infinity error logging after throwing ender pearl


XuZhen86:
I'm having an issue where the server console prints error crazily when I throw ender pearl.

Steps to reproduce:
1. ```mvn clean```
2. Build server ```2018.1.0-SNAPSHOT.c59b068``` with ```mvn package```
3. Login and throw ender pearl

Observed:
The throwing triggers the server console to print
```
13:56:27 [SEVERE] Error occurred while pulsing world world
java.lang.NullPointerException
        at net.glowstone.entity.projectile.GlowProjectile.pulsePhysics(GlowProjectile.java:62)
        at net.glowstone.entity.GlowEntity.pulse(GlowEntity.java:550)
        at net.glowstone.GlowWorld.pulse(GlowWorld.java:482)
        at net.glowstone.scheduler.WorldScheduler$WorldThread.run(WorldScheduler.java:170)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```
and it won't stop until I press ```^C``` to force stop the server.
The amount of ender pearl in player's hand will not decrease.
There is cool down indication.

PR #834, opened 2018-01-30
Entities with null bounding boxes no longer check collisions

Pr0methean:
E.g. an ender pearl won't collide with another ender pearl. Should improve performance and help prevent bugs similar to #830 in other projectile classes.

Also, ensures projectiles (even ones with a bounding box) never collide with themselves.


mastercoms (approved):

src/main/java/net/glowstone/entity/projectile/GlowProjectile.java
```diff
@@ -59,12 +63,14 @@ public GlowProjectile(Location location) {
 
     @Override
     protected void pulsePhysics() {
-        Vector size = boundingBox.getSize();
-        for (Entity entity : world.getNearbyEntities(
+        if (boundingBox == null) {
```
kaenganxt: `boundingBox != null` ;)

kaenganxt (dismissed):

src/main/java/net/glowstone/entity/projectile/GlowProjectile.java
Pr0methean: Fixed in 9a17f3d91450a8f9d7970a06dc0da22a1b36b96a.

PR #836, opened 2018-02-02
Update JUnit and PowerMock

Pr0methean:
This allows our unit tests to run on JDK 9, if the following JVM command-line options are added:

```
--add-opens java.base/java.lang=ALL-UNNAMED \
--add-opens java.base/java.lang.reflect=ALL-UNNAMED \
--add-opens java.base/java.util=ALL-UNNAMED \
--add-opens java.base/java.util.concurrent=ALL-UNNAMED \
--add-opens java.base/java.net=ALL-UNNAMED \
--add-opens java.logging/java.util.logging=ALL-UNNAMED
```

Although this PR uses the JUnit Vintage API, which is almost entirely compatible with the JUnit 4 API (hence only 2 test classes need code changes), it also adds a dependency on the JUnit Jupiter API for new and migrated tests. Both APIs can be used side-by-side. Migrating to Jupiter will probably eventually let us eliminate the `--add-opens`; but https://github.com/powermock/powermock/issues/830 means a full migration may not be possible right away.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.0.3</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-library</artifactId>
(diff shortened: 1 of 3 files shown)


Pr0methean:
NB: If Circle-CI times out immediately after this is merged, try it a second time; it probably has something to do with caches.

mastercoms (approved):

PR #838, opened 2018-02-03
Fix #837

### LINKED ISSUE #837: Attribute minimum value
`AttributeManager#setProperty(Key, double)` assumes that the minimum value for an attribute is `0`, but the acctual min. value for the `generic.luck` attribute is [`-1024`](https://minecraft.gamepedia.com/Attribute#Attributes_for_players).

Pr0methean:
Attributes can now have any minimum value, although the default is zero. `generic.luck` is added.

### DIFF
--- a/src/main/java/net/glowstone/entity/AttributeManager.java
+++ b/src/main/java/net/glowstone/entity/AttributeManager.java
@@ -13,6 +13,9 @@ import lombok.RequiredArgsConstructor;
 import net.glowstone.net.GlowSession;
 import net.glowstone.net.message.play.entity.EntityPropertyMessage;
 
+/**
+ * Manages the attributes described at https://minecraft.gamepedia.com/Attribute
+ */
 public class AttributeManager {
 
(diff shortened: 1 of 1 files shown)


aramperes:
Maybe the `Key#min` field should be added, using documentation from the wiki?

src/main/java/net/glowstone/entity/AttributeManager.java
```diff
@@ -132,8 +133,14 @@ public double getPropertyValue(Key key) {
         @Getter
         private final double def;
         @Getter
+        private final double min;
+        @Getter
         private final double max;
 
+        Key(String name, double def, double max) {
```
aramperes: Could you document either the fields or the parameters?

Pr0methean: Done in d9e3a0be852131a4ebdab4e8177301f871acf629.

aramperes (approved):

PR #842, opened 2018-02-04
Migrate non-PowerMock unit tests to Jupiter

Pr0methean:
Tests that use PowerMock will have to use the Vintage API until powermock/powermock#830 is fixed, but this PR converts all others to the Jupiter API (i.e. makes them real JUnit 5 tests, instead of JUnit 4 tests running on JUnit 5).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,8 @@
     <maven.compiler.target>1.8</maven.compiler.target>
     <kotlin.version>1.2.20</kotlin.version>
     <powermockVersion>2.0.0-beta.5</powermockVersion>
+    <junitVersion>5.0.3</junitVersion>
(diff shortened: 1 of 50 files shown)


kacperduras:
Is it better to gradually pass unit tests on a newer version of Junit? It has backward compatibility, so you do not have to do it at the same time only successively.

Pr0methean:
Yes, the rest will be done gradually. This PR covers only the simple cases.

Oh wait, are you asking about the opposite -- splitting up this PR to convert one test at a time? I've been told that doing refactors that way creates too many different PRs and spams people's notifications.

Pr0methean:
Not all tests are currently running in Circle-CI; I'm waiting to hear back from the JUnit Gitter how to fix this.

mastercoms (approved):

PR #829, opened 2018-01-29
Fishing (partial)

Pr0methean:
Update and merge the old fishing branch. Use the GlowProjectile class from #614 instead of creating a duplicate. Rename ItemFishingRode to ItemFishingRod.

This introduces the first tests for entity classes, and also an abstract `GlowEntityTest` with `testCreateSpawnMessage`, so it should eventually help create tests for other entities as well.

PR #849, opened 2018-02-14
Fix #847

### LINKED ISSUE #847: Boat item isn't consumed on the server-side.
The boat item (`ItemBoat`) isn't consumed when placing a boat.

The item appears to be consumed on the client-side, but by updating the inventory view (such as clicking the slot, leaving a vehicle, or simply re-logging) will show that the item is still there.

This is a duplication issue because breaking the boat will drop a new boat item, resulting in 2 boats in the player's inventory.

Reported by `jokbon` on Discord.

Pr0methean:


src/test/java/net/glowstone/block/itemtype/ItemBoatTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    void testRightClickBlock() {
+        boatItemHandler.rightClickBlock(
+                player, Mockito.mock(GlowBlock.class), BlockFace.DOWN, boatItemStack,
+                new Vector(0, 0, 0), EquipmentSlot.HAND);
+        verify(world, times(1)).spawn(any(Location.class), eq(Boat.class));
+        assertEmpty(inventory.getItemInMainHand());
+    }
+}
```
aramperes: Add newline at end of file

aramperes (dismissed):

src/main/java/net/glowstone/block/itemtype/ItemBoat.java
```diff
@@ -47,6 +48,7 @@ private void placeBoat(GlowPlayer player) {
             location.setYaw(player.getLocation().getYaw());
             Boat boat = targetBlock.getWorld().spawn(location, Boat.class);
             boat.setWoodType(woodType);
+            InventoryUtil.consumeHeldItem(player, holding);
```
aramperes, line 51: Don't you have to forcefully set the slot after changing the amount? IIRC it doesn't update the inventory until you use setSlot().

Pr0methean: Added a test case in `ItemProjectileTest` to confirm that it's written back to the GlowPlayerInventory even in the non-zero-remaining case. As for updating the client, that's handled the next time the player is pulsed, by `GlowPlayer` lines 863-5.

mastercoms (changes requested):
I tested it in-game, and the boat is not consumed.

Pr0methean:
I'll resume work on this once #829 is merged, since I'll likely need to extend `GlowEntityTest` from that PR.

mastercoms:
Closing due to https://github.com/GlowstoneMC/Glowstone/issues/847#issuecomment-370089545.

PR #851, opened 2018-02-17
Fix #848

### LINKED ISSUE #848: Entities fall through ice blocks
Gravity currently ignores ice, so entities will fall through it. Ice also appears to be ignored when generating the animals in the world generation as a solid block (entities will spawn inside of it).

KiuIras:
Hello folks!
This is my first PR, so let me know if I'm doing something wrong!

The PR should fix the #848 issue.
Checking for isSolid instead of isOccluding prevents items to fall trough "No occluding" items.
(I think that items are "occluding" when they block light)


Pr0methean (approved):

mastercoms:
Welcome and thank you for your contribution to Glowstone!

PR #822, opened 2018-01-23
Fix wrong ender pearl behaviors

XuZhen86:
This pull request tries to fix the bug of malfunctioning ender pearls.

Observed behavior:
1. No damage caused by using ender pearl in gamemode 0
2. Used ender pearls getting back to inventory
3. Player located inside a block after been teleported
4. Player’s orientation after teleporting resets to the orientation before teleporting

Expected behavior:
1. Player receive 5 fall damage after teleport
2. When player uses an ender pearl, the amount should decrease by one
3. Player located above the block
4. Player's instant orientation before teleporting should not be changed after teleporting

Solution:
1. Add damage after teleport
2. Add a wrapper method to decrease the ender pearl count if player is not in creative mode
3. Plus 1 to Y value of location before teleporting
4. Set pitch and yaw before teleporting

Another question: How can I pull request only on the current commit? For now, the commit history contains the previous fall damage fix. I tried to merge dev before commit but it didn't work.

PR #853, opened 2018-02-17
Fine tune ender pearl

XuZhen86:
This pull request tries to fine tune the details of the ender pearl.

Observed behavior:
1. Ender pearl trajectory is weird, can't travel very far.
2. No endermite spawn after using ender pearl
3. The 1 second cool down is broken
4. Ender pearl does not have a bonding box, causing it never collides with entities.

Expected behavior:
1. Ender pearl should have little friction when traveling in the air
2. Endermite has a 5% chance spawning
3. Player can only use ender pearl once per 1 second
4. Ender pearl does have a 0.25*0.25 bonding box. The data is found in [Entity](https://minecraft.gamepedia.com/Entity), "Types of entities" section, "Width * Height" column.

I previously submitted https://github.com/GlowstoneMC/Glowstone/pull/822, but later I found out the actual code merged is different from the one I submitted. The resulted bug is also fixed in this pull request.
https://github.com/GlowstoneMC/Glowstone/commit/9c3fce2b20a642eddf816b44437e345c5f3e4bbd did not fix the player's orientation problem because https://github.com/GlowstoneMC/Glowstone/blob/9c3fce2b20a642eddf816b44437e345c5f3e4bbd/src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java#L58
should be
```java
    entity.teleport(location, PlayerTeleportEvent.TeleportCause.ENDER_PEARL);
```

### DIFF
--- a/src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
+++ b/src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
@@ -1,8 +1,11 @@
 package net.glowstone.entity.projectile;
 
 import com.flowpowered.network.Message;
+import io.netty.util.internal.ThreadLocalRandom;
 import java.util.Arrays;
 import java.util.List;
+
+import net.glowstone.entity.monster.GlowEndermite;
 import net.glowstone.net.message.play.entity.EntityMetadataMessage;
 import net.glowstone.net.message.play.entity.EntityTeleportMessage;
 import net.glowstone.net.message.play.entity.EntityVelocityMessage;
@@ -12,6 +15,7 @@ import org.bukkit.block.Block;
 import org.bukkit.entity.EnderPearl;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.LivingEntity;
+import org.bukkit.event.entity.CreatureSpawnEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.player.PlayerTeleportEvent;
 import org.bukkit.projectiles.ProjectileSource;
@@ -27,7 +31,7 @@ public class GlowEnderPearl extends GlowProjectile implements EnderPearl {
      * @param location the position and facing of the thrower
      */
     public GlowEnderPearl(Location location) {
-        this(location, 2.0f);
+        this(location, 3.0f);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
```diff
@@ -1,8 +1,11 @@
 package net.glowstone.entity.projectile;
 
 import com.flowpowered.network.Message;
+import io.netty.util.internal.ThreadLocalRandom;
 import java.util.Arrays;
 import java.util.List;
+
```
aramperes: Remove the blank line here

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
```diff
@@ -40,33 +44,67 @@ public GlowEnderPearl(Location location, float speed) {
         super(location);
         setDrag(0.99, false);
         setDrag(0.99, true);
-        setHorizontalAirDrag(0.95);
-        setGravityAccel(new Vector(0,-0.06,0));
+        setHorizontalAirDrag(1);
+        setGravityAccel(new Vector(0, -0.03, 0));
```
aramperes: Can you put the Y acceleration (`-0.03`) in a named constant in the class? e.g. `VERTICAL_GRAVITY_ACCEL`

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
```diff
@@ -27,7 +31,7 @@
      * @param location the position and facing of the thrower
      */
     public GlowEnderPearl(Location location) {
-        this(location, 2.0f);
+        this(location, 3.0f);
```
aramperes, line 33: Where does the health value come from?

aramperes (dismissed):

src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java
XuZhen86: I think you mean https://github.com/Glowminers/Glowstone/blob/b36962cb2d0476a9bfadcd83437460eea94ddaf8/src/main/java/net/glowstone/entity/projectile/GlowEnderPearl.java#L103-L105.
When ender pearl collide with an entity, it counts as a punch from the thrower to the entity. For example, it makes wolf angry at the thrower and decrease reputation on villagers. But the entity takes no actual damage.

aramperes: Ah no, I was reading the `3.0f` as the health of the EnderPearl, because other entities generally pass their maxHealth in the constructor via `super(location, health)`. My mistake

mastercoms:
Looks great to me, nice work! Thank you for your contribution to Glowstone.

mastercoms (approved):

PR #855, opened 2018-02-22
Fix snow slab stacking

jtomlin12:
This pull request tries to fix the problem of a player being unable to continually stack snow. 

Observed:
When a player is stacking snow, they cannot stack another layer of snow if a block is filled.

Expected:
A player should be able to stack unlimited layers of snow on top of each other.


mastercoms (approved):

mastercoms:

PR #857, opened 2018-02-28
Fix Fire Ticks After Death

smcconke:
This pull request aims to fix the fire ticks persisting after a player has died and respawned.

**Observed Behavior:**  Once a player dies while on fire, they will respawn with the previous fire ticks and continue to take damage.

**Expected Behavior:**  Once the player dies while on fire, they will be cleansed of the amount of prior fire ticks.

This is my first pull request, even though it is rather simple, please let me know if there is any changes that can be made.  Thanks!


mastercoms:

PR #859, opened 2018-03-03
Directly depend on latest JAnsi release

### LINKED ISSUE #839: Cannot run on ARM device (Raspberry Pi 2), java.lang.UnsatisfiedLinkError
Title: Cannot launch glowstone.jar on raspberry pi 2 (raspbian stretch), java.lan.UnstatisfiedLinkError when loading jansi (apparently a dependency from Jline)

Glowstone build: glowstone-2018.1.0-20180201.145750-210 

Hello,
Lately, I have been trying to run Glowstone on a little raspberry pi 2, but I'm facing an error when launching Glowstone.jar (see log below).

The error apparently comes from jansi, which is apparently a dependency of Jline (which, as far as I can see, is used to pretify Glowstone's output). It fails to load its native library, apparently because jansi provides only x86 (IA 32) natives, which are not compatible with ARM ones. I already tried to install `libjansi-java` and `libjansi-java-native` but this will not solve any problem, as the libraries are bundled in the jars :/.
This issue was apparently already solved in jansi [here](https://github.com/fusesource/jansi/pull/65), but jline2 (or at least the version used here) does not take the latest version (1.12 instead of 1.16), which I think could solve the problem. 

Any ideas ?

Thanks in advance for your help :)

---
```Feb 03, 2018 10:21:13 AM net.glowstone.GlowServer main
SEVERE: Error during server startup.
java.lang.UnsatisfiedLinkError: Could not load library. Reasons: [no jansi32-2018.1.0-SNAPSHOT.c53014f in java.library.path, no jansi-2018.1.0-SNAPSHOT.c53014f in java.library.path, no jansi in java.library.path, /tmp/libjansi-32-2018-8859770119432880512.1: /tmp/libjansi-32-2018-8859770119432880512.1: cannot open shared object file: No such file or directory (Possible cause: can't load IA 32-bit .so on a ARM-bit platform)]
	at org.fusesource.hawtjni.runtime.Library.doLoad(Library.java:182)
	at org.fusesource.hawtjni.runtime.Library.load(Library.java:140)
	at org.fusesource.jansi.internal.CLibrary.<clinit>(CLibrary.java:42)
	at org.fusesource.jansi.AnsiConsole.wrapOutputStream(AnsiConsole.java:48)
	at org.fusesource.jansi.AnsiConsole.<clinit>(AnsiConsole.java:38)
	at net.glowstone.ConsoleManager.<init>(ConsoleManager.java:78)
	at net.glowstone.GlowServer.<init>(GlowServer.java:224)
	at net.glowstone.GlowServer.createFromArguments(GlowServer.java:477)
	at net.glowstone.GlowServer.main(GlowServer.java:437)
```

Pr0methean:
Fixes #839. Needs to be verified on an ARM computer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,11 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>org.fusesource.jansi</groupId>
+      <artifactId>jansi</artifactId>
+      <version>1.17</version>
+    </dependency>
+    <dependency>
       <groupId>jline</groupId>
       <artifactId>jline</artifactId>
       <version>2.14.5</version>


Pr0methean:
NB: At #839 I've asked @Cocosushi6 (reporter of the bug) to test this fix, since Coco is the only participant I know of who has an RPi.

aramperes:
This is tested and functional on a Pi 2B (debian 9).

aramperes (approved):
Works on my ~~machine~~Pi™

PR #860, opened 2018-03-08
Refactor out some reflective constructor calls

Pr0methean:
Should speed up:

- chunk loading with entities
- tree generation
- spawning entities
- writing NBT with list tags

Also removes unused imports globally.


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1405,48 +1413,54 @@ public double getHumidity(int x, int z) {
...
      * @return an instance of the spawned {@link Entity}
      * @throws IllegalArgumentException TODO: document the reason this can happen
      */
     public GlowEntity spawn(Location location, Class<? extends GlowEntity> clazz,
             SpawnReason reason) throws IllegalArgumentException {
-        GlowEntity entity = null;
 
         if (TNTPrimed.class.isAssignableFrom(clazz)) {
-            entity = new GlowTntPrimed(location, null);
+            return new GlowTntPrimed(location, null);
```
aramperes: This prevents the events from being called (under this statement) for Primed TNT.

src/main/java/net/glowstone/util/nbt/CompoundTag.java
```diff
@@ -471,23 +471,19 @@ public void putIntArray(String key, int... value) {
...
-            Constructor<? extends Tag> constructor = type.getConstructor();
-            List<Tag> result = new ArrayList<>(value.size());
-            for (V item : value) {
-                result.add(constructor.newInstance(item));
-            }
-            put(key, new ListTag<>(type, result));
-        } catch (ReflectiveOperationException e) {
-            throw new IllegalArgumentException("Unable to create list of type " + type, e);
+    public <V> void putList(String key, TagType type, List<V> value,
+            Function<? super V, ? extends Tag> tagCreator) {
```
aramperes, line 482: Why not store the method reference in the enum?

src/main/java/net/glowstone/GlowWorld.java
Pr0methean: Refactored in d78caf8ae29bfd9455a351a4f1dd267d2eb88b91 by overloading the GlowTntPrimed constructor so that it no longer needs separate logic.

src/main/java/net/glowstone/util/nbt/CompoundTag.java
Pr0methean: That would create type-safety issues, because an enum can't be generic, so I'm creating `put*List` wrappers instead.

src/main/java/net/glowstone/entity/EntityRegistry.java
```diff
@@ -159,107 +166,172 @@
...
-            .put(Wither.class, GlowWither.class)
-            .put(WitherSkeleton.class, GlowWitherSkeleton.class)
-            //TODO: Wither Skull
-            .put(Wolf.class, GlowWolf.class)
-            .put(Zombie.class, GlowZombie.class)
-            .put(ZombieHorse.class, GlowZombieHorse.class)
-            .put(ZombieVillager.class, GlowZombieVillager.class)
-            .build();
+            ENTITIES;
+    private static ImmutableBiMap.Builder<Class<? extends Entity>, Class<? extends GlowEntity>>
```
smartboyathome: It doesn't appear that we're actually taking advantage of the BiMap here, does it actually need to stay a BiMap? Also, since we already have a CustomEntityContext, I think we should create a BuiltinEntityDescriptor that encapsulates this and entityCtors. Doing this should allow us to move these builders into the static block rather than having them dangle off of the class.

src/main/java/net/glowstone/entity/EntityRegistry.java
```diff
@@ -159,107 +166,172 @@
...
+    /**
+     * Creates an entity of the given type at the given location.
+     *
+     * @param type     the entity type
+     * @param location the location to spawn the entity
+     * @return an entity that has not yet been sent to clients
+     */
+    public static GlowEntity constructEntity(Class<? extends Entity> type, Location location) {
+        if (ENTITY_CTORS.containsKey(type)) {
+            return ENTITY_CTORS.get(type).apply(location);
```
smartboyathome: Actually, I just realized from the below line that EntityStore is essentially a factory for entities, with its createEntity() method. We should just use that rather than duplicating that here, then custom entities will be handled the same as builtins.

smartboyathome (dismissed):

src/main/java/net/glowstone/entity/EntityRegistry.java
Pr0methean: Refactored; see https://github.com/Pr0methean/Glowstone/commit/6baa9be7f050dfb50a037da51f5f7613cb8083e7#diff-8769715641781a189890669ce58619a0R173

mastercoms (approved):

PR #863, opened 2018-03-11
Add more tests for entities

Pr0methean:
This PR creates a parallel class hierarchy of tests for GlowEntity descendants in `net.glowstone.entity` with no subpackage, and in `net.glowstone.entity.passive`. In general:

- For each class `E extends F` where `F extends GlowEntity`, there is a test class called `ETest` that extends `FTest`.
- When `E` is concrete, so is `ETest`.
- When `E` is abstract or effectively abstract, `ETest` is abstract and generic, and its declaration is `ETest<T extends E> extends FTest<T>`.
- `ETest` inherits `testCreateSpawnMessage` from `GlowEntityTest` (currently the only test method in that class).

I believe `GlowAgeableTest` and `GlowBatTest` are the only classes that add new non-stub test methods; this PR's main purpose is to apply existing and future test cases to more of the entity classes they are applicable to. However, this PR still increases the line coverage of `net.glowstone.entity.passive` to 36%.

After this PR is merged, similar test classes will be added to `net.glowstone.entity.monster`, `objects` and `projectile` (which are all the other subpackages that contain entity classes).


mastercoms (commented):
Looks good, but some of the files don't have new lines. Could you please fix that?

Pr0methean:
Should be fixed in e30588f.

aramperes (approved):

PR #865, opened 2018-03-11
Download runtime library dependencies, resolve conflicts against builtin libs.

smartboyathome:
This PR adds the ability for Glowstone to download dependencies for libraries specified in the glowstone.yml. It does so using an abstraction on top of Maven's Aether API, Naether, that makes dependency resolution fairly easy. I check the tree that is built up by this plugin against the libraries builtin to the server by searching the classpath for the dependency's POM, which maven-shade-plugin embeds into the JAR. I tried various methods (hence why this change took so long), and this seems to be the most stable way of attempting to do this.

In order to test this out, I downgraded Kotlin from being a builtin dependency to a library in the default glowstone.yml, since we don't actually use Kotlin in our own code. Watching the logs of server startup, you can see it download kotlin-runtime, kotlin-reflect, and kotlin-stdlib, the latter which is a dependency of kotlin-reflect.

Now, there's a few downsides to how I implemented this:
* The way I am checking for what dependencies Glowstone includes relies on an implementation detail from maven-shade-plugin. While not a large piece of code, it will have to be rewritten if we ever decide to move away from this plugin or from Maven in general. I tried exploring other possibilities, such as serializing the depndency tree at build time, but it proved to be a lot more work for not much more benefit.
* This requires embedding portions of Maven itself into Glowstone in order to perform the dependency resolution. While this doesn't tie us to Maven for builds like the dependency checking does, it still ties us to resolving dependencies in a certain way. If we decide to move build systems, it could be awkward to have two different ways of resolving dependencies depending on if they are runtime vs builtin. Currently, we're using Maven, so I think it's fine to use it for our library downloads as well.
* Naether provides the ability to download the dependencies itself, but I am explicitly disabling that since we have our own download manager. We may choose to use this at some point in the future, but I felt it was better to let this functionality bake before trying this out.

Let me know what you think of this change! I think the benefits of this outweigh the downsides listed above, but certainly there's room for improvement here.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -1050,7 +1050,7 @@ public final class GlowServer implements Server {
 
     private Naether createNaetherWithRepository(String repository, String libraryFolder) {
         Naether naether = new NaetherImpl();
-        naether.setLocalRepoPath(new File(libraryFolder).toPath().toAbsolutePath().toString());
+        naether.setLocalRepoPath(new File(libraryFolder).getAbsolutePath());
 
         try {
             // Must overwrite the collection here in order to remove Maven Central from it.
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -3,15 +3,22 @@ package net.glowstone;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.flowpowered.network.Message;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterators;
+import com.google.common.collect.Sets;
 import com.jogamp.opencl.CLDevice;
 import com.jogamp.opencl.CLPlatform;
+import com.tobedevoured.naether.NaetherException;
+import com.tobedevoured.naether.api.Naether;
+import com.tobedevoured.naether.impl.NaetherImpl;
+import com.tobedevoured.naether.util.RepoBuilder;
 import io.netty.channel.epoll.Epoll;
 import io.netty.channel.kqueue.KQueue;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.FileVisitor;
 import java.nio.file.Files;
@@ -28,6 +35,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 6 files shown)


src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1019,13 +1038,110 @@ private void loadConfig() {
...
+                        "/META-INF/maven/%s/%s/pom.xml",
+                        library.getGroupId(),
+                        library.getArtifactId()
+                )
+        ) != null;
+    }
+
+    private Naether createNaetherWithRepository(String repository, String libraryFolder) {
+        Naether naether = new NaetherImpl();
+        naether.setLocalRepoPath(new File(libraryFolder).toPath().toAbsolutePath().toString());
```
aramperes: `File#getAbsolutePath()` would be simpler

src/main/java/net/glowstone/util/config/ServerConfig.java
```diff
@@ -393,6 +393,12 @@ private boolean migrate() {
                     .add(new Library("org.apache.commons", "commons-lang3", "3.5",
                             HashAlgorithm.SHA1, "6c6c702c89bfff3cd9e80b04d668c5e190d588c6")
                             .toConfigMap())
+                    .add(new Library("org.jetbrains.kotlin", "kotlin-runtime", "1.2.21",
```
aramperes: Could you extract the Kotlin version to a constant? Makes sure the versions are kept in sync when updating.

smartboyathome: I can do that, but one thing to keep in mind is that updating the version in our code won't change the version in existing configs. Only new configs will inherit this version.

mastercoms: I find this to be a regression. Before, server owners were automatically updated to the latest lib version but now it must be a manual change.

smartboyathome: Mastercoms, by "this", do you meant all libraries that we moved into the config? Or are you talking specifically about moving Kotlin out of the POM and into the configs?

mastercoms: All the libraries. I think we should have two configuration sections: one for "builtin" libraries that Glowstone downloads by default which should only allow for a simple toggle to download them or not, and one for custom libraries that the server owner can specify.

smartboyathome: I would argue that if these libraries should be upgraded with a server upgrade, they should be embedded into the JAR. Otherwise, they are optional, only commonly required by plugins that the server operator may run, and should be under the server operator's control. With the "builtin" libraries section as you suggest, it'd be impossible for a server operator to upgrade or downgrade a library when a plugin depends on a specific version, especially with the SHAs now in place.

aramperes: How about we add a warning when a version of the downloaded library is lower than the one in the "builtin" library map?

mastercoms: These builtin libraries, excluding the Kotlin libs, are part of a CraftBukkit environment. We should be emulating that environment and thus we should have control over what that environment is. 

However, some server owners might not use plugins that require that CraftBukkit specific environment, so having a smaller install by downloading these optional libs at runtime is better. That's why they aren't shaded in the JAR.

Arguably, having a up to date and universal Kotlin environment controlled by us is also beneficial but I would be fine with just a warning about the Kotlin libs being out of date, as momo suggested.

smartboyathome:
In order to address the concerns in this review, I created Compatibility Bundles in #879. When that is implemented, I will rebase my changes and open a new PR.

Issue #866, opened 2018-03-16
Incompatibility with Java 8, when using CircleCI build


aramperes:
Using Java 8 update 111. I downloaded build `bd275e2` from CircleCI, where builds use **Java 9**.

I get this exception:
```
com.google.common.util.concurrent.ExecutionError: java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
        at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2215)
        at com.google.common.cache.LocalCache.get(LocalCache.java:4154)
        at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:4158)
        at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:5147)
        at com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache.java:5153)
        at net.glowstone.io.anvil.RegionFileCache.getRegionFile(RegionFileCache.java:89)
        at net.glowstone.io.anvil.AnvilChunkIoService.read(AnvilChunkIoService.java:55)
        at net.glowstone.chunk.ChunkManager.loadChunk(ChunkManager.java:151)
        at net.glowstone.chunk.GlowChunk.load(GlowChunk.java:210)
        at net.glowstone.chunk.GlowChunk.load(GlowChunk.java:205)
        at net.glowstone.chunk.GlowChunk.getHeight(GlowChunk.java:639)
        at net.glowstone.GlowWorld.getHighestBlockYAt(GlowWorld.java:1185)
        at net.glowstone.GlowWorld.getHighestBlockAt(GlowWorld.java:1190)
        at net.glowstone.generator.GlowChunkGenerator.canSpawn(GlowChunkGenerator.java:68)
Caused by: java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer;
        at net.glowstone.io.anvil.RegionFile.<init>(RegionFile.java:184)
        at net.glowstone.io.anvil.RegionFileCache$1.load(RegionFileCache.java:65)
        at net.glowstone.io.anvil.RegionFileCache$1.load(RegionFileCache.java:62)
        at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3716)
        at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2424)
[...]

PR #868, opened 2018-03-17
Switch to machine executor rather than Docker

Pr0methean:
May be the only way to stop OOM kills while still using CircleCI and properly supporting JDK8.

Relevant: #866

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ machine:
       _JAVA_OPTIONS: "-Xms1024m -Xmx1024m"
 defaults: &defaults
     machine:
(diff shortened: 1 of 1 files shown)


mastercoms:
Should `build-deploy` still be using docker?

Pr0methean:
No, because `build-deploy` has also been flaking with OOM kills (and would probably do so more often as we continued to increase test coverage). That said, I'd be in favor of a secondary build that used JDK9, and that would probably *have* to run on Docker if it was on CircleCI at all (all of CircleCI's `machine` images seem to be using JDK8).

mastercoms:
Then why is only `build` being switched from Docker to machine?

Pr0methean:
Wasn't aware of that; should be fixed now.

PR #870, opened 2018-03-17
Validate config entry type

### LINKED ISSUE #869: Crash on invalid config value type
Glowstone build: `2018.3.0-SNAPSHOT.98b00f3 (MC: 1.12.2)` 

After searching for a problem with Glowstone server starts, I've discovered a bug in our system, which sets all lines with only `monsters:` in it to `monsters: true/false`. Glowstone completely crashes on start because of this (see output below) and doesn't even log anything, which makes it hard to find the issue. I don't know if the server should start when the config is invalid (reading `ServerConfig.java` it seems that it should only warn and use the default value), but it should probably still log the error and maybe give an information about the invalid option to simplify the debugging here.

To reproduce:

Set a `glowstone.yml` config option to an invalid type (e.g. set `creatures.limit.monsters` to `true`)

---

Output:
```
22:37:28 [INFO] This server is running Glowstone version 2018.3.0-SNAPSHOT.98b00f3 (MC: 1.12.2) (Implementing API version 1.12.2-R2.0-SNAPSHOT)
22:37:29 [SEVERE] Error during server startup.
java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.Integer
	at java.util.function.Predicate.lambda$or$2(Predicate.java:101)
	at net.glowstone.util.config.ServerConfig.load(ServerConfig.java:269)
	at net.glowstone.GlowServer.loadConfig(GlowServer.java:980)
	at net.glowstone.GlowServer.(GlowServer.java:444)
	at net.glowstone.GlowServer.createFromArguments(GlowServer.java:494)
	at net.glowstone.GlowServer.main(GlowServer.java:454)
```

aramperes:
Adds a predicate for value class type.

Fixes #869


src/main/java/net/glowstone/util/config/ServerConfig.java
```diff
@@ -606,30 +657,32 @@ public boolean test(String value) {
...
          * Checks if the value is greater than (positive) or equal to zero.
          */
-        static final Predicate<Integer> ABSOLUTE = POSITIVE.or(ZERO);
+        static final Predicate<Number> ABSOLUTE = POSITIVE.or(ZERO);
         /**
          * Checks if the value is a valid port number.
          */
-        static final Predicate<Integer> PORT = POSITIVE.and((number) -> number < 49151);
+        static final Predicate<Integer> PORT = typeCheck(Integer.class)
+                .and(POSITIVE).and((number) -> number < 49151);
```
Pr0methean: Should be <= 49151: the first dynamic port is 49152.

src/main/java/net/glowstone/util/config/ServerConfig.java
```diff
@@ -650,7 +703,18 @@ public boolean test(String value) {
...
+
+        /**
+         * Creates a {@link Predicate} that checks if the value is an instance of the
+         * specified class.
+         *
+         * @param expected the expected class.
+         * @return the predicate
+         */
+        static <T> Predicate<T> typeCheck(Class<T> expected) {
+            return expected::isInstance;
```
Pr0methean, line 716: Inline this into lines 405-585, e.g. `DISABLE_GENERATION("world.disable-generation", false, Boolean::isInstance)`

aramperes: Wouldn't work in predicate chains, e.g. `Validators.typeCheck(Integer.class).and(Validators.ABSOLUTE)`

I'll replace it for single validators, i.e. `String.class::isInstance`

Pr0methean (approved):

PR #874, opened 2018-03-20
[i18n] Extract first few strings to resource bundle

Pr0methean:
This extracts the localizable strings in GlowServer, lines 1-887, into a resource bundle. The rest can be done in a similar manner. Should a naming convention be established for the string keys first?


aramperes:
For one, I think there should be a prefix distinction between console output and in-game messages/strings.

e.g. `console.info.something-something` and `game.something-else`

Log levels should be prefixed too if the message is strictly related to that level, e.g. `warning` vs. `info`.

Also, periods should be there to split prefixes/groups of similar messages, while dashes should be used for multiple words. For example: 
```
console.info.some-feature.enabled=Some feature is enabled.
console.info.some-feature.disabled=Some feature is disabled.
```

I think we should avoid the use of (conjugated) verbs and use more general terms, e.g. `console.warning.proxy-enabled-online-server` instead of `proxy.support.is.enabled.but.online.mode.is.enabled`

Pr0methean:
@momothereal How do these new message names look?

mastercoms (approved):

mastercoms:
@momothereal does it look good to you?

aramperes (approved):

PR #877, opened 2018-03-23
ServerProvider

Pr0methean:
This eliminates the need to statically mock Bukkit.getServer(), except for tests that touch Glowkit code. Thus, most entity tests (and one or two others) no longer use Powermock, and the CI build-deploy should finish in under 4 minutes again.


mastercoms (dismissed):
I would rather if there weren't static imports for getServer.

mastercoms (approved):

PR #880, opened 2018-03-26
Add support for PaperServerListPingEvent

stephan-gh:
PaperServerListPingEvent is an extended version of Bukkit's standard ServerListPingEvent, but allows full control over the status response. This makes it possible to change the player sample, modify the server version and many other things.

Until now, this was only possible by using ProtocolLib to modify the packets directly. With this event, it can be also done on Glowstone where ProtocolLib isn't working (yet).

### TODO before merging
Right now, the PR doesn't compile anymore as-is. The remaining changes (unrelated to the new event) need to be made by someone more familiar with the Glowstone implementation:

- [x] Needs an upstream merge in [Glowkit](https://github.com/GlowstoneMC/Glowkit), depends on https://github.com/GlowstoneMC/Glowkit/pull/18
- [x] Glowstone needs to be updated for new methods added in Bukkit/Spigot/Paper. There seem to be several new methods in the Paper API that need to be implemented.
- [x] Depends on https://github.com/GlowstoneMC/Glowstone/issues/881

### Test plugin
The new event is supported in the development builds of [ServerListPlus](https://www.spigotmc.org/resources/serverlistplus.241/).
Download `ServerListPlus-3.4.9-SNAPSHOT-Universal.jar` from [Jenkins](https://ci.minecrell.net/job/ServerListPlus/), all features should be working.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.destroystokyo.paper.Title;
+import com.destroystokyo.paper.profile.PlayerProfile;
 import com.flowpowered.network.Message;
 import com.flowpowered.network.util.ByteBufUtils;
 import com.google.common.base.Preconditions;
@@ -2341,6 +2342,16 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
     @Override
     public boolean hasResourcePack() {
         return resourcePackStatus == PlayerResourcePackStatusEvent.Status.SUCCESSFULLY_LOADED;
+    }
+
+    @Override
+    public PlayerProfile getPlayerProfile() {
+        return getProfile();
+    }
+
+    @Override
+    public void setPlayerProfile(PlayerProfile playerProfile) {
+        throw new UnsupportedOperationException("Not implemented yet.");
     }
 
     @Override
--- a/src/main/java/net/glowstone/entity/meta/profile/GlowPlayerProfile.java
+++ b/src/main/java/net/glowstone/entity/meta/profile/GlowPlayerProfile.java
@@ -15,6 +15,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.logging.Level;
+import javax.annotation.Nullable;
 import lombok.Getter;
 import net.glowstone.GlowServer;
 import net.glowstone.ServerProvider;
@@ -211,8 +212,18 @@ public class GlowPlayerProfile implements PlayerProfile {
     }
 
     @Override
+    public String setName(@Nullable String name) {
+        throw new UnsupportedOperationException("Not implemented yet.");
+    }
+
+    @Override
     public UUID getId() {
         return uniqueId == null ? null : uniqueId.getNow(null);
+    }
+
+    @Override
+    public UUID setId(@Nullable UUID uuid) {
+        throw new UnsupportedOperationException("Not implemented yet.");
     }
 
     /**
@@ -228,6 +239,11 @@ public class GlowPlayerProfile implements PlayerProfile {
     @Override
     public Set<ProfileProperty> getProperties() {
         return Sets.newHashSet(this.properties.values());
+    }
+
+    @Override
+    public boolean hasProperty(String property) {
+        throw new UnsupportedOperationException("Not implemented yet.");
     }
 
     @Override
@@ -284,6 +300,11 @@ public class GlowPlayerProfile implements PlayerProfile {
         return name != null && getId() != null && properties.containsKey("textures");
     }
 
+    @Override
+    public boolean completeFromCache() {
+        return completeCached();
+    }
+
     /**
      * Looks up the UUID if it's missing and hasn't already been attempted, and waits for it.
      *
@@ -293,6 +314,11 @@ public class GlowPlayerProfile implements PlayerProfile {
         completeAsync();
         uniqueId.join();
         return isComplete();
+    }
+
+    @Override
+    public boolean complete(boolean textures) {
+        throw new UnsupportedOperationException("Not implemented yet.");
     }
 
     /**
--- a/src/main/java/net/glowstone/net/GlowStatusClient.java
+++ b/src/main/java/net/glowstone/net/GlowStatusClient.java
@@ -1,0 +1,28 @@
+package net.glowstone.net;
+
+import com.destroystokyo.paper.network.StatusClient;
+import java.net.InetSocketAddress;
+import javax.annotation.Nullable;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class GlowStatusClient implements StatusClient {
+
+    private final GlowSession session;
+
+    @Override
+    public InetSocketAddress getAddress() {
+        return session.getAddress();
+    }
+
+    @Override
+    public int getProtocolVersion() {
+        return session.getVersion();
+    }
+
+    @Nullable
+    @Override
+    public InetSocketAddress getVirtualHost() {
+        return session.getVirtualHost();
+    }
+}
--- a/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
+++ b/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
@@ -37,7 +37,6 @@ public final class StatusRequestHandler implements
         int sampleCount = server.getPlayerSampleCount();
         if (players.size() <= sampleCount) {
             sampleCount = players.size();
-            Collections.shuffle(players, random);
         } else {
             // Avoid shuffling the entire list if we only need a few players
             for (int i = 0; i < sampleCount; i++) {
--- a/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
+++ b/src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
@@ -6,7 +6,7 @@ import com.destroystokyo.paper.profile.PlayerProfile;
 import com.flowpowered.network.MessageHandler;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
@@ -33,7 +33,7 @@ public final class StatusRequestHandler implements
     private static void choosePlayerSample(GlowServer server, PaperServerListPingEvent event) {
         ThreadLocalRandom random = ThreadLocalRandom.current();
 
-        List<Player> players = new ArrayList<>(server.getOnlinePlayers());
+        List<Player> players = Arrays.asList(server.getOnlinePlayers());
         int sampleCount = server.getPlayerSampleCount();
         if (players.size() <= sampleCount) {
             sampleCount = players.size();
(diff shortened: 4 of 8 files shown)


src/main/java/net/glowstone/net/GlowStatusClient.java
```diff
@@ -0,0 +1,29 @@
...
+    public int getProtocolVersion() {
+        return session.getVersion();
+    }
+
+    @Nullable
+    @Override
+    public InetSocketAddress getVirtualHost() {
+        return session.getVirtualHost();
+    }
+
```
momothereal: Remove whitespace

aramperes: Remove whitespace

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
```diff
@@ -1,79 +1,95 @@
...
 import org.bukkit.entity.Player;
-import org.bukkit.event.server.ServerListPingEvent;
 import org.bukkit.util.CachedServerIcon;
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 
 public final class StatusRequestHandler implements
     MessageHandler<GlowSession, StatusRequestMessage> {
 
+    private static final UUID FAKE_UUID = new UUID(0, 0);
```
aramperes: Change to `BLANK_UUID`

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
```diff
@@ -1,79 +1,95 @@
...
-            server.getMaxPlayers());
-        event.players = new ArrayList<>(server.getOnlinePlayers());
-        event.icon = server.getServerIcon();
         event.serverType = server.getServerType();
         event.clientModsAllowed = server.getAllowClientMods();
+
+        // Add player sample list
+        List<Player> playerList = new ArrayList<>(server.getOnlinePlayers());
+        int count = Math.min(playerList.size(), server.getPlayerSampleCount());
+        Collections.shuffle(playerList);
```
aramperes: Is shuffling really necessary? I understand if that's Vanilla behavior, but it seems like to me like an unnecessary performance hit (especially with a large player count)

aramperes (dismissed):

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
stephan-gh: The main purpose here is to avoid showing the same players all the time. This could be simplified more, but then it would be less random. Up to you.

mastercoms:
I'll try to update Glowkit soon, but I would appreciate if someone else could do it as I'll be pretty busy for the next week or so.

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
Pr0methean: Implement a shuffle that stops choosing players when you have count.

Pr0methean (changes requested):

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
stephan-gh: What does "when you have count" mean?

Pr0methean: I mean, if `server.getPlayerSampleCount() < playerList.size()`, then only choose (and only copy) `server.getPlayerSampleCount()` random elements and then stop building the shuffled copy. That way, the running time, and the memory the shuffled copy consumes, will stop increasing linearly with `playerList.size()` once they reach the sample size.

stephan-gh:
I've optimized the code a bit to avoid shuffling the entire list if only a few players are needed. However, the copy of the entire player list is still needed because we need to swap entries and access them by their index.

smartboyathome:
Unfortunately, the algorithm in your most recent change seems to bias against the beginning of the list, since the swap happens to the first k members, where k is the number of players you are selecting. In addition, if the number of players is less than k, you don't need to randomize them (since the whole point of the randomization is to try to return a different subset).

You should look into using something like the Reservoir Sampling here. This algorithm should allow you to randomly select items from the list without using any form of shuffle. Instead, it weights probabilities over the iteration of the players list, running in O(n) time (where n is the number of players). It can be done without an array copy due to this.

stephan-gh:
> Unfortunately, the algorithm in your most recent change seems to bias against the beginning of the list, since the swap happens to the first k members, where k is the number of players you are selecting.

How does the swap to the first k members bias it to the beginning of the list?

The algorithm moves the randomly chosen entries to the beginning of the list, and chooses further entries only from the remaining list to avoid duplicates. The entire list is considered when choosing entries.

Essentially, it is a partial iteration of the [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle), stopping when we have chosen enough entries.

> In addition, if the number of players is less than k, you don't need to randomize them (since the whole point of the randomization is to try to return a different subset).

Technically not, but IMO it would be a bit weird if the players are randomly sorted when `players.size() > sampleCount`  but not if you have less players online. The overhead in that case is probably rather low anyway.

> You should look into using something like the Reservoir Sampling here. This algorithm should allow you to randomly select items from the list without using any form of shuffle. Instead, it weights probabilities over the iteration of the players list, running in O(n) time (where n is the number of players). It can be done without an array copy due to this.

I'm not too concerned about the copy of the players list personally. It won't be _that_ large usually. If you'd like to change it, feel free to add more commits here or make a PR on my fork.

Pr0methean (approved):
Please make sure CircleCI passes before submitting! It looks like we're still using a version of Paper that doesn't define PaperServerListPingEvent; you'll probably need to update the version in pom.xml.

Pr0methean:
@smartboyathome Minecrell is right: his algorithm is unbiased, and its running time is linear in the *sample* size.

@Minecrell Since `GlowServer.getOnlinePlayers` already makes a defensive copy, we can just use `Arrays.asList`.

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
```diff
@@ -1,79 +1,110 @@
...
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 
 public final class StatusRequestHandler implements
     MessageHandler<GlowSession, StatusRequestMessage> {
 
+    private static final UUID BLANK_UUID = new UUID(0, 0);
+
+    private static void choosePlayerSample(GlowServer server, PaperServerListPingEvent event) {
+        ThreadLocalRandom random = ThreadLocalRandom.current();
```
Pr0methean, line 34: Check shouldHidePlayers() first?

Pr0methean (changes requested):

stephan-gh:
> Please make sure CircleCI passes before submitting! It looks like we're still using a version of Paper that doesn't define PaperServerListPingEvent; you'll probably need to update the version in pom.xml.

Yeah, this is mentioned in the PR description. Updating Glowkit needs to be done separately before this PR can be merged. The reason I haven't included that in this PR is that there were many new methods added unrelated to the event implementation (e.g. in the `PlayerProfile` API). To test my changes locally, I have only added the stub methods to get everything to compile.

If it helps I could submit a PR for [Glowkit](https://github.com/GlowstoneMC/Glowkit) to merge the Glowkit patches on top of the current Paper API. However, unless you want to merge the stub methods someone else with more experience with the Glowstone implementation will need to implement them.

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
stephan-gh: This would be nice, but this method is run _before_ the event is called so it would be never set at this point. The only way to optimize this would be to override `getPlayerSample()` on the event and lazily initialize the sample list, however `shouldHidePlayers()` will likely not be used often so this isn't quite worth optimizing.

src/main/java/net/glowstone/net/handler/status/StatusRequestHandler.java
```diff
@@ -1,79 +1,110 @@
...
+
+        List<Player> players = Arrays.asList(server.getOnlinePlayers());
```
stephan-gh: In newer Bukkit versions (since 1.7.10 I think), `Server.getOnlinePlayers()` no longer returns an array but rather a _view_ of the online players. Therefore, this will cause a compile error.

Pr0methean:
The Glowkit PR sounds fine to me. Could you please put all the stubs in a separate commit, and once they're merged, open an issue linking to the commit?

stephan-gh:
@Pr0methean Done. CircleCI should compile and might pass the tests now once https://github.com/GlowstoneMC/Glowkit/pull/18 is merged (otherwise it can't find the updated Glowkit artifact). For some reasons some of the tests are failing with unrelated NPEs locally (even on the main `dev` branch) so I'm not entirely sure.

stephan-gh:
@Pr0methean Build seems to pass now, yay! :)

src/main/java/net/glowstone/entity/meta/profile/GlowPlayerProfile.java
```diff
@@ -210,11 +211,21 @@ public CompoundTag toNbt() {
...
+        throw new UnsupportedOperationException("Not implemented yet.");
+    }
+
     @Override
     public UUID getId() {
         return uniqueId == null ? null : uniqueId.getNow(null);
     }
 
+    @Override
+    public UUID setId(@Nullable UUID uuid) {
```
yannicklamprecht, line 225: I wouldn’t recommend setting the UUID after GameProfile creation. That could lead to unpleasant behavior. 

Player#getGameProfile#setId could be called at any time. Blame me if I’m wrong.

yannicklamprecht: The gameprofile is like an identity card. Setting the UUID is comparable to changing the ID of  your IDC after creation.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -2343,6 +2344,16 @@ public boolean hasResourcePack() {
...
     }
 
+    @Override
+    public PlayerProfile getPlayerProfile() {
+        return getProfile();
+    }
+
+    @Override
+    public void setPlayerProfile(PlayerProfile playerProfile) {
+        throw new UnsupportedOperationException("Not implemented yet.");
```
yannicklamprecht, line 2354: This method should ensure that only GameProfiles with the same UUID can be set, Else throw an exception.
Modification of the players UUID shouldn’t be allowed after creation.

src/main/java/net/glowstone/entity/meta/profile/GlowPlayerProfile.java
stephan-gh: This is a method added upstream by Paper: https://ci.destroystokyo.com/userContent/apidocs/com/destroystokyo/paper/profile/PlayerProfile.html#setId-java.util.UUID-

If I remember correctly, the Paper implementation usually returns copies to avoid this problem. It doesn't matter yet because this is just a stub, and won't be implemented as part of this PR. (I've only added it so the code compiles correctly)

src/main/java/net/glowstone/entity/GlowPlayer.java
stephan-gh: Again unrelated to this PR (since this is just a stub), but it is in fact designed to make it easy to change a player's identity. It is specifically designed to make it possible to change the identity of the player, or something like that. (Not entirely sure, see https://ci.destroystokyo.com/userContent/apidocs/org/bukkit/entity/Player.html#setPlayerProfile-com.destroystokyo.paper.profile.PlayerProfile-)

aramperes (approved):

PR #883, opened 2018-03-26
[i18n] Remaining GlowServer messages

Pr0methean:


src/main/java/net/glowstone/GlowServer.java
```diff
@@ -397,7 +397,7 @@
     private WorldStorageProviderFactory storageProviderFactory = null;
     /**
      * Whether the server should just generate and load configuration files, then exit.
-     *
+     *
```
aramperes: 👀

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1306,7 +1324,7 @@ public void addAdvancement(Advancement advancement) {
     /**
      * Creates an {@link AdvancementsMessage} containing a list of advancements the server has,
      * along with some extra actions.
-     *
+     *
```
aramperes: 👀

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -1323,7 +1341,7 @@ public AdvancementsMessage createAdvancementsMessage(boolean clear, List<Namespa
     /**
      * Creates an {@link AdvancementsMessage} containing a given list of advancements, along with
      * some extra actions.
-     *
+     *
```
aramperes: 👀

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -2511,18 +2536,18 @@ public String getConsoleLogDateFormat() {
...
      * Gets the server type.
-     *
+     * 
      * <p>Currently, this value is set to {@code VANILLA}.
      *
      * @return the server type.
      */
     public String getServerType() {
-        return "VANILLA";
+        return strings.getString("glowstone.server-type.vanilla");
```
aramperes: I don't think this should be mutable

aramperes (dismissed):

src/main/java/net/glowstone/GlowServer.java
mastercoms: It's used in packets, so yes, it should not be localized.

Pr0methean: Reverted.

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -2537,7 +2537,7 @@ public String getConsoleLogDateFormat() {
     /**
      * Gets the server type.
      *
-     * <p>Currently, this value is set to {@code VANILLA}.
+     * <p>Currently, this value is set to {@code VANILLA} and is not localized.
```
mastercoms: I don't like this doc addition. It's like specifying an enum's members are not localized, since this is essentially what the server type string is. It can be `BUKKIT` for Bukkit+Forge modded servers, `VANILLA` for non-Forge servers or `FML` for Forge servers. If you want, you can add that to the documentation.

Pr0methean: Rewritten.

Pr0methean: Note that StatusRequestHandler includes the server type in its output, in a form that AFAICT is likely to be displayed to players, so the explanation isn't as simple as for an enum value's name.

mastercoms: Yes, the client takes the server type and then displays a name according to that type. For example, `BUKKIT` -> `Bukkit modded server` IIRC.

Pr0methean: So it can still be localized client-side? Rewritten accordingly.

mastercoms (approved):

src/main/java/net/glowstone/GlowServer.java
Pr0methean: And I assume that localizing to client-side "Bukkit modded server" implies a modded client? I don't remember Mojang officially acknowledging the existence of Bukkit.

mastercoms: Yes, this is not part of the standard protocol, it's for Forge clients.

PR #888, opened 2018-03-28
Living entities now take fire damage

smcconke:
This pull request aims to implement fire damage for living entities:

**Observed Behavior:**  Entities would not catch fire when standing in a fire block.  Since entities were not being detected in the fire block, fire ticks were not updated and the entity takes zero damage.

**Expected Behavior:**  Entities will catch fire when standing in fire.  While standing in fire, the entity will take 1 tick of damage, equivalence of half a heart in-game.  Once the entity is no longer standing in fire, they will receive 8 ticks of damage, equivalence of four hearts in-game.  If the entity were to enter the fire again, they will continue to take damage and add an additional 8 ticks of fire damage once they exit.

This was implemented in a similar manner as lava damage, checking if the entity is standing in fire and dealing damage to the entity.  We must check if the entity is dead or not, that way if they were to die in the fire they do not receive the fire ticks after death.  Once exiting the fire if they were previously standing in fire, then their fire ticks will increment by 180.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -244,6 +244,10 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
      * If this entity has stood in fire (for fire application).
      */
     private boolean stoodInFire;
+    /**
+     * The ticks an entity stands adjacent to fire and lava.
+     */
+    private int adjacentBurnTicks;
 
     /**
      * Creates a mob within the specified world.
@@ -313,19 +317,12 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
             damage(1, DamageCause.SUFFOCATION);
         }
 
-        // fire damage
+        // fire and lava damage
         if (getLocation().getBlock().getType() == Material.FIRE) {
             damage(1, DamageCause.FIRE);
             // not applying additional fire ticks after dying in fire
             stoodInFire = !isDead();
-        } else {
-            if (stoodInFire) {
-                setFireTicks(getFireTicks() + 180);
-                stoodInFire = false;
-            }
-        }
-
-        if (getLocation().getBlock().getType() == Material.LAVA
+        } else if (getLocation().getBlock().getType() == Material.LAVA 
                 || getLocation().getBlock().getType() == Material.STATIONARY_LAVA) {
             damage(4, DamageCause.LAVA);
             if (swamInLava) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -309,6 +313,18 @@ public void pulse() {
             damage(1, DamageCause.SUFFOCATION);
         }
 
+        // fire damage
+        if (getLocation().getBlock().getType() == Material.FIRE) {
+            damage(1, DamageCause.FIRE);
```
aramperes: The `pulse` method is executed **once per tick**, which is every 50ms. ~~This check would damage the entity 20 times per second if they step in fire.~~ Also, the `setFireTicks` call below would increment the fire duration by 180 ticks every tick it isn't on fire.


edit: actually the invincibility timer takes care of the damaging, but the fireTicks counter issue would still be problematic

aramperes (changes requested):

src/main/java/net/glowstone/entity/GlowLivingEntity.java
smcconke: The fireTicks is only incremented when the entity has just recently exited the fire block, so by assigning False to stoodInFire we won't have a fireTicks counter incrementing by 180 with every tick.  Sorry about the confusion, I tested this on a live server where the damage in the fire and out of the fire proved successful.  Please let me know if I should make any changes.

aramperes: Hey, it's my bad for the confusion. After putting second thought to this, the logic does make sense 😊

Although you can simplify the `} else { if (stoodInFire) {` branches to an else-if statement.

Pr0methean:
Shouldn't we be searching at least a 3x2x3 neighborhood for fire? I'm pretty sure standing on a block that's *next* to a fire block will ignite an entity in vanilla.

aramperes:
@Pr0methean I don't think so about standing next to a block, but it should take the bounding-box of the entity into consideration.

smcconke:
I was thinking this as well.  I was basing this off of the lava implementation and noticed you will not take damage or catch fire if you are adjacent to the block.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
smcconke: Thank you for the feedback though!  It seems I will have to do some more work in this implementation.

smcconke:
Hi All,

My most recent commit took into consideration the bounding-box of the entity for fire and lava.  I tried to refactor the code in order to incorporate the lava damage with the fire.  Please let me know if I can make any changes.

Thanks.

aramperes (approved):

aramperes:
Works like a charm, thank you for your contribution (and apologies for the delay)

PR #889, opened 2018-03-28
Download runtime library dependencies, resolve conflicts.

smartboyathome:
Adapting the changes of #865 after #879 was merged in.

This PR adds the ability for Glowstone to download dependencies for libraries specified in the glowstone.yml. It does so using an abstraction on top of Maven's Aether API, Naether, that makes dependency resolution fairly easy. I check the tree that is built up by this plugin against the libraries builtin to the server by searching the classpath for the dependency's POM, which maven-shade-plugin embeds into the JAR. I tried various methods (hence why this change took so long), and this seems to be the most stable way of attempting to do this.

Now, there's a few downsides to how I implemented this:
* The way I am checking for what dependencies Glowstone includes relies on an implementation detail from maven-shade-plugin. While not a large piece of code, it will have to be rewritten if we ever decide to move away from this plugin or from Maven in general. I tried exploring other possibilities, such as serializing the depndency tree at build time, but it proved to be a lot more work for not much more benefit.
* This requires embedding portions of Maven itself into Glowstone in order to perform the dependency resolution. While this doesn't tie us to Maven for builds like the dependency checking does, it still ties us to resolving dependencies in a certain way. If we decide to move build systems, it could be awkward to have two different ways of resolving dependencies depending on if they are runtime vs builtin. Currently, we're using Maven, so I think it's fine to use it for our library downloads as well.
* Naether provides the ability to download the dependencies itself, but I am explicitly disabling that since we have our own download manager. We may choose to use this at some point in the future, but I felt it was better to let this functionality bake before trying this out.
* Checksums are required for publishing on Maven Central, so we could take advantage of those in many cases, but I decided to put that as a possible future task rather than making this change even bigger.

*Do note that due to concerns in the previous commit, I reverted my Kotlin changes. Kotlin will still be included with the Glowstone JAR file.*

As before, let me know what you think of this change! I think the benefits of this outweigh the downsides listed above, but certainly there's room for improvement here.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -1085,7 +1085,7 @@ public class GlowServer implements Server {
                 Library library = entry.getValue().get(0);
                 if (serverContainsLibrary(library)) {
                     conflicts.add(entry.getKey().toString());
-                } else {
+                } else if (!library.isExcludeDependencies()) {
                     Naether naether = clients.computeIfAbsent(
                         library.getRepository(),
                         k -> createNaetherWithRepository(k, libraryFolder)
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -3,18 +3,25 @@ package net.glowstone;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.flowpowered.network.Message;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.MultimapBuilder;
 import com.google.common.collect.Multimaps;
+import com.google.common.collect.Sets;
 import com.jogamp.opencl.CLDevice;
 import com.jogamp.opencl.CLPlatform;
+import com.tobedevoured.naether.NaetherException;
+import com.tobedevoured.naether.api.Naether;
+import com.tobedevoured.naether.impl.NaetherImpl;
+import com.tobedevoured.naether.util.RepoBuilder;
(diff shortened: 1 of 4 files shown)


Pr0methean:
Will this affect the ability to use libraries that aren't on Maven Central?

src/main/java/net/glowstone/GlowServer.java
```diff
@@ -425,6 +434,13 @@
      * Whether the macOS/BSD kqueue native transport is available for Netty.
      */
     public static final boolean KQUEUE = KQueue.isAvailable();
+    /**
+     * Libraries that are known to cause problems when loaded up via the server config or plugin
+     * dependencies.
+     */
+    private static final Set<LibraryKey> blacklistedRuntimeLibs = ImmutableSet.of(
+            new LibraryKey("it.unimi.dsi", "fastutil")
```
kashike, line 438: Why is this blacklisted? What problems does it cause?

smartboyathome: It causes classpath conflicts with FastUtil-Lite, which we include in the server JAR.

smartboyathome:
After a while, finally got back to this. I made a small change to better address using libraries that aren't in a maven repository. Prior to the most recent commit, using a library not in a maven repo would need to use an invalid URL as the repository key in order to not affect any other dependencies being resolved. Now, you can set the exclude-dependencies flag within the library's config section to true in order to keep from attempting to resolve the dependencies for these libraries.

aramperes (approved):

mastercoms:
@smartboyathome Could you please resolve the merge conflicts?

smartboyathome:
Sure, I can do that later today.

smartboyathome:
Merge conflicts resolved.

PR #891, opened 2018-03-29
GlowServer.getProfile without async lookup

### LINKED ISSUE #881: Server.createProfile(String name) should not trigger an UUID lookup
Problematic code:

https://github.com/GlowstoneMC/Glowstone/blob/911ba9ea99a4750d6ca2f99065fd7fd08a7f77dd/src/main/java/net/glowstone/GlowServer.java#L1689-L1697

This should not lookup the UUID for the given name. `PlayerProfile` is defined to have at least **either** the UUID or name defined, but not necessarily both. This code may be used with names that don't really exist, and therefore shouldn't trigger an UUID lookup.

An example for such usage is the new `PaperServerListPingEvent` added in https://github.com/PaperMC/Paper/pull/980. Fake `PlayerProfile`s may be used to display custom messages in the list of players displayed when hovering the player count in the server list.

I've ported `PaperServerListPingEvent` to Glowstone in #880, and running [ServerListPlus](https://www.spigotmc.org/resources/serverlistplus.241/) on it then results in a significant delay and the following error: https://gist.github.com/Minecrell/448ad637b2e9d033b1864f522da8d773 because it is looking up the fake profile.

It seems like this was done in #817 because `GlowPlayerProfile` doesn't support having no UUID defined (see https://github.com/GlowstoneMC/Glowstone/pull/817#discussion_r162812665). However, it should be refactored to support this to match the API specifications.

This is needed for #880.

Pr0methean:
Fixes #881. Only the new overload that takes a boolean parameter will trigger a UUID lookup from the Mojang server; to maintain compatibility with existing behavior, the existing code now uses that overload and sets the parameter to true.

### DIFF
--- a/src/main/java/net/glowstone/GlowOfflinePlayer.java
+++ b/src/main/java/net/glowstone/GlowOfflinePlayer.java
@@ -89,6 +89,7 @@ public final class GlowOfflinePlayer implements OfflinePlayer {
     // Core properties
 
     private void loadData() {
+        profile.completeCached();
         try (PlayerReader reader = server.getPlayerDataService().beginReadingData(getUniqueId())) {
             hasPlayed = reader.hasPlayedBefore();
             if (hasPlayed) {
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -1744,7 +1744,19 @@ public class GlowServer implements Server {
 
     @Override
     public GlowPlayerProfile createProfile(UUID id, String name) {
-        return new GlowPlayerProfile(name, id);
+        return createProfile(id, name, false);
+    }
+
+    /**
+     * Creates a player profile.
+     *
+     * @param name The player's name.
+     * @param uuid The player's UUID; may be null.
+     * @param asyncLookup If true and {@code uuid} is null, the UUID is looked up asynchronously.
+     * @return The player's profile.
+     */
+    public GlowPlayerProfile createProfile(UUID uuid, String name, boolean asyncLookup) {
+        return new GlowPlayerProfile(name, uuid, asyncLookup);
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


stephan-gh:
Looks fine, but you need to remove the null check in the `GlowPlayerProfile` constructor to make it work properly:

```
10:41:59 [SEVERE] Could not pass event StatusEvent to ServerListPlus v3.4.9-SNAPSHOT
org.bukkit.event.EventException
        ...
Caused by: java.lang.NullPointerException: uuid must not be null
        at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:787)
        at net.glowstone.entity.meta.profile.GlowPlayerProfile.<init>(GlowPlayerProfile.java:93)
        at net.glowstone.entity.meta.profile.GlowPlayerProfile.<init>(GlowPlayerProfile.java:60)
        at net.glowstone.GlowServer.createProfile(GlowServer.java:1760)
        at net.glowstone.GlowServer.createProfile(GlowServer.java:1748)
        at net.glowstone.GlowServer.createProfile(GlowServer.java:1743)
        at net.glowstone.GlowServer.createProfile(GlowServer.java:215)
        at net.minecrell.serverlistplus.bukkit.handlers.PaperEventHandler.handlePaperServerListPing(PaperEventHandler.java:101)
        at net.minecrell.serverlistplus.bukkit.handlers.PaperEventHandler.onServerListPing(PaperEventHandler.java:28)
        at com.destroystokyo.paper.event.executor.asm.generated.GeneratedEventExecutor3.execute(Unknown Source)
        at org.bukkit.plugin.EventExecutor$2.execute(EventExecutor.java:70)
        ... 45 more
```

Pr0methean:
@Minecrell Done.

stephan-gh:
Working fine now, thanks for making the changes.

mastercoms (approved):

PR #893, opened 2018-03-31
Fixed TODO: "Is main hand always the correct slot?"

fionera:
This pull request removes the TODO in the Bow Item

Its now checked in which Hand the Bow is so he can be damaged

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemBow.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemBow.java
@@ -103,8 +103,12 @@ public class ItemBow extends ItemTimedUsage {
             if (ThreadLocalRandom.current().nextDouble() < 0.2) {
                 launchedArrow.setCritical(true);
             }
-            // TODO: Is main hand always the correct slot?
-            player.setItemInHand(InventoryUtil.damageItem(player, bow));
+
+            if (player.getItemInHand() == bow) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/block/itemtype/ItemBow.java
```diff
@@ -103,8 +103,12 @@ public void endUse(GlowPlayer player, ItemStack bow) {
             if (ThreadLocalRandom.current().nextDouble() < 0.2) {
                 launchedArrow.setCritical(true);
             }
-            // TODO: Is main hand always the correct slot?
-            player.setItemInHand(InventoryUtil.damageItem(player, bow));
+
+            if (player.getItemInHand() == bow) {
```
aramperes, line 107: The way Vanilla functionality works, you can predict in which slot the bow was in:

- If the item type of the main slot is `Material.BOW`, the main slot was used.
- Else, the off slot was used.

(The main slot takes priority, so even if the player has bows in both hands, you can expect that the one in the main hand was used).

This avoids the risk that the bow item is changed in the meantime, thus potentially breaking any equality check.

aramperes (changes requested):

mastercoms:
Merged in https://github.com/GlowstoneMC/Glowstone/commit/db9e6718ee3ae8ef4c553569c29947ed430a0499, with the material check.

PR #894, opened 2018-03-31
Implement isSlimeChunk

fionera:
isSlimeChunk was alwas false. This implements the algorithm for it.

### DIFF
--- a/src/main/java/net/glowstone/chunk/GlowChunk.java
+++ b/src/main/java/net/glowstone/chunk/GlowChunk.java
@@ -110,9 +110,7 @@ public class GlowChunk implements Chunk {
     @Setter
     private boolean populated;
 
-    /**
-     * Cache for isSlimeChunk.
-     */
+    @Setter
     private int isSlimeChunk = -1;
 
     /**
@@ -204,7 +202,7 @@ public class GlowChunk implements Chunk {
             boolean includeBiomeTempRain) {
         return new GlowChunkSnapshot(x, z, world, sections,
                 includeMaxBlockY ? heightMap.clone() : null, includeBiome ? biomes.clone() : null,
-                includeBiomeTempRain);
+                includeBiomeTempRain, isSlimeChunk());
     }
 
     @Override
--- a/src/main/java/net/glowstone/chunk/GlowChunk.java
+++ b/src/main/java/net/glowstone/chunk/GlowChunk.java
@@ -4,7 +4,14 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import lombok.Data;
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/chunk/GlowChunk.java
```diff
@@ -3,13 +3,8 @@
...
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+
+import java.util.*;
```
kashike: Simply adding an import would work, and would be better.

kashike (changes requested):

src/main/java/net/glowstone/chunk/GlowChunk.java
fionera: Will fix it. IntelliJs automatic import optimize was on :/

src/main/java/net/glowstone/chunk/GlowChunk.java
```diff
@@ -170,7 +177,17 @@ public GlowBlock getBlock(int x, int y, int z) {
...
     @Override
     public boolean isSlimeChunk() {
-        return false; // TODO: implement slime chunks
+        if (isSlimeChunk == -1) {
+            boolean isSlimeChunk = new Random(this.world.getSeed()
+                    + (long) (this.x * this.x * 0x4c1906)
+                    + (long) (this.x * 0x5ac0db)
+                    + (long) (this.z * this.z) * 0x4307a7L
+                    + (long) (this.z * 0x5f24f) ^ 0x3ad8025f).nextInt(10) == 0;
+
```
Pr0methean, line 188: Could you please inline java.util.Random's formula? This method may end up being called more frequently than we can afford to create objects!

fionera: Thats why i added the cache or what do you mean?

src/main/java/net/glowstone/chunk/GlowChunk.java
```diff
@@ -170,7 +177,17 @@ public GlowBlock getBlock(int x, int y, int z) {
 
     @Override
     public boolean isSlimeChunk() {
-        return false; // TODO: implement slime chunks
+        if (isSlimeChunk == -1) {
+            boolean isSlimeChunk = new Random(this.world.getSeed()
```
aramperes, line 183: Can you add a link or source for this formula (e.g. `Taken from the Minecraft Wiki`)

src/main/java/net/glowstone/chunk/GlowChunk.java
Pr0methean: What I mean is to just calculate whether a new `Random` created with the given seed *would* output 0 on the first call to `nextInt(10)`, rather than actually calling the constructor of `Random`. That way, you're not creating an object that has to be garbage-collected later, even if the JVM's runtime compiler fails to optimize that creation and garbage collection out. (It *should* optimize them out, but compiler optimizations don't always do what they should.)

Pr0methean: On second thought, if you can't do this, it's okay; we'll optimize it later.

fionera: Maybe give the Object in Constructor? That way you only clone it and set the seed per chunk.

Pr0methean: My concern was about the allocation of memory and the process of garbage collection, both of which would happen just the same whether the object's constructor ran or not.

fionera: Ah ok. Now its saved in the Regionfiles

Pr0methean (approved):

PR #895, opened 2018-03-31
Create i18n enums for better localized string validation.

smartboyathome:
This design tries to make our string localization more stable by forcing us to use enum constants. By doing this, we are able to write a test against it to verify that all strings exist, guaranteeing we won't get an exception when fetching the strings from the config file. Depending on how this design is taken, we may decide to incorporate some of the work from #892 and supersede that PR, or we may decide to merge that and rebase this.


src/main/java/net/glowstone/i18n/LocalizedString.java
```diff
@@ -0,0 +1,13 @@
+package net.glowstone.i18n;
+
+import java.util.ResourceBundle;
+
+public interface LocalizedString {
+    ResourceBundle STRINGS = ResourceBundle.getBundle("strings");
+
```
Pr0methean: Add `String format(Object... args)` and a `void log(Object... args)` that knows the appropriate log level.

Pr0methean:
Should the enum values maybe be sorted by subject matter rather than by log level?

Pr0methean:
Also, how hard would it be to do the conversion if #892 was merged first? You'd just have to convert static methods to instance methods, but you'd have to be able to look up the enum entries by their key at edit-time (which I don't think IntelliJ IDEA can do, unless there's a plugin specifically for this design pattern; but I don't know what tools you're using).

aramperes:
Although I think this is fine, how about making the interfaces be subinterfaces of `LocalizedString`? e.g. `LocalizedString.Console.Info.IMPORT`

smartboyathome:
Pr0methean, It won't be the end of the world if #892 gets merged before this does, it'll just force me to redo all the conversions over to using these implementations since every single one will be in conflict. Upside is, as you said, it would get me the log* and format methods.

Momo, I figured out a way to do that, but the test I wrote will need to be updated to use reflection in order to gather all the LocalizedString instances. I'm working on this change right now.

aramperes (approved):

PR #897, opened 2018-04-11
Enhance Entity Physics Calculation

XuZhen86:
This pull request tries to solve the problem that entity trajectory is glitchy, especially when the player breaks a block and the item drops.

The major problem is, in GlowEntity.pulsePhysics(), only one case of the formula is used. According to Wiki, there are actually two cases used on different entities.

> Note that when thrown objects and arrows are simulated, the drag is applied before the acceleration, rather than after; this is why their terminal velocities are whole numbers while the others aren't.
[Entity](https://minecraft.gamepedia.com/Entity#cite_note-thrown-1)

This issue caused the server-side trajectory of many entities differ from the client-side trajectory. The glitchy happens when after the client animated the anticipated trajectory of the second tick, the server sends a different location of the second tick. The difference is significant enough that the client has to "glitch" the entity to the location the server sent. Because for each tick, the anticipated location and the server-calculated location is different, the client has to "glitch" every tick.

Works done:
1. Rewrite GlowEntity.pulsePhysics() to include the missing formula and enhance readability
2. Correct entity physics parameter according to Wiki
3. Small fixes

Known issues:
1. Entity still "vibrates" on the floor

To-do:
1. ~~Implement terminal velocity for entities.~~
2. Use bounding box interception to calculate the exact location where the entity touches the floor. This should solve the "vibration" problem.
3. Fix anything broken caused by 2?

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1481,6 +1481,12 @@ public class GlowWorld implements World {
         return (T) spawn(location, descriptor.getEntityClass(), reason);
     }
 
+    /**
+     * Spawn an item at the given {@link Location} without shooting effect.
+     *
+     * @param location  the {@link Location} to spawn the item at
+     * @param item      the {@ItemStack} the item should have
+     */
     @Override
     public GlowItem dropItem(Location location, ItemStack item) {
         GlowItem entity = new GlowItem(location, item);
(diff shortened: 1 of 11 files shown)


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1491,14 +1497,33 @@ public GlowItem dropItem(Location location, ItemStack item) {
...
+        // Calculate initial velocity using radius and offsetY as constant
+        // offsetX and offsetZ are calculated using random and Pythagorean theorem
+        double radius = 0.1;
+        double offsetX = tlr.nextDouble(radius * 2) - radius;
+        double offsetY = 0.15;
+        double offsetZ = Math.sqrt(Math.pow(radius, 2) - Math.pow(offsetX, 2));
+
+        // The previous calculation always gives a non-negative zOffset
+        // This adds a 50% chance of offsetZ being negative
+        if (tlr.nextInt(2) == 0) {
```
aramperes: Use `nextBoolean` instead

aramperes (changes requested):

aramperes (approved):

Pr0methean:
I don't think terminal velocity has to be implemented separately. Isn't it just the vertical velocity at which gravity and drag happen to cancel out, like in real life?

XuZhen86:
You are right. I was thinking about this but did not have a chance to verify. It's looks more obvious when you say they can cancel out like in real life.
> Note that when thrown objects and arrows are simulated, the drag is applied before the acceleration, rather than after; _this is why their terminal velocities are whole numbers while the others aren't_.

Pr0methean:
Actually, I'm not 100% sure I'm right, because the discrete-step nature of the physics model may cause artifacts that I'm not aware of. For example, if terminal velocity is overshot on one tick, then it will be undershot on the next. Hopefully the amount of overshoot or undershoot goes to zero asymptotically, but I'm not 100% sure that's the case; maybe there are period-2 attractors.

aramperes:
Thank you for your valuable contributions to the project so far 😃. We have invited you as a member of the Glowstone Project organization; you should receive an email from Github shortly.

In addition, we'd like to invite you to join our [Discord](https://discord.gg/TFJqhsC) server, where you'll be granted the Developer role. Thanks!

XuZhen86:
Hello, momothereal. After carefully considering your invitation, I decide not to join the Glowstone Project organization. My reason is, as I'm becoming a Junior in Software Engineering with heavy course works in the next semester, I will not have enough time to fulfill my responsibility as an active developer on Github. Therefore, I appreciate you giving me a chance to step up from a contributor, but I just don't want to see the moment I have to step down because I can't do much in this project.

We as a team, Glowminers, chose Glowstone as our project for the ```Intro to Software Eng``` class. During the semester, we learn the basics of software development by contributing on Github. Glowstone taught us more than we expected and we had a great time working on the project and of course, with you developers. Our team secured the ```initiative``` in the class by fixing and merging 14 bug fixes, which is far more than any other team.

That being said, it is time for the team to say goodbye to Glowstone. The class gave us a chance to do real contribution on GitHub and we are glad to see, at the end of this semester, the difference we made to GitHub and Glowstone. We will make contribution to Glowstone individually but not as a team. In all, we will recommend Glowstone to the students in ```Intro to Software Eng``` class next semester and we look forward to contribute to Glowstone in the future!

aramperes:
Hey!

As a CompSci student as well I concur about the course load 😄. We're really glad you guys chose Glowstone for your term project, and we're looking forward to hear from future students in your program. Thank you for the effort you've put into this project, and we hope to see more from you all in the future.

Membership inside the Glowstone project does not imply any responsibilities when it comes to time spent, activity, contributions, etc. Members of the project are trusted for their own discipline, code quality, and for understanding the core goal of the project. That said, we understand if you prefer not carrying the title for personal reasons. However, we're always open if you decide to change your mind! :)

Again, thank you for your exceptional work, and we're looking forward to hearing from members of your team in the future!

PR #900, opened 2018-04-17
Start interacting with animals

FlorentClarret:
Hi,

This PR adds the possibility to feed animals (give them breeding foods). I got all the data from here : https://minecraft.gamepedia.com/Breeding#Breeding_foods.

I did not implement the love mode, healing and so on yet. The only thing done is that the item is removed from the player inventory. Also, wolves or ocelots can not be tamed yet. I'm planning to do this on other PRs if this one is OK for you.

Don't hesitate to comment the code or ask if you have any question/remark. Thanks

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowAnimal.java
+++ b/src/main/java/net/glowstone/entity/GlowAnimal.java
@@ -1,10 +1,17 @@
 package net.glowstone.entity;
 
+import java.util.EnumSet;
+import java.util.Set;
 import net.glowstone.entity.ai.EntityDirector;
 import net.glowstone.entity.ai.MobState;
+import net.glowstone.net.message.play.player.InteractEntityMessage;
+import net.glowstone.util.InventoryUtil;
+import org.bukkit.GameMode;
 import org.bukkit.Location;
+import org.bukkit.Material;
 import org.bukkit.entity.Animals;
 import org.bukkit.entity.EntityType;
+import org.bukkit.inventory.ItemStack;
 
 /**
  * Represents an Animal, such as a Cow.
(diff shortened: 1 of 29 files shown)


src/main/java/net/glowstone/entity/GlowAnimal.java
```diff
@@ -31,4 +38,36 @@ public GlowAnimal(Location location, EntityType type, double maxHealth) {
...
+
+        return false;
+    }
+
+    /**
+     * Returns a set containing the breeding foods for the current animal.
+     * @return a set containing Material
+     */
+    public Set<Material> getBreedingFoods() {
+        return EnumSet.noneOf(Material.class);
```
aramperes: Perhaps cache that set in a static field?

FlorentClarret: This default is almost never used, because most of all the subclasses override this method. That's why i didn't cache this one. Fixed now ;)

aramperes (approved):

src/main/java/net/glowstone/entity/GlowAnimal.java
```diff
@@ -31,4 +42,36 @@ public GlowAnimal(Location location, EntityType type, double maxHealth) {
...
+                player.getInventory().consumeItemInMainHand();
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Returns a set containing the breeding foods for the current animal.
+     * @return a set containing Material
```
Pr0methean: Mention that the returned set is immutable.

src/main/java/net/glowstone/entity/passive/GlowOcelot.java
```diff
@@ -1,15 +1,20 @@
...
+import org.bukkit.Material;
 import org.bukkit.Sound;
 import org.bukkit.entity.AnimalTamer;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.Ocelot;
 
 public class GlowOcelot extends GlowTameable implements Ocelot {
 
+    private static final Set<Material> BREEDING_FOODS = Sets.immutableEnumSet(Material.RAW_FISH);
+
```
Pr0methean: Doesn't raw salmon also work?

src/main/java/net/glowstone/inventory/GlowInventory.java
```diff
@@ -198,7 +198,7 @@ public boolean itemShiftClickAllowed(int slot, ItemStack stack) {
      * @param clickedItem The item at which was clicked
      */
     public void handleShiftClick(GlowPlayer player, InventoryView view, int clickedSlot,
-            ItemStack clickedItem) {
+                                 ItemStack clickedItem) {
         clickedItem = player.getInventory().tryToFillSlots(clickedItem, 8, -1, 35, 8);
```
Pr0methean: Revert this to the old indentation.

Pr0methean (changes requested):

src/main/java/net/glowstone/entity/GlowAnimal.java
FlorentClarret: Done, and I also added a test case to prevent subclasses to return mutable sets

src/main/java/net/glowstone/inventory/GlowInventory.java
FlorentClarret: Done

src/main/java/net/glowstone/entity/passive/GlowOcelot.java
FlorentClarret: Before 1.13, raw salmon is not a specific value of Material, it just has the data byte set to 1 so this is going to work on 1.12.*. However, it will need to be updated for 1.13.

Pr0methean (approved):

aramperes:
Thank you!

PR #901, opened 2018-04-17
Enhance Player Throw

jordanmurray:
This pull request attempts to solve the problem of a player not being able to throw an object with much velocity. The fix simply adds calculations to the initial velocity of the thrown object. The initial velocity is also randomized to match behavior on official server.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowHumanEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowHumanEntity.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 import lombok.Getter;
 import lombok.Setter;
 import net.glowstone.EventFactory;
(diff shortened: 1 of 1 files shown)


aramperes:
Hey, thank you for your contribution. May I ask where the values come from? Are they from trial-and-error?

jordanmurray:
Yes, the values come from a process of trial-and-error to find the closest similar behavior to the official server.

src/main/java/net/glowstone/entity/GlowHumanEntity.java
```diff
@@ -499,8 +500,15 @@ public GlowItem drop(ItemStack stack) {
 
         Location dropLocation = location.clone().add(0, getEyeHeight(true) - 0.3, 0);
         GlowItem dropItem = world.dropItem(dropLocation, stack);
-        Vector vel = location.getDirection().multiply(new Vector(0.11F, 0.0F, 0.11F));
-        vel.setY(0.006F);
+
+        Vector vel = location.getDirection().multiply(0.3);
+        ThreadLocalRandom tlr = ThreadLocalRandom.current();
+        double randOffset = 0.02;
```
Pr0methean: Extract these constants to private static final fields.

Pr0methean (changes requested):

src/main/java/net/glowstone/entity/GlowHumanEntity.java
aramperes: These values are approximate and have no real significance, I don't think constants are necessary. But it shouls have comments clarifying the calculations a bit.

Pr0methean (approved):

PR #902, opened 2018-04-22
Item Placement with Signs

smcconke:
This PR aims to fix the block placement interaction involving both sign materials, SIGN_POST and WALL_SIGN.

**Observed Behavior:**  While holding items only pertaining to BlockType, the player can place these materials around and on top of the sign.  When placing the blocks, the player's target block will be a variation of the sign and the face will be either North, East, South, West, or Up.  The same output occurs when attempting to place a sign onto another variation of the sign.

**Expected Behavior:**  With Vanilla Minecraft, a player **cannot** place a block or sign onto a placed SIGN_POST or WALL_SIGN unless the player is sneaking while right clicking on the target.

The implementation uses a private variable, pertaining to all BlockTypes, that will be set once a player attempts to right click and place a block.  This approach allows for all BlockTypes to check this condition without passing the instance of the player as another argument in canPlaceAt.  While compiling BlockNeedsAttached.java, checkstyle produced an error pertaining to javadocs, therefore documentation was given for canAttachTo.  In BlockSign.java, canPlaceAt is overrided to provide a different implementation from that of BlockNeedsAttached.


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -52,6 +53,15 @@
...
     protected SoundInfo placeSound = new SoundInfo(Sound.BLOCK_WOOD_BREAK, 1F, 0.75F);
 
+    /**
+     * Determines whether or not the player is sneaking while placing a block.
+     * 
+     * @return If the player is sneaking while placing.
+     */
+    @Getter
+    @Setter
+    private boolean shiftClickPlace = false;
```
Pr0methean: This won't work properly in the case where multiple players try to interact with the same block in a short timespan. Instead, whatever method calls `canPlaceAt` or `canAttachTo` should check whether the *relevant* player `isSneaking()`.

Pr0methean (changes requested):

smcconke:
The following changes take into consideration of the relevant player by considering if the instance of the `GlowPlayer player` passed in `rightClickBlock` is sneaking.  The switch statement will determine when `canPlaceAt` is called for the specified `targetMat`.  `BlockSign` now returns True if it can attach to the block, or if the target material is that of `SIGN_POST` or `WALL_SIGN`.

Also, when determining the target material, would it be best to place this in a method within `BlockType` since we are determining target material in `BlockSign` as well in `canPlaceAt`?

Thanks!

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -369,7 +352,19 @@ public final void rightClickBlock(GlowPlayer player, GlowBlock against, BlockFac
         }
 
         // call canBuild event
-        boolean canBuild = canPlaceAt(target, face);
+        boolean canBuild = true;
```
Pr0methean: Revert this, and then you can use a simple `if (!player.isSneaking())` with no `else`.

Pr0methean (approved):

aramperes:
Thank you for your valuable contributions to the project so far 😃. We have invited you as a member of the Glowstone Project organization; you should receive an email from Github shortly.

In addition, we'd like to invite you to join our [Discord](https://discord.gg/TFJqhsC) server, where you'll be granted the Developer role. Thanks!

PR #904, opened 2018-04-27
i18nize more strings

Pr0methean:
Internationalizes some more strings; marks non-localizable strings as NON-NLS.


aramperes (approved):

aramperes (dismissed):
You have a few unused imports to remove

aramperes (approved):

PR #905, opened 2018-04-27
Refactor CompoundTag reading with a functional syntax

Pr0methean:
We can now check a tag's type and process it if it's the expected type, all with one method call and one copy of the tag-name literal.


src/main/java/net/glowstone/inventory/GlowMetaSkull.java
```diff
@@ -98,10 +98,12 @@ void writeNbt(CompoundTag tag) {
     void readNbt(CompoundTag tag) {
         super.readNbt(tag);
         if (tag.containsKey("SkullOwner")) {
-            if (tag.isString("SkullOwner")) {
-                owner.set(GlowPlayerProfile.getProfile(tag.getString("SkullOwner")).join());
-            } else if (tag.isCompound("SkullOwner")) {
-                owner.set(GlowPlayerProfile.fromNbt(tag.getCompound("SkullOwner")).join());
+            if (!tag.readString(
```
aramperes: I think flipping the key (string) and the function params for "read*" methods would make their usage more readable.

aramperes (dismissed):
There's a lot of stuff here, I'll do a more extensive review later.

src/main/java/net/glowstone/inventory/GlowMetaSkull.java
Pr0methean: Then they can't be variadic, since varargs have to come last. I'll look into how often I'm actually using them with more than one key argument, now that the conversions are done.

yannicklamprecht:
I like it how Java 1.8 can cleanup code using consumables.

src/main/java/net/glowstone/block/entity/FlowerPotEntity.java
```diff
@@ -25,15 +25,14 @@ public FlowerPotEntity(GlowBlock block) {
...
-            contents = ItemIds.getItem(tag.getString("Item")).getNewData((byte) contentsData);
-        } else if (tag.isInt("Item")) {
+        final Material[] item = {null};
+        // NBT data uses material ID names (post-1.8).
+        if (!tag.readString("Item", itemString -> item[0] = ItemIds.getItem(itemString))) {
             // NBT data uses material IDs (pre-1.8).
-            contents = Material.getMaterial(tag.getInt("Item")).getNewData((byte) contentsData);
+            tag.readInt("Item", itemInt -> item[0] = Material.getMaterial(itemInt));
+        }
+        if (item[0] != null) {
```
mastercoms: I'm not sure if I like this. Could `readX` allow for chaining with `andThen` or similar (maybe we need a method for failures as well). Then if it's still necessary, we could move the bool that returns if the tag exists to a new method.

src/main/java/net/glowstone/block/entity/MobSpawnerEntity.java
```diff
@@ -27,19 +27,11 @@ public MobSpawnerEntity(GlowBlock block) {
...
         super.loadNbt(tag);
-
-        if (tag.isString("EntityId")) {
-            spawning = EntityType.fromName(tag.getString("EntityId"));
-            if (spawning == null) {
-                spawning = DEFAULT;
-            }
-        } else {
+        tag.readString("EntityId", name -> spawning = EntityType.fromName(name));
+        if (spawning == null) {
```
mastercoms: This could also be chained.

src/main/java/net/glowstone/io/entity/EndermanStore.java
```diff
@@ -16,37 +16,35 @@ public EndermanStore() {
...
-                }
-            }
+            });
         }
-        if (carried != null) {
-            if (compound.isShort("carriedData")) {
-                carried.setData((byte) compound.getShort("carriedData"));
-            }
-            entity.setCarriedMaterial(carried);
+        if (carried[0] != null) {
```
mastercoms: Chaining could be beneficial here as well.

src/main/java/net/glowstone/util/nbt/CompoundTag.java
```diff
@@ -124,8 +134,12 @@ public void mergeInto(CompoundTag other, boolean overwrite) {
     ////////////////////////////////////////////////////////////////////////////
     // Simple gets
 
-    public boolean getBool(String key) {
-        return containsKey(key) && getByte(key) != 0;
+    public boolean getBoolDefaultFalse(String key) {
```
mastercoms: Could we just have `getBool(String)` and `getBool(String, boolean)`?

src/main/java/net/glowstone/io/nbt/NbtSerialization.java
```diff
@@ -127,25 +124,24 @@ public static CompoundTag writeItem(ItemStack stack, int slot) {
...
     public static World readWorld(GlowServer server, CompoundTag compound) {
-        World world = null;
-        if (compound.isLong("WorldUUIDLeast") && compound.isLong("WorldUUIDMost")) {
-            long uuidLeast = compound.getLong("WorldUUIDLeast");
-            long uuidMost = compound.getLong("WorldUUIDMost");
-            world = server.getWorld(new UUID(uuidMost, uuidLeast));
-        }
-        if (world == null && compound.isString("World")) {
-            world = server.getWorld(compound.getString("World"));
+        final World[] world = {null};
```
mastercoms: Chaining might be able to be done here, not sure if it would be messy though.

src/main/java/net/glowstone/io/nbt/NbtStructureDataService.java
```diff
@@ -80,28 +75,32 @@ public NbtStructureDataService(GlowWorld world, File structureDir) {
...
+                            if (maybeFeatures != null) {
+                                features = maybeFeatures;
                             }
                         }
                     } catch (IOException e) {
                         server.getLogger().log(Level.SEVERE,
                             "Failed to read structure data from " + structureFile, e);
                     }
                 }
+                if (data == null) {
```
mastercoms: Chain?

mastercoms (dismissed):

src/main/java/net/glowstone/util/nbt/CompoundTag.java
Pr0methean: Done; see getBoolean(String, boolean).

mastercoms (approved):

PR #907, opened 2018-04-28
Feeding baby animals accelerate their growth

FlorentClarret:
Hi,

Currently, animals can be fed, but it does nothing except removing the item in the player's main hand. With this PR, feeding babies will slowly accelerate their growth. Most of the time, feeding a baby takes 10% off the remaining time to grow up. However, it's a bit different for horses and llamas because it depends on the item.

I'm not really sure that my english is always correct here, so if you see something strange, tell me.


src/main/java/net/glowstone/entity/GlowAgeable.java
```diff
@@ -147,16 +148,14 @@ public boolean entityInteract(GlowPlayer player, InteractEntityMessage message)
...
                     if (player.getGameMode() == GameMode.SURVIVAL
                         || player.getGameMode() == GameMode.ADVENTURE) {
-                        // Consume the egg
-                        if (item.getAmount() > 1) {
-                            item.setAmount(item.getAmount() - 1);
-                        } else {
-                            player.getInventory()
-                                .setItem(message.getHandSlot(), InventoryUtil.createEmptyStack());
-                        }
+                        player.getInventory().consumeItemInMainHand();
```
aramperes: You can get the hand slot that was used from the message data

src/main/java/net/glowstone/entity/GlowAnimal.java
```diff
@@ -74,4 +75,13 @@ public boolean entityInteract(GlowPlayer player, InteractEntityMessage message)
     public Set<Material> getBreedingFoods() {
         return DEFAULT_BREEDING_FOODS;
     }
+
+    @Override
+    protected int computeGrowthAmount(Material material) {
+        if (!isAdult() && getBreedingFoods().contains(material)) {
+            return Math.abs(getAge() / 10);
+        } else {
```
aramperes: The else statement is redundant because of the return statement.

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
```diff
@@ -67,7 +78,7 @@ private int getHorseFlags() {
         }
         if (this instanceof GlowHorse) {
             GlowHorse horse = (GlowHorse) this;
-            if (getInventory() != null && ((HorseInventory) getInventory()).getSaddle() != null) {
+            if (getInventory() != null && getInventory().getSaddle() != null) {
```
aramperes: Replace the null check for the saddle with `InventoryUtil.isEmpty`

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
```diff
@@ -90,4 +101,20 @@ private int getHorseFlags() {
...
+    @Override
+    protected int computeGrowthAmount(Material material) {
+        int amount = 0;
+
+        // We need to be a baby and only tamed horses can be fed with hay block
+        if (!isAdult() && !(Material.HAY_BLOCK == material && !isTamed())) {
+            Integer mapResult = GROWING_FOODS.get(material);
+
+            if (mapResult != null) {
+                amount = Math.min(mapResult, Math.abs(getAge()));
```
aramperes: No need for that local variable. Just return at this line, and return `0` and the end of the method.

src/main/java/net/glowstone/entity/passive/GlowLlama.java
```diff
@@ -11,10 +14,20 @@
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.Llama;
 
+/**
+ * Represents a llama.
+ * The data come from https://minecraft.gamepedia.com/Llama
```
aramperes: Typo: the data come**s** from

Although this is debatable because *data* is the plural form of *datum*. However, *data* is used as a singular mass noun in everyday usage.

src/main/java/net/glowstone/entity/passive/GlowLlama.java
```diff
@@ -81,4 +94,20 @@ protected GlowLlamaInventory createNewInventory() {
...
+
+    @Override
+    protected int computeGrowthAmount(Material material) {
+        int amount = 0;
+
+        if (!isAdult()) {
+            Integer mapResult = GROWING_FOODS.get(material);
+
+            if (mapResult != null) {
+                amount = Math.min(mapResult, Math.abs(getAge()));
```
aramperes: See previous review comment about the local variable.

aramperes (changes requested):

FlorentClarret:
Done ;)

Pr0methean (approved):

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
```diff
@@ -67,7 +79,7 @@ private int getHorseFlags() {
         }
         if (this instanceof GlowHorse) {
             GlowHorse horse = (GlowHorse) this;
-            if (getInventory() != null && ((HorseInventory) getInventory()).getSaddle() != null) {
+            if (getInventory() != null && InventoryUtil.isEmpty(getInventory().getSaddle())) {
```
aramperes: This should be the opposite (`!InventoryUtil.isEmpty`)

aramperes (changes requested):

aramperes (approved):

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
FlorentClarret: oops, my bad

PR #908, opened 2018-04-29
[i18n] 3 nesting levels is enough!

Pr0methean:
This simplifies LocalizedStrings by limiting the nesting depth within it to 3 levels. Excessive nesting bloats the jar file, slows down IDEs, forces very short lines due to the indentation, and makes it harder to find a relevant message. In several cases, a 4th-level sub-interface had the same name (modulo case) as a message, which can't have helped readability.


aramperes (approved):

FlorentClarret (approved):

smartboyathome (approved):
Overall, I agree with these changes. We might want to rethink the levels altogether. Was never fully comfortable with the sub-interface thing, but I didn't have any hard data to back up my worries, and it ran fine on my machine.

PR #491, opened 2017-06-08
[WIP] Tool vs hardness

satoshinm:
Extends https://github.com/GlowstoneMC/Glowstone/pull/473 variable block hardness based on tool:

* Tool types: 1.5x slower if can tool can mine the block, 5x slower if cannot
* Tool materials: 2x, 4x, 6x, 8x, or 12x faster for wood, stone, iron, diamond, gold tool materials

See http://minecraft.gamepedia.com/Breaking#Speed

PR #910, opened 2018-04-30
Digging fixes

### LINKED ISSUE #506: Tool should affect block mining speed
Implement http://minecraft.gamepedia.com/Breaking#Speed, to summarize:

* Tool types: 1.5x slower if can tool can mine the block, 5x slower if cannot
* Tool materials: 2x, 4x, 6x, 8x, or 12x faster for wood, stone, iron, diamond, gold tool materials
* Special cases: shears 15x (5x on wool) and sword 1.5x (15x on cobwebs)

https://github.com/GlowstoneMC/Glowstone/pull/473 added basic block hardness, but the tool type and material also needs to be taken into account. I started working on this in https://github.com/GlowstoneMC/Glowstone/pull/491 but ran out of time to finish it and test it so I am opening this issue instead for someone else to if anyone is interested.

To test, sign in with two accounts, have one player start digging a block and the other watching it. Use a hard block such as stone or obsidian, and try mining with each tool type and material (not only bare hands).

If implemented correctly, the block break animation in both windows should show identical breakage stages for all tools and materials.

### LINKED ISSUE #852: Digging snow doesn't produce snowballs
Digging snow with a stone (or better) shovel doesn't produce snowballs as expected. It only works with a wooden showel.

When I use a shovel of stone or better, it needs only "one ticks" to break the snow. The only message received from the "handle" method in DiggingHandler class says:
DiggingMessage(state=0, x=-114, y=70, z=94, face=1)

where state=0 stays for "START_DIGGING".
When I use the wooden shovel it takes two ticks and I get both state 0 and state 2 (FINISH_DIGGING) message.

I suppose that messages are sent by clients to the server, maybe the problem is with the managment of message queue?

Pr0methean:
Fixes #852, fixes #506

Block breaking now finishes when the correct number of ticks have elapsed, not when we receive a FINISH_DIGGING message. The response to a premature FINISH_DIGGING is now to re-send the affected block. We also now ignore START_DIGGING on the block we're already digging (which can happen if the client thinks digging is finished when it isn't and the player is still holding left-click).

### DIFF
--- a/src/main/java/net/glowstone/block/BuiltinMaterialValueManager.java
+++ b/src/main/java/net/glowstone/block/BuiltinMaterialValueManager.java
@@ -62,7 +62,7 @@ public class BuiltinMaterialValueManager implements MaterialValueManager {
 
         private Object get(String name) {
             Object got = section.get(name);
-            if (got == null) {
+            if (got == null && this != defaultValue) {
                 return defaultValue.get(name);
             }
             return got;
@@ -76,7 +76,8 @@ public class BuiltinMaterialValueManager implements MaterialValueManager {
 
         @Override
         public ToolType getTool() {
-            return ToolType.valueOf((String) get("tool"));
+            String toolName = (String) get("tool");
+            return toolName == null ? null : ToolType.valueOf(toolName);
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


Pr0methean:
@mastercoms says she doesn't believe @satoshinm (who hasn't signed the CLA, but who submitted some of this PR's commits as part of the abandoned PR #491) would have copied this from Mojang, so I've dismissed the CLA warning.

mastercoms (approved):

PR #911, opened 2018-04-30
Add a TickUtil class to convert minutes/seconds to ticks

FlorentClarret:
Currently, when we are computing durations (for potions or animals for example), we are always calculating with the formula or directly put the final value in ticks. 

I do not find it really easy to read and I'm always wondering how long 4800, 24000 or 6000 ticks really is, that's why I added a utility class to perform the conversion. Thanks to that, we directly know the duration of an effect.

### DIFF
--- a/src/main/java/net/glowstone/util/TickUtil.java
+++ b/src/main/java/net/glowstone/util/TickUtil.java
@@ -1,0 +1,46 @@
+package net.glowstone.util;
+
+/**
+ * Utility class to ease conversion between minutes/seconds and ticks.
+ */
+public final class TickUtil {
+
+    private TickUtil() {
+    }
+
+    /**
+     * Converts seconds (int) into ticks.
+     * @param seconds The seconds to convert.
+     * @return The corresponding number of ticks.
+     */
+    public static int secondsToTicks(int seconds) {
+        return 20 * seconds;
+    }
+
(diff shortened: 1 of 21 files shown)


src/main/java/net/glowstone/util/TickUtil.java
```diff
@@ -0,0 +1,46 @@
+package net.glowstone.util;
+
+/**
+ * Utility class to ease conversion between minutes/seconds and ticks.
```
aramperes: Since this class only does SI -> ticks, change this to "from standard time units to in-game ticks."

src/main/java/net/glowstone/util/TickUtil.java
```diff
@@ -0,0 +1,46 @@
...
+    private TickUtil() {
+    }
+
+    /**
+     * Converts seconds (int) into ticks.
+     * @param seconds The seconds to convert.
+     * @return The corresponding number of ticks.
+     */
+    public static int secondsToTicks(int seconds) {
+        return 20 * seconds;
```
aramperes: Put the 20's in a constant, e.g. TICKS_PER_SECOND

aramperes (changes requested):

aramperes (approved):

PR #912, opened 2018-05-02
Fix bug in OreVein's for-loop

Muqsit:
This probably was a limiting reagent. For better understanding, this is what was being executed prior:
```java
double radius = 2.0;
int origin = 12;
int diff = (int) origin - radius;

for (int x = diff; x <= diff; x++) {
    //...
}
```


aramperes (approved):

aramperes:

PR #919, opened 2018-05-20
Refactors to reduce nesting depth

Pr0methean:
Eliminates deeply-nested ifs using break/continue/return statements. Extracts method `TerrainObject.killPlantAbove(Block)`, which was both duplicated and overly nested, and de-duplicates the PLANT_TYPES collection that it uses.

This PR reduces the nesting depth to 5 or less within 20 of the 27 methods where it currently exceeds that number, found using the "Overly nested method" inspection in IntelliJ IDEA.

Also fixes apparent bugs in MesaGroundGenerator where `colorNoise::noise` was being called with the X coordinate as both X and Y parameters, which would have caused horizontal stripes in mesas where colored terracotta was present.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
@@ -138,7 +138,7 @@ public class BlockRedstone extends BlockNeedsAttached {
                 case LEVER:
                     Lever lever = (Lever) target.getState().getData();
                     if (lever.isPowered()) {
-                        updateDataTo15(me);
+                        setFullyPowered(me);
                         return;
                     }
                     break;
@@ -146,27 +146,27 @@ public class BlockRedstone extends BlockNeedsAttached {
                 case WOOD_BUTTON:
                     Button button = (Button) target.getState().getData();
                     if (button.isPowered()) {
-                        updateDataTo15(me);
+                        setFullyPowered(me);
                         return;
                     }
                     break;
                 case DIODE_BLOCK_ON:
                     Diode diode = (Diode) target.getState().getData();
                     if (face == diode.getFacing().getOppositeFace()) {
-                        updateDataTo15(me);
+                        setFullyPowered(me);
                         return;
                     }
                     break;
                 case REDSTONE_BLOCK:
                 case REDSTONE_TORCH_ON:
-                    updateDataTo15(me);
+                    setFullyPowered(me);
                     return;
                 case OBSERVER:
                     boolean powered = BlockObserver.isPowered(target);
                     BlockFace outputFace = BlockObserver.getFace(target).getOppositeFace();
                     if (powered && target.getRelative(outputFace).getLocation()
                         .equals(me.getLocation())) {
-                        updateDataTo15(me);
+                        setFullyPowered(me);
                         return;
                     }
                     break;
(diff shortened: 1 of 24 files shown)


src/main/java/net/glowstone/block/blocktype/BlockRedstone.java
```diff
@@ -266,6 +240,17 @@ public void updatePhysics(GlowBlock me) {
         }
     }
 
+    /**
+     * Sets the block data to 15 and calls {@link #extraUpdate(GlowBlock)} if it wasn't already 15.
+     * @param me the block to update
+     */
+    protected void updateDataTo15(GlowBlock me) {
```
mastercoms: Rename this to `setPowered` and update the Javadoc accordingly (something like "this block of redstone is set to be directly powered and updates its connections")

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -803,81 +803,82 @@ public void setHealth(double health) {
...
+        if (this instanceof GlowPlayer) {
+            GlowPlayer player = (GlowPlayer) this;
+            ItemStack mainHand = player.getInventory().getItemInMainHand();
+            ItemStack offHand = player.getInventory().getItemInOffHand();
+            if (!InventoryUtil.isEmpty(mainHand) && mainHand.getType() == Material.TOTEM) {
+                player.getInventory().setItemInMainHand(InventoryUtil.createEmptyStack());
+                player.setHealth(1.0);
+                active = true;
+                return;
+            } else if (!InventoryUtil.isEmpty(offHand) && offHand.getType() == Material.TOTEM) {
```
mastercoms: This doesn't need to be an else if since the if branch returns.

src/main/java/net/glowstone/generator/ground/MesaGroundGenerator.java
```diff
@@ -102,42 +103,43 @@ public void generateTerrainColumn(ChunkData chunkData, World world, Random rando
...
                                 << 1)) {
-                                topMat = colored ? GRASS : COARSE_DIRT;
-                                chunkData.setBlock(x, y, z, topMat);
-                            } else if (y > seaLevel + 2 + surfaceHeight) {
-                                int color = colorLayer[(y + (int) Math
+                            topMat = colored ? GRASS : COARSE_DIRT;
+                            chunkData.setBlock(x, y, z, topMat);
+                        } else if (y > seaLevel + 2 + surfaceHeight) {
+                            int color = colorLayer[(y + (int) Math
                                     .round(colorNoise.noise(chunkX, chunkX, 0.5D, 2.0D) * 2.0D))
```
mastercoms: The indenting on this method looks off.

src/main/java/net/glowstone/generator/ground/MesaGroundGenerator.java
```diff
@@ -102,42 +103,43 @@ public void generateTerrainColumn(ChunkData chunkData, World world, Random rando
...
                                     % colorLayer.length];
-                                setColoredGroundLayer(chunkData, x, y, z,
+                            setColoredGroundLayer(chunkData, x, y, z,
                                     y < seaLevel || y > 128 ? 1 : colored ? color : -1);
```
mastercoms: Indenting looks weird here too.

mastercoms (dismissed):

mastercoms (approved):

PR #921, opened 2018-05-22
Add ClearCommandTest, and refactor *CommandTest

Pr0methean:
Extracts the superclass CommandTest, makes all the existing command tests extend it (some via a new helper class CommandTestWithFakePlayers), and pulls up testExecuteFailsWithoutPermission (adding it to one or two command tests that lacked it). Adds ClearCommandTest.


Pr0methean:
This should be tested on either @mastercoms' or @momothereal's machine before merging, because of the recent ServerShim leakage issue.

aramperes (approved):

Issue #922, opened 2018-05-24
API Event Coverage


aramperes:
### *Disclaimer for new contributors*
Glowstone is an open-source project under the MIT license. In addition, we have a [Contributors License Agreement](https://glowstone.net/cla) which includes restrictions on the source of contributions (e.g. decompiled code from Mojang, Spigot, etc.). You will be asked to sign the CLA when submitting your first Pull Request.

### Task

As suggested by creatorfromhell, this is a to-do list of API events that are not partially or fully implemented in Glowstone, including a direct link to their documentation.

Events are an important aspect of our coverage of the Bukkit, Spigot, and Paper APIs.

In **bold** are events whose feature is implemented, but missing the event firing.

### Block Events

- [x] [BeaconEffectEvent](https://papermc.io/javadocs/paper/1.12/com/destroystokyo/paper/event/block/BeaconEffectEvent.html) (#1092)
- [x] [**BlockExplodeEvent**](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/BlockExplodeEvent.html) (#936)
- [ ] [BlockMultiPlaceEvent](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/BlockMultiPlaceEvent.html)
- [x] [**BlockPhysicsEvent**](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/BlockPhysicsEvent.html)
- [X] [BlockPistonExtendEvent](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/BlockPistonExtendEvent.html) (#1013)
- [x] [BlockPistonRetractEvent](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/BlockPistonRetractEvent.html)
- [X] [**BlockRedstoneEvent**](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/BlockRedstoneEvent.html) (#1018)
- [x] [**CauldronLevelChangeEvent**](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/CauldronLevelChangeEvent.html) (2f16c28)
- [ ] [EntityBlockFormEvent](https://papermc.io/javadocs/paper/1.12/org/bukkit/event/block/EntityBlockFormEvent.html)

### Command Events
*Nothing here*

### Enchantment Events
*Nothing here*

### Entity Events
[...]

PR #925, opened 2018-05-27
Implement ItemMergeEvent, ExpBottleEvent, CreeperPowerEvent (#922)

VaiTon:


src/main/java/net/glowstone/entity/monster/GlowCreeper.java
```diff
@@ -59,7 +63,15 @@ protected Sound getHurtSound() {
     public void damage(double amount, Entity source, DamageCause cause) {
         super.damage(amount, source, cause);
         if (DamageCause.LIGHTNING.equals(cause)) {
```
aramperes: Off-topic, but could you add a check for `!isPowered()` in this if statement?

src/main/java/net/glowstone/entity/projectile/GlowThrownExpBottle.java
```diff
@@ -28,6 +31,8 @@ public void collide(LivingEntity entity) {
 
     private void spawnOrb() {
         int xp = ThreadLocalRandom.current().nextInt(9) + 3;
+        ExpBottleEvent event = EventFactory.getInstance()
+                .callEvent(new ExpBottleEvent(this, xp));
```
aramperes, line 35: Set the `xp` variable to the result of `event.getExperience()`

aramperes (changes requested):

src/main/java/net/glowstone/entity/projectile/GlowThrownExpBottle.java
VaiTon: But `xp` has to be calculated before calling the new event

aramperes: I mean after the event is fired. The event object is mutable by plugins, which can use setExperience to change the behaviour. 

So, after event is called, use `xp = event.getExperience();` and keep the rest as is.

aramperes (approved):

aramperes:

PR #926, opened 2018-05-28
Command.execute now always returns true when run without permission

Pr0methean:
Currently, some Command subclasses have execute() methods that return false when the permission check fails, indicating that a usage message should be sent after the no-permission message. This PR fixes that; they all now return true when that happens.

Also, each Command subclass now has a stub test class that inherits `testExecuteWithoutPermission` (renamed from `testExecuteFailsWithoutPermission`, since with `ColorCommand` it instead tests that permission *isn't* required). `SaveToggleCommand` instead has two, one for /save-off and one for /save-on. The missing ones were generated automatically, using a shell script that's also in this PR; only `ColorCommandTest`, `SaveOffCommandTest` and `SaveOnCommandTest` were subsequently edited.

The abstract classes `CommandTest` and `CommandTestWithFakePlayers` are moved from `net.glowstone.command.minecraft` to `net.glowstone.command` since they're now also extended in `net.glowstone.command.glowstone`.

NB: Part of the theory behind this PR is that people are more likely to add the other missing test cases in future commits, if the classes they're changing already have test classes.


mastercoms (dismissed):
Returning false on a command returns usage instructions. Test permission failing already informs the user of their lack of permissions, so we have to return true to prevent duplicated/bogus command warnings.

Pr0methean:
Actually, that's not true: `execute()` itself prints the usage message before returning, if one is needed; see for example https://github.com/GlowstoneMC/Glowstone/blob/99a0e54c3a3c9e45fdd65e3cf2a6758c0fa1ded6/src/main/java/net/glowstone/command/minecraft/TellCommand.java#L34

If I'm wrong, you'll have to show me the calling code (I can't find any in the main source root with IntelliJ's Find Usages command).

mastercoms:
Apologies, that is only true for plugin commands. However, the convention is still to return true for permission errors (see Bukkit and Paper `command.defaults`).

Pr0methean:
Bukkit is also inconsistent; see e.g. https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/command/defaults/WeatherCommand.java (returns false) and https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/command/defaults/GameRuleCommand.java (returns true). We'll probably have to talk to the Bukkit devs to find out what the rule is for which is which.

mastercoms:
Both return true when `!testPermission()` is true.

https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/command/defaults/WeatherCommand.java#L28
https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/command/defaults/GameRuleCommand.java#L30

As for when usage instructions are sent, I see that an extra action is performed for `/gamerule` where it lists the game rules, and thus it returns `true`.

mastercoms (approved):

PR #928, opened 2018-05-29
Refactor: Use createDirectories and FileUtils.copyDirectory

Pr0methean:
Original author was probably unaware of `Files.createDirectories`, since there's no documented reason why we're not using it. @SpaceManiac was added as he seems to be that original author.


aramperes:
I feel like this was done to display helpful errors instead of stacktraces/generic I/O errors.

aramperes:
Actually, isn't the point of that file visitor to simply continue the copying process if a file fails? I think using `createDirectories` or `copyDirectory` would abort if a file causes an exception.

PR #929, opened 2018-05-30
Fix SheepDyeWoolEvent cancellation visual glitch

LuminousLoreBloom:
There appears to be a client-server desync which causes the client to think the sheep is dyed even when the server has not told the client so, and is only returned to normal until relogging. The updated behaviour sends the original metadata immediately, and the sheep appears as before dying the wool was attempted.

### DIFF
--- a/src/main/java/net/glowstone/entity/passive/GlowSheep.java
+++ b/src/main/java/net/glowstone/entity/passive/GlowSheep.java
@@ -1,8 +1,6 @@
 package net.glowstone.entity.passive;
 
 import com.google.common.collect.Sets;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ThreadLocalRandom;
 import lombok.Getter;
@@ -10,8 +8,6 @@ import net.glowstone.EventFactory;
 import net.glowstone.entity.GlowAnimal;
 import net.glowstone.entity.GlowPlayer;
 import net.glowstone.entity.meta.MetadataIndex;
-import net.glowstone.entity.meta.MetadataMap;
-import net.glowstone.net.message.play.entity.EntityMetadataMessage;
 import net.glowstone.net.message.play.player.InteractEntityMessage;
 import net.glowstone.util.InventoryUtil;
 import org.bukkit.DyeColor;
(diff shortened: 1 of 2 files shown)


LuminousLoreBloom:
Changed Status to In Progress as the held item count isn't resynced back to client.

LuminousLoreBloom:
This PR is now ready

src/main/java/net/glowstone/entity/passive/GlowSheep.java
```diff
@@ -123,8 +127,15 @@ public boolean entityInteract(GlowPlayer player, InteractEntityMessage message)
...
                     DyeColor color = dye.getColor();
 
-                    SheepDyeWoolEvent dyeEvent = new SheepDyeWoolEvent(this, color);
+                    SheepDyeWoolEvent dyeEvent = EventFactory.getInstance().callEvent(
+                            new SheepDyeWoolEvent(this, color));
                     if (dyeEvent.isCancelled()) {
+                        List<MetadataMap.Entry> entryList = new ArrayList<>();
+                        entryList.add(new MetadataMap.Entry(
+                                MetadataIndex.SHEEP_DATA, getColorByte()));
+                        player.getSession().send(new EntityMetadataMessage(
```
aramperes: This works, but I would rather have the metadata updated forcefully through the MetadataMap. The metadata is sent to the client by `createUpdateMessage`, which picks up changes inside the map.

What I would do is overload the [`MetadataMap#set`](https://github.com/GlowstoneMC/Glowstone/blob/dev/src/main/java/net/glowstone/entity/meta/MetadataMap.java#L44) method to add a `force` param, which would add the needed field inside the change list even if the [value hasn't changed](https://github.com/GlowstoneMC/Glowstone/blob/dev/src/main/java/net/glowstone/entity/meta/MetadataMap.java#L75-L77).

Let me know if you don't want to do this inside your PR, I will do these changes after your PR is merged otherwise.

LuminousLoreBloom: I see what you mean and I do see the advantages of your suggestion over the former. I have gone ahead and made the change.

aramperes (approved):

aramperes:

PR #930, opened 2018-05-31
Prevent inlining of constants

aki-ks:
The java compiler inlines uses of constants. That can cause bugs.

Here's an example:
``` java
System.out.println("This is a minecraft " + GlowServer.GAME_VERSION + " server.");
```
This code does not print the game version of the running server. It prints the version of the server jar-file it was compiled against.

A similar example can be seen [here](https://github.com/GlowstoneMC/ProtocolLib/blob/7dc822cadeb2b927cae77cb5066b7b2effbaae96/modules/API/src/main/java/com/comphenix/protocol/utility/PlatformUtil.java#L26)

When calling an identity method (e.g. `NoInline.of(...)`), javac recognizes that the value is not constant and will not inline uses of that constant.

This must be done with all constants that might/will change at some point in time (e.g. `GlowServer.GAME_VERSION` or `GlowServer.PROTOCOL_VERSION`).

If there are more such constant, please apply the change for those too.

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -143,6 +143,7 @@ import net.glowstone.util.CompatibilityBundle;
 import net.glowstone.util.GlowHelpMap;
 import net.glowstone.util.GlowServerIcon;
 import net.glowstone.util.GlowUnsafeValues;
+import net.glowstone.util.NoInline;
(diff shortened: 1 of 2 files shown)


mastercoms (approved):

mastercoms:

Side note, in the future you may want to make a new branch for PRs instead of commiting directly to your repo's `dev` branch.

PR #931, opened 2018-05-31
Fire PlayerToggleFlightEvent (fixes part of #922)

jhg023:
I have modified existing code to fire a `PlayerToggleFlightEvent` whenever `GlowPlayer#setFlying` is called **and** `flying` is toggled (i.e. calling `setFlying(true)` twice when `flying` was initially set to `false` will not fire the event twice).

This modification was tested with a simple plugin that monitored `PlayerToggleFlightEvent` and sent a message to the `Player` when they toggled their flight.

I also noticed that you were initially checking `GlowPlayer#getAllowFlight` within `PlayerAbilitiesHandler.java`, but `GlowPlayer#setFlying` already performs that check, so I removed it.

This is my first pull-request, so please let me know if there's anything else I should do.

Thank you!

 - Jacob G.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -1644,11 +1644,13 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
 
     @Override
     public void setFlying(boolean value) {
-        if (flying != (flying = value && canFly)) {
-            EventFactory.getInstance().callEvent(
-                    new PlayerToggleFlightEvent(this, flying));
+        boolean flying = value && canFly;
+        PlayerToggleFlightEvent event = EventFactory.getInstance().callEvent(
+                new PlayerToggleFlightEvent(this, flying));
+        if (!event.isCancelled()) {
+            this.flying = flying;
+            sendAbilities();
         }
-        sendAbilities();
     }
 
     @Override
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -175,7 +175,6 @@ import org.bukkit.event.player.PlayerResourcePackStatusEvent;
 import org.bukkit.event.player.PlayerRespawnEvent;
 import org.bukkit.event.player.PlayerTeleportEvent;
 import org.bukkit.event.player.PlayerTeleportEvent.TeleportCause;
-import org.bukkit.event.player.PlayerToggleFlightEvent;
 import org.bukkit.event.player.PlayerToggleSneakEvent;
 import org.bukkit.event.player.PlayerToggleSprintEvent;
 import org.bukkit.event.player.PlayerUnregisterChannelEvent;
@@ -1644,13 +1643,8 @@ public class GlowPlayer extends GlowHumanEntity implements Player {
 
     @Override
     public void setFlying(boolean value) {
-        boolean flying = value && canFly;
-        PlayerToggleFlightEvent event = EventFactory.getInstance().callEvent(
-                new PlayerToggleFlightEvent(this, flying));
-        if (!event.isCancelled()) {
-            this.flying = flying;
-            sendAbilities();
-        }
+        flying = value && canFly;
+        sendAbilities();
     }
 
     @Override
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -175,6 +175,7 @@ import org.bukkit.event.player.PlayerResourcePackStatusEvent;
 import org.bukkit.event.player.PlayerRespawnEvent;
 import org.bukkit.event.player.PlayerTeleportEvent;
 import org.bukkit.event.player.PlayerTeleportEvent.TeleportCause;
+import org.bukkit.event.player.PlayerToggleFlightEvent;
 import org.bukkit.event.player.PlayerToggleSneakEvent;
 import org.bukkit.event.player.PlayerToggleSprintEvent;
 import org.bukkit.event.player.PlayerUnregisterChannelEvent;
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1643,7 +1644,10 @@ public void setAllowFlight(boolean flight) {
 
     @Override
     public void setFlying(boolean value) {
-        flying = value && canFly;
+        if (flying != (flying = value && canFly)) {
+            EventFactory.getInstance().callEvent(
+                    new PlayerToggleFlightEvent(this, flying));
```
aramperes: This doesn't check if the event is cancelled. The `flying` field should only be updated after the event is fired, and only if the event wasn't cancelled.

aramperes (changes requested):

src/main/java/net/glowstone/entity/GlowPlayer.java
jhg023: Thanks for the quick response; that makes sense, I'll fix that up now.

jhg023:
I've performed the requested changes; please let me know if further changes are required.  I wasn't sure whether or not to set `flying` to `event.isFlying()`, but I was able to save a method call by storing the value locally.

aramperes (changes requested):
I think it would be better to fire the event inside the `PlayerAbilitiesHandler`. I don't think the event should be fired in the case where a plugin calls the `setFlying` method.

jhg023:
@momothereal I was just editing my post to say something similar.  Should the event fire even if the player is not allowed to fly (`canFly` is set to `false`)?

aramperes:
I wouldn't think so, no need to fire the event if `canFly` is false. But calling `setFlying` could be necessary to sync the client.

jhg023:
Thanks, that's what I thought.  But I suspect that we'll still need the check for `canFly` inside `setFlying` in the case that a plugin wants to call that method, so I won't modify `GlowPlayer` whatsoever.

src/main/java/net/glowstone/net/handler/play/player/PlayerAbilitiesHandler.java
```diff
@@ -14,8 +16,14 @@ public void handle(GlowSession session, PlayerAbilitiesMessage message) {
         // other values should match what we've sent in the past but are ignored here
 
         GlowPlayer player = session.getPlayer();
+        if (!player.getAllowFlight()) {
```
aramperes: Hm. This would prevent the player from disabling flight if they aren't allowed to fly in the first place.

This packet could also contain other information than just flight status, so returning here would break that.

I think the logic should look something like this:

1. If the flying flag is the same as the current flight status, don't do anything.
2. If the flying flag is different, and
   i. The flying flag is set to *true*, and the player is allowed to fly, then fire the event
   ii. The flying flag is set to *false*, fire the event
3. Check if the event is cancelled. If
   i. the event is cancelled, send the abilities to the player and return (sync)
   ii. the event is not cancelled, call the setFlying method

src/main/java/net/glowstone/net/handler/play/player/PlayerAbilitiesHandler.java
```diff
@@ -15,7 +17,13 @@ public void handle(GlowSession session, PlayerAbilitiesMessage message) {
 
         GlowPlayer player = session.getPlayer();
         boolean flying = (message.getFlags() & 0x02) != 0;
```
aramperes: Looks good, however rename this local variable to `flyingFlag` to avoid confusion.

src/main/java/net/glowstone/net/handler/play/player/PlayerAbilitiesHandler.java
```diff
@@ -15,7 +17,13 @@ public void handle(GlowSession session, PlayerAbilitiesMessage message) {
...
-
-        player.setFlying(player.getAllowFlight() && flying);
+        boolean isFlying = player.isFlying();
```
aramperes: Add a comment such as `// the current flying state`

src/main/java/net/glowstone/net/handler/play/player/PlayerAbilitiesHandler.java
```diff
@@ -15,7 +17,13 @@ public void handle(GlowSession session, PlayerAbilitiesMessage message) {
...
+        if (isFlying != flying) {
+            if (!flying || player.getAllowFlight()) {
```
momothereal: Add a comment describing the logic for this if-statement.

e.g. `// fire the event if either the player disabled flying, or enabled it and is allowed to fly`

src/main/java/net/glowstone/net/handler/play/player/PlayerAbilitiesHandler.java
```diff
@@ -15,7 +18,20 @@ public void handle(GlowSession session, PlayerAbilitiesMessage message) {
...
-
-        player.setFlying(player.getAllowFlight() && flying);
+        boolean isFlying = player.isFlying();
+        boolean canFly = player.getAllowFlight();
+        if (isFlying != flying) {
+            if (!flying || canFly) {
+                PlayerToggleFlightEvent event = EventFactory.getInstance()
+                        .callEvent(new PlayerToggleFlightEvent(player, flying));
+                boolean creative = player.getGameMode() == GameMode.CREATIVE;
+                int flags = (creative ? 8 : 0) | (canFly ? 4 : 0)
```
aramperes: Use the `sendAbilities` method inside `GlowPlayer` to avoid duplicating logic/code.

If the method is private, consider making an utility method for it. Could be in `GlowServer` for example (`sendPlayerAbilities(GlowPlayer)`)

src/main/java/net/glowstone/net/handler/play/player/PlayerAbilitiesHandler.java
```diff
@@ -15,7 +18,20 @@ public void handle(GlowSession session, PlayerAbilitiesMessage message) {
 
         GlowPlayer player = session.getPlayer();
         boolean flying = (message.getFlags() & 0x02) != 0;
-
-        player.setFlying(player.getAllowFlight() && flying);
+        boolean isFlying = player.isFlying();
+        boolean canFly = player.getAllowFlight();
+        if (isFlying != flying) {
+            if (!flying || canFly) {
```
aramperes: Could you describe the logic in a comment? i.e. Fire the event if either the player disables flight, or enables it and is allowed to do so.

aramperes (changes requested):

aramperes (approved):

PR #933, opened 2018-06-01
Implement ProjectileLaunchEvent

LuminousLoreBloom:
This PR implements the ProjectileLaunchEvent.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -75,6 +75,7 @@ import org.bukkit.event.entity.EntityDeathEvent;
 import org.bukkit.event.entity.EntityToggleGlideEvent;
 import org.bukkit.event.entity.PlayerDeathEvent;
 import org.bukkit.event.entity.PlayerLeashEntityEvent;
+import org.bukkit.event.entity.ProjectileLaunchEvent;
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -798,7 +801,22 @@ public boolean isWithinSolidBlock() {
...
+            projectile = EntityStorage.create(type, location);
+        } else {
+            throw new IllegalArgumentException("Could not find projectile of type '"
+                    + type.getSimpleName() + "'");
+        }
+
+        ProjectileLaunchEvent launchEvent = EventFactory.getInstance()
+                .callEvent(new ProjectileLaunchEvent(projectile));
+        if (launchEvent.isCancelled()) {
+            projectile.remove();;
```
jhg023: Spotted that double semicolon ;)

LuminousLoreBloom:
I've rolled back the changes I originally intended to make only implementing the bare necessity for #922 .

My original thought process was to make it so the entity didn't have to be added for the world (sending spawn entity packets to all players) only for it to be removed a tick later (sending destroy entity packets to all players), but this requires a bigger rewrite of some of the spawning code than I thought and therefore I think it is best suited for another PR. 

This is ready now though.

Pr0methean (approved):

PR #935, opened 2018-06-01
Jline3 and other console improvements

mastercoms:
This moves Glowstone to jline3. It comes with a better formatted console, compatibility improvements and more.

In addition to jline3, there are also new console manager tasks in addition to the already present command task: evaluation tasks and console tasks. ~~`EvalTask`s parse Java code from the console and~~ `ConsoleTask`s add new ways to customize the console through binds (not implemented), settings, keymaps and widgets.

Co-authored-by: Pr0methean <4961925+Pr0methean@users.noreply.github.com>


src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -35,27 +32,28 @@
...
- * based on CraftBukkit.
+ * Handles all logging and input-related console improvements.
  */
 public final class ConsoleManager {
 
-    private static final Logger logger = Logger.getLogger("");
     @NonNls private static String CONSOLE_DATE = "HH:mm:ss";
     @NonNls private static String FILE_DATE = "yyyy/MM/dd HH:mm:ss";
-    @NonNls private static String CONSOLE_PROMPT = ">";
+    @NonNls private static String CONSOLE_PROMPT = "> "; // TODO: fix prompt
```
Pr0methean: What's the problem here? Document in detail or link a bug.

src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -64,8 +62,38 @@
...
+        addReplacement(ChatColor.WHITE, "\u001B[0;37;1m"); // NON-NLS
+        addReplacement(ChatColor.MAGIC, "\u001B[5m"); // NON-NLS
+        addReplacement(ChatColor.BOLD, "\u001B[21m"); // NON-NLS
+        addReplacement(ChatColor.STRIKETHROUGH, "\u001B[9m"); // NON-NLS
+        addReplacement(ChatColor.UNDERLINE, "\u001B[4m"); // NON-NLS
+        addReplacement(ChatColor.ITALIC, "\u001B[3m"); // NON-NLS
+        addReplacement(ChatColor.RESET, "\u001B[39;0m"); // NON-NLS
+    }
+
+    private static void addReplacement(ChatColor formatting, String ansi) {
```
Pr0methean: Make Ansi @NonNls, and then the // NON-NLS comments probably won't be needed.

src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -179,62 +151,64 @@ private void setupColors() {
...
+            } else {
+                builder.append(formatMessage(record));
+            }
+            builder.append('\n');
 
-        public LoggerOutputStream(Level level) {
-            this.level = level;
-        }
+            if (record.getThrown() != null) {
+                // StringWriter's close() is trivial
```
Pr0methean: Use try-with-resources anyway, for future-proofing. It's probably equally concise.

src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -179,62 +151,64 @@ private void setupColors() {
...
-            if (!record.isEmpty() && !record.equals(separator)) {
-                logger.logp(level, "LoggerOutputStream", "log" + level, record); // NON-NLS
+    protected String colorize(String string) {
+        if (string.indexOf(ChatColor.COLOR_CHAR) < 0) {
+            return string; // no colors in the message
+        }
+        for (ChatColor color : colors) {
+            if (this.color && replacements.containsKey(color)) {
+                string = string.replaceAll("(?i)" + color, replacements.get(color));
+            } else {
```
Pr0methean: Just use the ternary operator here.

src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -296,46 +270,57 @@ public synchronized void flush() {
...
+            List<String> completions = null;
             try {
-                List<String> completions = server.getScheduler()
-                        .syncIfNeeded(() -> server.getCommandMap().tabComplete(sender, buffer));
-                if (completions == null) {
-                    return cursor;  // no completions
-                }
-                candidates.addAll(completions);
+                completions = server.getScheduler().syncIfNeeded(() -> server.getCommandMap()
+                    .tabComplete(sender, line.line()));
```
Pr0methean: Keep the lambda on one line.

src/main/java/net/glowstone/i18n/LocalizedStrings.java
```diff
@@ -6,233 +6,260 @@
...
                 );
             }
 
-            LoggableLocalizedString LOOTING_MANAGER = new LoggableLocalizedStringImpl(
+            LoggedLocalizedString LOOTING_MANAGER = new LoggedLocalizedString(
                     "console.looting-manager.load-failed", Level.SEVERE
             );
 
+            interface Manager {
+                LoggedLocalizedString COMMAND = new LoggedLocalizedString(
```
Pr0methean: This belongs under Console.Warn.

src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -296,46 +270,57 @@ public synchronized void flush() {
...
                     }
-
-                    if (command == null || command.trim().isEmpty()) {
-                        continue;
+                    if (command != null && !(command = command.trim()).isEmpty()) {
+                        reader.getTerminal().writer().println(colorize(ChatColor.RESET
+                            + "====" + ChatColor.GOLD + "g>" + ChatColor.RESET + '"' + command
+                                + '"')); // NON-NLS
+                        if (command.startsWith("$")) {  // NON-NLS
+                            server.getScheduler().runTask(null,
```
Pr0methean: Document what this and the following branch do.

src/main/java/net/glowstone/util/compiler/EvalTask.java
```diff
@@ -0,0 +1,185 @@
...
+import javax.tools.ToolProvider;
+import net.glowstone.GlowServer;
+import net.glowstone.ServerProvider;
+import net.glowstone.util.ReflectionProcessor;
+import org.bukkit.ChatColor;
+
+/**
+ * A task for which a source string must be compiled and ran to complete.
+ */
+public class EvalTask implements Runnable {
```
Pr0methean: Compilation needs to be cached.

src/main/java/net/glowstone/util/compiler/EvalTask.java
```diff
@@ -0,0 +1,185 @@
...
+
+        List<JavaFileObject> compilationUnit = new ArrayList<>();
+        compilationUnit.add(new JavaSource("REPLShell", source));
+
+        JavaCompiler.CompilationTask task = compiler.getTask(null, classDataFileManager,
+            null, null, null, compilationUnit);
+        if (task.call()) {
+            try {
+                Object returned = MethodInvocationUtils
+                    .invokeStaticMethod(getCompiledClass("REPLShell"), "run",
```
Pr0methean: Move the methods of MethodInvocationUtils to this class, unless and until another class uses them.

src/main/java/net/glowstone/ConsoleManager.java
```diff
@@ -296,46 +270,57 @@ public synchronized void flush() {
...
+                                new EvalTask(command.substring(1), command.startsWith("$$")));
```
Pr0methean: Add a config setting that disables this mode. Disable it by default, unless that would break mod compatibility, and maybe even then. That way, gaining control of the console doesn't allow arbitrary code execution if the legit admin doesn't use it.

src/main/java/net/glowstone/i18n/LogRecordableLocalizedString.java
```diff
@@ -0,0 +1,14 @@
+package net.glowstone.i18n;
```
Pr0methean: Why not just add these methods to LoggableLocalizedString instead?

src/main/java/net/glowstone/util/compiler/EvalTask.java
```diff
@@ -0,0 +1,185 @@
...
+import java.io.OutputStream;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import javax.tools.FileObject;
+import javax.tools.ForwardingJavaFileManager;
+import javax.tools.JavaCompiler;
```
Pr0methean: Can we build this as a nested jar or something, and avoid trying to load it on JVMs that don't have the `java.compiler` module (JDK 9+) or are the compact profiles (JDK 8), rather than breaking compatibility with those JVMs?

src/main/java/net/glowstone/util/compiler/EvalTask.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        @Override
+        public JavaFileObject getJavaFileForOutput(Location location, String className,
+            JavaFileObject.Kind kind, FileObject sibling) throws IOException {
+            return new JavaClass(className);
+        }
+    }
+
+    public class ClassDataOutputStream extends OutputStream {
```
Pr0methean: What does this do differently than the underlying ByteArrayOutputStream? If this can't simply be replaced by a BAOS, document why not, and can we use a FilterOutputStream instead?

Pr0methean (changes requested):

Pr0methean:
@mastercoms asked me to commit the changes I could. I've done that; the REPL implementation is now on branch `jline3repl`.

Pr0methean (approved):

VaiTon:
Why is this stalled?

mastercoms:
Unfortunately, we encountered regressions in the terminal on Windows. I haven't tested the latest version to see if it was fixed.

PR #941, opened 2018-06-04
Fire EntityRemoveFromWorldEvent (fixes part of #922)

jhg023:
The `EntityRemoveFromWorldEvent` is now fired when `GlowEntity#remove` is called.

This pull request also moves `leaveVehicle()` from `GlowLivingEntity#setHealth` (I added it in #939) to `GlowEntity#remove`, as it makes more sense for it to be there.

This has been tested and works accordingly.

Thanks!

### DIFF
--- a/src/main/java/net/glowstone/entity/EntityManager.java
+++ b/src/main/java/net/glowstone/entity/EntityManager.java
@@ -2,6 +2,7 @@ package net.glowstone.entity;
 
 import static com.google.common.collect.Multimaps.newSetMultimap;
 
+import com.destroystokyo.paper.event.entity.EntityRemoveFromWorldEvent;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/entity/EntityManager.java
```diff
@@ -93,6 +96,7 @@ void register(GlowEntity entity) {
      * @param entity The entity.
      */
     void unregister(GlowEntity entity) {
+        EventFactory.getInstance().callEvent(new EntityRemoveFromWorldEvent(entity));
```
Pr0methean, line 99: Shouldn't this be checked for cancellation?

Pr0methean (changes requested):

src/main/java/net/glowstone/entity/EntityManager.java
jhg023: Neither the event nor any of its superclasses implement `Cancellable`

Pr0methean (approved):

PR #944, opened 2018-06-05
Fire EntityTeleportEvent (fixes part of #922)

jhg023:
The `EntityTeleportEvent` is now fired when a `GlowEntity` that is **not** a `GlowPlayer` attempts to teleport, as stated by its [documentation](https://glowstone.net/jd/glowkit/org/bukkit/event/entity/EntityTeleportEvent.html).

This has **not** been tested, as Enderman teleportation has not yet been implemented, but a `TODO` comment has been added for whenever that occurs.

Thanks!


Pr0methean:
Looks good, but needs to pass on CircleCI.

Pr0methean:
@momothereal This currently fires an EntityTeleportEvent inside each entity's constructor. Is that the correct behavior?

jhg023:
I think this event should be excluded from the entity's constructor.  Assuming some user wants to disable Enderman teleportation, cancelling this event would lead to the Enderman never spawning in the first place, right?

mastercoms (approved):

PR #948, opened 2018-06-27
Fixes #947

### LINKED ISSUE #947: Bed bug in inventory. (no pun)
I found an area of beds and decided to excavate them. But I noticed some unusual behaviour. They were breaking into half-blocks, and giving me 2 beds. Also they went into illegal item slots:

Here's pictures of half-beds:

Edit:

I found a bug with animal behaviour, if you punch it, they just get moved backwards:

And with item physics:

VaiTon:

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockBed.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockBed.java
@@ -92,6 +92,16 @@ public class BlockBed extends BlockType {
         }
     }
 
+    @Override
+    public void onNearBlockChanged(GlowBlock block, BlockFace face, GlowBlock changedBlock,
+                                   Material oldType, byte oldData, Material newType, byte newData) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/block/blocktype/BlockBed.java
```diff
@@ -92,6 +92,16 @@ public static GlowBlock getFoot(GlowBlock block) {
         }
     }
 
+    @Override
+    public void onNearBlockChanged(GlowBlock block, BlockFace face, GlowBlock changedBlock,
+                                   Material oldType, byte oldData, Material newType, byte newData) {
+        if (changedBlock.equals(getHead(block)) || changedBlock.equals(getFoot(block))) {
+            if (newType ==  Material.AIR) {
```
mastercoms: There's an extra space here between the `==` and `Material.AIR`.

mastercoms (approved):

PR #949, opened 2018-06-27
Fix bed drop color

aki-ks:
Destroyed Beds (no matter what color they had) did always drop in white.
They will now drop in the correct color.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockBed.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockBed.java
@@ -92,9 +92,8 @@ public class BlockBed extends BlockType {
 
     @Override
     public Collection<ItemStack> getDrops(GlowBlock block, ItemStack tool) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/block/blocktype/BlockBed.java
```diff
@@ -92,6 +90,13 @@ public static GlowBlock getFoot(GlowBlock block) {
         }
     }
 
+    @Override
+    public Collection<ItemStack> getDrops(GlowBlock block, ItemStack tool) {
+        GlowBed bed = (GlowBed)block.getState();
```
mastercoms: Add a space after the cast.

Pr0methean: Done in 754eced (drive-by).

mastercoms (approved):

PR #951, opened 2018-07-02
Implement coordinate-based teleporting

aki-ks:


aki-ks:
Why is there a /tp and a /teleport command?
The /tp command has the same (and more) capabilities than the /teleport command.
Should I turn the /teleport command into a synonym for the /tp command?

aramperes:
@aki-ks No, these are two separate commands:

https://minecraft.gamepedia.com/Commands/tp

https://minecraft.gamepedia.com/Commands/teleport

aki-ks:
"In Bedrock Edition and Java Edition 1.13, /tp is simply an alias for /teleport." (https://minecraft.gamepedia.com/Commands/teleport)

mastercoms:
We're currently targeting 1.12.2. If you'd like to implement a 1.13 feature, then we would have to hold this PR for when we start the 1.13 branch, which won't be happening until PaperSpigot releases their 1.13 API.

MatrixTunnel:
The original /tp functionality was removed and replaced with /teleport in 1.13. They were separate in 1.12

PR #961, opened 2018-08-17
Use FastUUID for UuidUtils

aramperes:
Replaces UuidUtils with [FastUUID](https://github.com/jchambers/fast-uuid/blob/master/src/main/java/com/eatthepath/uuid/FastUUID.java).

very fast

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,6 +85,11 @@
       <artifactId>flow-network</artifactId>
       <version>1.2.4-SNAPSHOT</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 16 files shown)


kashike:
Why not just pull it in as a dependency?

aramperes:
I feel like it makes more sense to just integrate it directly, as it's a 1-file dependency, and has the same license?

kashike:
I prefer to not copy things in like this myself - it makes more sense to use an already established tiny dependency and shadow it into the JAR.

mastercoms (approved):

PR #967, opened 2018-10-02
[i18n] Localizable map for GameMode decoding

Pr0methean:
This is the proof of concept for a new approach to localizing game-mode names and abbreviations. It uses a second resource bundle, which is converted during class init to an `ImmutableSortedMap<String,GameMode>` with the same keys.


mastercoms:
Seems like your import clean up in the main i18n branch removed some necessary imports for these changes.

mastercoms (approved):

PR #969, opened 2018-10-12
[WIP] Fixed Explosions

yehan2002:
- Fixed explosion radius
-  Exclude players in creative mode and spectator mode from explosions

### DIFF
--- a/src/main/java/net/glowstone/Explosion.java
+++ b/src/main/java/net/glowstone/Explosion.java
@@ -166,7 +166,7 @@ public final class Explosion {
 
         damageEntities();
         Collection<GlowPlayer> affectedPlayers
-                = collectPlayersInRadius(EXPLOSION_VISIBILITY_RADIUS * this.power);
+                = collectPlayersInRadius(EXPLOSION_VISIBILITY_RADIUS * (int)this.power);
         for (GlowPlayer player : affectedPlayers) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/Explosion.java
```diff
@@ -165,9 +166,12 @@ public boolean explodeWithEvent() {
 
         damageEntities();
         Collection<GlowPlayer> affectedPlayers
-                = collectPlayersInRadius(EXPLOSION_VISIBILITY_RADIUS);
+                = collectPlayersInRadius(EXPLOSION_VISIBILITY_RADIUS * (int)this.power);
         for (GlowPlayer player : affectedPlayers) {
-            playOutExplosion(player, blocks);
+            if (player.getGameMode() != GameMode.CREATIVE
+                    & player.getGameMode() != GameMode.SPECTATOR) {
```
Pr0methean: Use a switch statement here. If `getGameMode()` ever becomes nontrivial, we'll want to call it only once, and a switch is more elegant than extracting into a variable.

Pr0methean (changes requested):

mastercoms:
Why would we not send an explosion message to the client if they are in creative or spectator mode?

mastercoms:
Could you please address the above comment?

PR #972, opened 2018-11-05
Fix compilation error.

cedws:
This PR fixes a compilation error caused by a method signature change in `flow-network`, updated in commit c6e1ed01170598006ca16dc3cebd08723d20b3eb.

This PR does not pass the CheckStyle rules. I spent a few hours trying to get CheckStyle to apply the style rules but gave up out of frustration.

### DIFF
--- a/src/main/java/net/glowstone/net/pipeline/CodecsHandler.java
+++ b/src/main/java/net/glowstone/net/pipeline/CodecsHandler.java
@@ -40,7 +40,7 @@ public final class CodecsHandler extends MessageToMessageCodec<ByteBuf, Message>
 
         // write body
         ByteBuf messageBuf = ctx.alloc().buffer();
-        messageBuf = reg.getCodec().encode(messageBuf, msg);
+        messageBuf = reg.getCodec().encode(null, messageBuf, msg);
 
         out.add(Unpooled.wrappedBuffer(headerBuf, messageBuf));
     }
@@ -52,7 +52,7 @@ public final class CodecsHandler extends MessageToMessageCodec<ByteBuf, Message>
         Codec<?> codec = protocol.newReadHeader(msg);
 
         // read body
-        Message decoded = codec.decode(msg);
+        Message decoded = codec.decode(null, msg);
         if (msg.readableBytes() > 0) {
             ConsoleMessages.Warn.Net.MESSAGE_TOO_LONG.log(msg.readableBytes(), decoded);
         }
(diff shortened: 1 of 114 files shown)


cedws:
I would be okay with signing the CLA, but the service requires access to see my private email address. Let me know if you have any workarounds for this.

src/main/java/net/glowstone/net/pipeline/CodecsHandler.java
```diff
@@ -40,7 +40,7 @@ protected void encode(ChannelHandlerContext ctx, Message msg, List<Object> out)
 
         // write body
         ByteBuf messageBuf = ctx.alloc().buffer();
-        messageBuf = reg.getCodec().encode(messageBuf, msg);
+        messageBuf = reg.getCodec().encode(null, messageBuf, msg);
```
aramperes, line 43: We were thinking of reverting the changes in flow-network because of this. In retrospect, it doesn't make much sense for Codecs to have access to the session or context (that job should be left to the MessageHandlers). Passing `null` simply bypasses this new feature, which we aren't sure we need anyway.

mastercoms:
@c-edw Please just state that you, c-edw, have read and agree to the Glowstone CLA. You won't need to use the tool then.

But yeah, we might be reverting this change to flow-network, so this PR might be rejected because of that.

Finally, the reason why the check style is not passing is because the lines are too long. You would have to add new lines after a comma for method parameters to fix it for this PR.

src/main/java/net/glowstone/net/pipeline/CodecsHandler.java
cedws: @momothereal I see. I assume that's why you left it on `1.2.5`, while things are discussed over on `flow-network`. If those changes are reverted, I don't think this PR is needed.

aramperes: Yeah, I updated the dependency without really checking if it had backwards-incompatible changes :) We've all been pretty busy which is why it's been left like this.

If you are still interested, feel free to apply a revert to https://github.com/GlowstoneMC/network/pull/1 on the `GlowstoneMC/network` repo. We'll then publish a release and update the Glowstone repo accordingly.

cedws:
*I have read and agree to the Glowstone CLA.*

Yup, that's fine. I would humbly like to suggest keeping `dev` in a state that passes CI. I wasn't sure if nobody had the time to fix it, or just didn't notice it.

Thanks guys.

PR #973, opened 2018-11-05
Fix invalid bed placement.

### LINKED ISSUE #950: Bed removed from inventory and not placed
Glowstone build: `f8e2fbcf15419a465fe742f40ff6423514a9412b`

When you try to place a bed where it can't be placed (for example if head block is occupied by another block or on a side of a block) it just disappear from player inventory as it has been placed but nothing changes in world.

cedws:
Fixes #950. Feedback is appreciated. I haven't accepted the CLA from the bot for privacy reasons but have been told that a declaration here is fine.

*I have read and agree to the Glowstone CLA.*

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -164,6 +164,18 @@ public class BlockType extends ItemType {
     }
 
     /**
+     * Check whether the block can be placed by a given player at a given location.
+     *
+     * @param player The player who placed the block.
+     * @param block The location the block is being placed at.
+     * @param against The face the block is being placed against.
+     * @return Whether the placement is valid.
+     */
+    public boolean canPlaceAt(GlowPlayer player, GlowBlock block, BlockFace against) {
+        return canPlaceAt(block, against);
+    }
+
+    /**
      * Called when a block is placed to calculate what the block will become.
      *
      * @param player the player who placed the block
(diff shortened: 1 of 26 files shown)


src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -163,6 +163,18 @@ public boolean canPlaceAt(GlowBlock block, BlockFace against) {
...
 
+    /**
+     * Check whether the block can be placed by a given player at a given location.
+     *
+     * @param player The player who placed the block.
+     * @param block The location the block is being placed at.
+     * @param against The face the block is being placed against.
+     * @return Whether the placement is valid.
+     */
+    public boolean canPlaceAt(GlowPlayer player, GlowBlock block, BlockFace against) {
```
LuminousLoreBloom, line 164: Redundant player arg?

cedws: It isn't redundant. It's an overload method.

LuminousLoreBloom: Ah. Why not add the param to the existing method instead, changing all the Block subclass override method and keep the one with lesser args as the overload?

aramperes (approved):

src/main/java/net/glowstone/block/blocktype/BlockType.java
mastercoms: I agree, it would make more sense to make all pathways leads to the longer method, just so the bed won't mess up in new places.

cedws: Are you suggesting to remove the overload and just change the parameters of the original method?

mastercoms: Change the parameters of the original method, but add an overload that has the original parameters, with a null player.

cedws: OK, I think I understand what you mean now. The reason I didn't do it like that is because most classes extending BlockType don't need GlowPlayer to be passed to `canPlaceAt`. It would be a huge diff to change the original method and all occurrences, so I wanted to avoid that too.

mastercoms: You wouldn't be changing all occurrences. You would just be making it so when something calls `canPlaceAt(GlowBlock, BlockFace)`, it returns the result of `canPlaceAt(GlowPlayer, GlowBlock, BlockFace)`. What you're doing now makes it so that there are two different logic paths for the same method name, which definitely breaks contracts.

PR #976, opened 2018-11-13
Fixes #847

### LINKED ISSUE #847: Boat item isn't consumed on the server-side.

j1o1h1n:
This fix builds on the earlier work by Pr0methean [fb5cc7b].  The issue was that `UseItemHandler` restored the original item into the inventory.  I guess this was done in order to delete items when they are consumed.

I made a similar fix to `BlockPlacementHandler`.

### DIFF
--- a/src/test/java/net/glowstone/net/handler/play/player/UseItemHandlerTest.java
+++ b/src/test/java/net/glowstone/net/handler/play/player/UseItemHandlerTest.java
@@ -1,0 +1,96 @@
+package net.glowstone.net.handler.play.player;
+
+import net.glowstone.EventFactory;
+import net.glowstone.GlowServer;
+import net.glowstone.GlowWorld;
+import net.glowstone.block.GlowBlock;
+import net.glowstone.entity.EntityIdManager;
+import net.glowstone.entity.EntityManager;
+import net.glowstone.entity.GlowPlayer;
+import net.glowstone.entity.objects.GlowBoat;
+import net.glowstone.inventory.GlowPlayerInventory;
+import net.glowstone.net.GlowSession;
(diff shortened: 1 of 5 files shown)


src/test/java/net/glowstone/net/handler/play/player/UseItemHandlerTest.java
```diff
@@ -0,0 +1,116 @@
...
+    private EntityIdManager entityIdManager;
+    private GlowSession session;
+    private GlowPlayer player;
+    private GlowBlock emptyBlock;
+    private GlowBlock targetBlock;
+    private EventFactory eventFactory;
+
+    @Before
+    public void setupMocks() {
+        // cache the event factory
```
Pr0methean, line 47: I'm pretty sure some of this is redundant to the superclass's `@Before` method.

Pr0methean (changes requested):

src/test/java/net/glowstone/net/handler/play/player/UseItemHandlerTest.java
j1o1h1n: Thank you for pointing that out.  I tried to use the serverShim but was unable to, it mocks World and the test I wrote needs GlowWorld.

I have removed the unused superclass.

Pr0methean (approved):

PR #981, opened 2018-11-28
Command localization, step 1

Pr0methean:
This creates an abstract base class for commands which extend VanillaCommand and need i18n, so they will speak to players in their own locale if available. (To non-player CommandSender instances, they'll speak the server locale.)

Subsequent PRs will do the same for BukkitCommand, make all the concrete command classes extend these two, and add tests to ensure no missing or unused entries in `commands.properties`.

### DIFF
--- a/src/main/java/net/glowstone/command/minecraft/GlowVanillaCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/GlowVanillaCommand.java
@@ -1,0 +1,123 @@
+package net.glowstone.command.minecraft;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.ResourceBundle;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import net.glowstone.entity.GlowPlayer;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.command.defaults.VanillaCommand;
+import org.jetbrains.annotations.NonNls;
+
+/**
+ * A subclass of {@link VanillaCommand} with the additional feature that when the command sender is
+ * a {@link GlowPlayer}, description, usage and permission-error messages are looked up in the
+ * client's locale, temporarily overriding whatever has been or is subsequently set in
+ * {@link #setDescription(String)}, {@link #setUsage(String)} or
+ * {@link #setPermissionMessage(String)}.
+ */
+public abstract class GlowVanillaCommand extends VanillaCommand {
+
+  private static final String BUNDLE_BASE_NAME = "commands";
+  private static final String DESCRIPTION_SUFFIX = ".description";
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/command/minecraft/GlowVanillaCommand.java
```diff
@@ -0,0 +1,126 @@
...
+        if (!(sender instanceof GlowPlayer)) {
+            return innerExecute(sender, commandLabel, args);
+        }
+        localeChangeLock.lock();
+        try {
+            Locale locale = Locale.forLanguageTag(((GlowPlayer) sender).getLocale());
+            if (locale.equals(lastLoadedLocale)) {
+                return innerExecute(sender, commandLabel, args);
+            }
+            lastLoadedLocale = locale;
```
mastercoms: Last loaded locale is set to the sender's locale if it does not equal the current locale, but then after the  command is executed, the permission message, description and usage is reverted to the old locale, while last loaded locale is still set to the sender's locale.

src/main/java/net/glowstone/command/minecraft/GlowVanillaCommand.java
```diff
@@ -0,0 +1,126 @@
...
+public abstract class GlowVanillaCommand extends VanillaCommand {
+
+    private static final String BUNDLE_BASE_NAME = "commands";
+    private static final String DESCRIPTION_SUFFIX = ".description";
+    private static final String USAGE_SUFFIX = ".usage";
+    private static final String PERMISSION_SUFFIX = ".no-permission";
+    private static final ResourceBundle SERVER_LOCALE = ResourceBundle.getBundle(BUNDLE_BASE_NAME);
+
+    private final Lock localeChangeLock = new ReentrantLock();
+    private volatile Locale lastLoadedLocale;
```
mastercoms: Maybe instead of a single locale system, we could have a lazy loading multi-locale system where unloaded locales are put into a new map on demand.

aramperes (approved):

mastercoms (dismissed):

src/main/java/net/glowstone/command/minecraft/GlowVanillaCommand.java
Pr0methean: Done; I split the lookup into two Guava-cached steps: String -> ResourceBundle (static) and ResourceBundle -> CommandMessages (per command)

src/main/java/net/glowstone/command/minecraft/GlowVanillaCommand.java
Pr0methean: Fixed: we no longer update the inherited fields. Caveats about the setters now apply to all players who aren't using the server locale, which is probably simpler and more debuggable than how they worked in the first commit.

mastercoms (approved):
Great work! :)

mastercoms:

PR #982, opened 2018-11-28
Localizedcommand checkstyle fixes to GlowVanillaCommand (change to PR #981)

crashdemons:
Formats GlowVanillaCommand to fix most of the CheckStyle warnings.

There is still an error for missing javadoc and newline in the (I assume incomplete) commands.properties.
I didn't fix those because I didn't want to add false information or correct errors that indicate further work.


crashdemons:
closed as handled by other commits on that branch now.

PR #983, opened 2018-12-01
#922 - Implement the PlayerLevelChangeEvent

FlorentClarret:
Contributes to issue #922

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowPlayer.java
+++ b/src/main/java/net/glowstone/entity/GlowPlayer.java
@@ -170,6 +170,7 @@ import org.bukkit.event.player.PlayerChangedWorldEvent;
 import org.bukkit.event.player.PlayerCommandPreprocessEvent;
 import org.bukkit.event.player.PlayerDropItemEvent;
 import org.bukkit.event.player.PlayerGameModeChangeEvent;
+import org.bukkit.event.player.PlayerLevelChangeEvent;
 import org.bukkit.event.player.PlayerPortalEvent;
 import org.bukkit.event.player.PlayerRegisterChannelEvent;
 import org.bukkit.event.player.PlayerResourcePackStatusEvent;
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1677,8 +1678,14 @@ public void setWalkSpeed(float value) throws IllegalArgumentException {
 
     @Override
     public void setLevel(int level) {
-        this.level = Math.max(level, 0);
-        sendExperience();
+        int newLevel = Math.max(level, 0);
+
+        if (newLevel != this.level) {
+            EventFactory.getInstance().callEvent(
+                    new PlayerLevelChangeEvent(this, this.level, newLevel));
```
Pr0methean, line 1685: Check if the returned event `isCancelled`.

src/main/java/net/glowstone/entity/GlowPlayer.java
```diff
@@ -1697,7 +1704,8 @@ public void giveExp(int xp) {
             exp += value;
             if (exp >= 1) {
                 exp -= 1;
-                value = 1.0f / getExpToLevel(++level);
+                setLevel(level + 1);
```
Pr0methean, line 1707: This makes the XP bar empty even if the level-up was cancelled. Shouldn't it stay full in that case?

Pr0methean (changes requested):

src/main/java/net/glowstone/entity/GlowPlayer.java
FlorentClarret: I might be wrong, but the `PlayerLevelChangeEvent` does not implement the `Cancellable` interface. So I can't check any `isCancelled` method.

Pr0methean (approved):

mastercoms (approved):

PR #957, opened 2018-07-20
The Wettest Update (1.13)

mastercoms:
Glowstone is proud to announce the Update Aquatic (well, Mojang is, we're pretty panicked right now)

#### The big one
- [ ] [1.13 API](https://www.spigotmc.org/threads/bukkit-craftbukkit-spigot-bungeecord-1-13-pre7-development-builds.328883/) compatibility, [Summary](https://www.spigotmc.org/threads/minecraft-1-13-what-to-expect.288714/)
  - [ ] Bukkit [1.13 update](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/f8b2086d60942eb2cd7ac25a2a1408cb790c222c)

#### The other big one
- [ ] [1.13 Protocol compatibility](https://wiki.vg/Protocol) (#958)

#### Feature completeness
So. Many. Features. This list is WIP and will only cover major feature additions/changes. If you are a developer, feel free to edit and add more items and detail to this list. For a comprehensive list of changes, refer to [the wiki](https://minecraft.gamepedia.com/1.13).

Most features in this list are non-blocking to this PR getting merged, but they'll be nice to implement. However, there are a few things that have to be implemented or else we wouldn't really be 1.13 (mostly just the stuff in the `General` section below).

**General**
- [ ] Data packs support

Can I interrupt this list for a second to just say how _great_ data packs are going to be for Glowstone? They solve a ton of issues with us not being able to replicate game data from vanilla. Ok, let's carry on.

- [ ] Tags
- [ ] Where did our IDs go? (also known as "the flattening") (#960)

**Blocks**
- [ ] Different air types
- [ ] Stripped logs
- [ ] Stripped wood
- [ ] Turtle Egg
- [ ] New water physics behavior

**Items**
- [ ] Debug stick
- [ ] New recipes _TODO_
  - [ ] Wood block
- [ ] Trident weapon
- [ ] New spawn eggs _TODO_

**World**
- [ ] Coral reefs
- [ ] Icebergs
- [ ] Shipwrecks
- [ ] Underwater caves
- [ ] Underwater ruins
- [ ] Kelp
- [ ] Bubble columns
- [ ] Seagrass
- [ ] Sea Pickle
- [ ] Biomes (https://minecraft.gamepedia.com/Java_Edition_1.13-pre5#Changes)
  - [ ] The End
    - [ ] Small End Islands
    - [ ] End Midlands
    - [ ] End Highlands
    - [ ] End Barrens
  - [ ] Ocean
[...]

PR #985, opened 2018-12-16
Implement the substatistics system

FlorentClarret:
On my previous PR (#984), I noticed that the substatistics system was not yet implemented. In this PR, I implemented it in the GlowStatistic and in the StatisticMap. I would like to have your opinion on it.

In order to test it, I also implemented the killEntity statistic and substatistic. Everything seems to work fine.


src/main/java/net/glowstone/constants/GlowStatistic.java
```diff
@@ -127,12 +54,111 @@ private GlowStatistic() {
...
+            STATISTICS.put(stat, "stat." + UPPER_UNDERSCORE.to(LOWER_CAMEL, stat.name()));
+        }
+
+        // Specific case
+        STATISTICS.put(PLAY_ONE_TICK, "stat.playOneMinute");
+    }
+
+    private static void generateMaterialStatistics() {
+        for (Material material : Material.values()) {
+            String name = material.name().toLowerCase()
```
aramperes: I think this should make use of the `ItemIds` class, which has a registry for such IDs.

src/main/java/net/glowstone/constants/GlowStatistic.java
```diff
@@ -127,12 +54,111 @@ private GlowStatistic() {
...
+        for (EntityType entityType : EntityType.values()) {
+            Class<? extends Entity> entityClass = entityType.getEntityClass();
+
+            if (entityClass == null) {
+                continue;
+            }
+
+            if (Monster.class.isAssignableFrom(entityClass)) {
+                set(ENTITY_KILLED_BY, entityType, "entityKilledBy."
+                        + UPPER_UNDERSCORE.to(UPPER_CAMEL, entityType.name()));
```
aramperes: Why not use `EntityType#getName()`?

aramperes (changes requested):

src/main/java/net/glowstone/constants/GlowStatistic.java
FlorentClarret: I didn't want to use `getName()` because it's marked as deprecated.

src/main/java/net/glowstone/constants/GlowStatistic.java
FlorentClarret: Good call, I forgot about this class. Thanks

src/main/java/net/glowstone/constants/GlowStatistic.java
aramperes: I believe the deprecation is meant to warn plugin developers about 1.13. We use this method elsewhere in Glowstone and we are still targetting 1.12.

FlorentClarret: I modified using `getName` but some cases are not working. For example, I have ZombiePigman instead of PigZombie. I can add the specific cases after the for loop just like I did for the `PLAY_ONE_TICK`. As you wish

aramperes: That's odd... So the statistic IDs do not match Vanilla entity IDs, i.e. the ones you use in `/summon`? Or is our SummonCommand implementation broken?

FlorentClarret: I think the SummonCommand implementation is correct, it's just that the stat IDs do not match the Vanilla ones for the entities. Basically, it's just the entity class name. I might be wrong but it's working fine as it is.
I might change my implementation to avoid the `UPPER_UNDERSCORE.to(UPPER_CAMEL, entityType.name())` and switch to just `entityClass.getSimpleName()`

Pr0methean: No regression on #957 please!

mastercoms:
Just a reminder that this does not build with the latest `dev` branch merge.

FlorentClarret:
Thanks @mastercoms for the reminder, I did not see that this didn't build with the latest merge. The issue is now fixed.
If you think that there's still something wrong with this MR, let me know and I will fix it.

aramperes (approved):

PR #986, opened 2018-12-17
Implements some missing events

VaiTon:
### Events implemented until now:

- [x] ProjectileHitEvent
- [x] BlockPhysicsEvent
- [x] PlayerExpChangeEvent
- [x] BroadcastMessageEvent

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockType.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockType.java
@@ -26,7 +26,6 @@ import org.bukkit.Sound;
 import org.bukkit.block.BlockFace;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.LivingEntity;
-import org.bukkit.event.Event;
 import org.bukkit.event.block.BlockCanBuildEvent;
 import org.bukkit.event.block.BlockPhysicsEvent;
 import org.bukkit.event.block.BlockPlaceEvent;
@@ -40,10 +39,10 @@ import org.bukkit.util.Vector;
  */
 public class BlockType extends ItemType {
 
-    protected static final BlockFace[] SIDES = new BlockFace[]{BlockFace.NORTH, BlockFace.EAST,
-            BlockFace.SOUTH, BlockFace.WEST};
-    protected static final BlockFace[] ADJACENT = new BlockFace[]{BlockFace.NORTH, BlockFace.EAST,
-            BlockFace.SOUTH, BlockFace.WEST, BlockFace.UP, BlockFace.DOWN};
+    protected static final BlockFace[] SIDES = new BlockFace[] {BlockFace.NORTH, BlockFace.EAST,
+        BlockFace.SOUTH, BlockFace.WEST};
+    protected static final BlockFace[] ADJACENT = new BlockFace[] {BlockFace.NORTH, BlockFace.EAST,
+        BlockFace.SOUTH, BlockFace.WEST, BlockFace.UP, BlockFace.DOWN};
 
     protected List<ItemStack> drops;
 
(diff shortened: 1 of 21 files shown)


VaiTon:
@mastercoms i am requesting your help because i can't find the reason why the method `callEvent` on line 17 of GlowVehicle returns null while doing tests.

src/test/java/net/glowstone/entity/GlowPlayerTest.java
```diff
@@ -310,12 +312,19 @@ public void testGiveExpLevels() {
     @Test
     public void testGiveExp() {
         entity.giveExp(20);
+        boolean expEvent = true;
         verify(eventFactory).callEvent(argThat(input -> {
-            PlayerLevelChangeEvent event = (PlayerLevelChangeEvent) input;
```
Pr0methean, line 314: The test should still specify which class of message is fired, and it should have cases that create both. If this specific case creates both types of message in an unspecified order, then verify it twice, add the argument to a list in the matcher, and then run the relevant asserts against the list.

src/test/java/net/glowstone/entity/GlowPlayerTest.java
```diff
@@ -57,7 +59,7 @@
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 
-@PrepareForTest({Bukkit.class, ChunkManager.class})
+@PrepareForTest( {Bukkit.class, ChunkManager.class})
```
Pr0methean: Delete the space before the curly bracket here.

Pr0methean (changes requested):

src/test/java/net/glowstone/entity/GlowPlayerTest.java
VaiTon: Done

src/test/java/net/glowstone/entity/GlowPlayerTest.java
VaiTon: Done.

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -310,12 +311,16 @@ public void onBlockChanged(GlowBlock block, Material oldType, byte oldData, Mate
      * @param block The block
      */
     public void updatePhysics(GlowBlock block) {
-        // do nothing
+        BlockPhysicsEvent event = EventFactory.getInstance()
+            .callEvent(new BlockPhysicsEvent(block, block.getTypeId()));
+        if (event.isCancelled()) {
+            return;
+        }
     }
```
Pr0methean, line 321: This won't make subclasses' updatePhysics cancelable. To do that, you have to create another empty method, call it at the end of updatePhysics, and make subclasses override that method instead of overriding updatePhysics.

Pr0methean (changes requested):

src/main/java/net/glowstone/block/blocktype/BlockType.java
VaiTon: Done

src/main/java/net/glowstone/block/blocktype/BlockType.java
```diff
@@ -293,29 +294,42 @@ public void onNearBlockChanged(GlowBlock block, BlockFace face, GlowBlock change
...
-        byte data) {
+                               byte data) {
         // do nothing
     }
 
     /**
      * Called when the BlockType should calculate the current physics.
+     * <br/>
+     * <br/>
+     * Subclasses <b>SHOULD NOT</b> override this method,
```
aramperes: Make the method final instead if possible. If it's not possible, remove those brs and use `<p>`, and indicate where overrides *are* needed.

aramperes (dismissed):

Pr0methean (approved):

PR #987, opened 2018-12-20
Localize BanCommand

Pr0methean:
This localizes BanCommand and also demonstrates how localized commands can work. Since each CommandSender has a different ResourceBundle, all LocalizedStringImpl instances a command uses must be created on demand, and thus LocalizedStringsTest's approach isn't applicable here.


src/main/java/net/glowstone/i18n/LocalizedString.java
```diff
@@ -8,4 +9,6 @@
     String get(Object... args);
 
     void send(CommandSender recipient, Object... args);
+
+    void sendInColor(CommandSender recipient, ChatColor color, Object... args);
```
mastercoms, line 13: I would prefer if we also supported multiple color codes within the message with the section sign (`§`).

But perhaps it's out of the scope of this PR.

mastercoms (approved):
Very well done!

src/main/java/net/glowstone/i18n/LocalizedString.java
Pr0methean: For multiple colors, you'd want to add the color-codes to the resource strings themselves anyway, to indicate which part was which color.

mastercoms: Yes, this is what I meant. Does that work already?

PR #988, opened 2018-12-23
Localize /ban, /ban-ip, /banlist, /clear

Pr0methean:
Localizes 4 more commands. Refactors sendInColor to put the color parameter first. Replaces several command-specific messages with a new generic "no such player" message to save L10n effort and allow object reuse; this can be done in future with other shared messages as well. Tests that description and usage messages exist in CommandTest if the command under test extends GlowVanillaCommand. (Once they all do, this condition will be removed.)


src/main/java/net/glowstone/command/minecraft/ClearCommand.java
```diff
@@ -215,8 +219,8 @@ private boolean clearAll(CommandSender sender, Player player, Material material,
                 }
             }
         }
-        sender.sendMessage(
-            "Cleared the inventory of " + player.getName() + ", removing " + count + " items");
+        new LocalizedStringImpl("clear.done", resourceBundle).send(
```
mastercoms: Could this be changed so the singular form `item` is used when `count == 1`?

mastercoms (dismissed):

src/main/java/net/glowstone/command/minecraft/ClearCommand.java
Pr0methean: Done (clear.done.singular).

mastercoms (approved):

PR #990, opened 2018-12-26
[i18n] Replace CommandUtils.prettyPrint with ListFormatter from ICU4J

Pr0methean:
Only applies to `/banlist` right now, but will apply to all commands that use `CommandUtils.prettyPrint` once the rest are converted to extend `GlowVanillaCommand`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,11 @@
       <artifactId>fast-uuid</artifactId>
       <version>0.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.ibm.icu</groupId>
+      <artifactId>icu4j</artifactId>
+      <version>63.1</version>
+    </dependency>
     <!-- NonNls annotation: https://mvnrepository.com/artifact/org.jetbrains/annotations -->
     <dependency>
       <groupId>org.jetbrains</groupId>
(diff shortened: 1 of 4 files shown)


smartboyathome:
It looks like ICU4J is distributed under a license I hadn't looked into much before, the Unicode License. I did a little digging, but couldn't find much information on it. Mind spending a bit of time to make sure it is compatible with our project?

Pr0methean:
@smartboyathome According to https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.aoplt00/aop2lt2024.htm, it's just an attribution license, except for the following unusual clause:

"Except as contained in this notice, the name of a copyright holder shall not be used in advertising or otherwise to promote the sale, use or other dealings in these Data Files or Software without prior written authorization of the copyright holder."

Pr0methean:
According to https://www.gnu.org/licenses/license-list.en.html#Unicode it's indeed compatible with LGPL and GPL.

aramperes:
ICU4J is a 12MB library. Is it worth adding this dependency for a single feature (pretty list formatting), and would there be a simple way to implement it ourselves?

Pr0methean:
We could extract the list patterns from the CLDR at build time (runtime wouldn't help, since the ZIP file we'd need is 20MB), but that'd be complicated to implement. We'll likely eventually be using ICU4J for more than just list formatting anyway. Why not add Proguard to our build process to remove unused library code?

Pr0methean:
An alternative would be to drop the word "and" and use a comma-separated list, which I *think* works in all languages.

Issue #994, opened 2018-12-28
Major performance hit/crash due to excessive threads


aramperes:
A critical performance hit/JVM crash was introduced in commit f4de5f65632df2125191f427927f1697392247fb (pull request #986). After the server completes world initialization, it immediately hangs and crashes due to lack of memory.

I am using the following startup command: `java -Xms768M -XX:+UseG1GC -jar glowstone.jar`. Allocating more memory does not seem to resolve the issue.

A JVM crash log is produced, which looks like:

```
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 32756 bytes for ChunkPool::allocate
# Possible reasons:
#   The system is out of physical RAM or swap space
#   In 32 bit mode, the process size limit was hit

[ ... ]

#
#  Out of Memory Error (allocation.cpp:273), pid=14260, tid=0x00003d70
#
# JRE version: Java(TM) SE Runtime Environment (8.0_131-b11) (build 1.8.0_131-b11)
# Java VM: Java HotSpot(TM) Client VM (25.131-b11 mixed mode windows-x86 )
# Failed to write core dump. Minidumps are not enabled by default on client versions of Windows
#

[ ... ]

---------------  P R O C E S S  ---------------

Java Threads: ( => current thread )
  0x3fc5d400 JavaThread "Glowstone-scheduler-1823" [_thread_blocked, id=19100, stack(0x7ff70000,0x7ffc0000)]
  0x3fc5a800 JavaThread "Glowstone-scheduler-1822" [_thread_blocked, id=19096, stack(0x7fee0000,0x7ff30000)]
  0x3fc5a000 JavaThread "Glowstone-scheduler-1821" [_thread_blocked, id=19124, stack(0x7fe50000,0x7fea0000)]
  0x3fc57c00 JavaThread "Glowstone-scheduler-1820" [_thread_blocked, id=19084, stack(0x7fdc0000,0x7fe10000)]
  0x3fc5c000 JavaThread "Glowstone-scheduler-1819" [_thread_blocked, id=14192, stack(0x7fd30000,0x7fd80000)]
  0x3fc55c00 JavaThread "Glowstone-scheduler-1818" [_thread_blocked, id=19076, stack(0x7fca0000,0x7fcf0000)]

[ ... ]

[...]

PR #996, opened 2018-12-28
Switch to ForkJoinPool

Pr0methean:
Scales the number of threads before letting tasks queue up. Should improve performance when there's a BlockPhysicsEvent handler (#994).


mastercoms (changes requested):
The benefit of using a ForkJoinPool is unclear (see https://github.com/netty/netty/issues/4466). If a performance improvement can be demonstrated, I will reconsider this PR.

mastercoms:
Will close this for now.

PR #1000, opened 2018-12-31
i18nize /deop and /difficulty; extract out LocalizedEnumNames

Pr0methean:
Most of the functionality of GameModeUtils is now in LocalizedEnumNames, generified for classes other than GameMode, and now also used with Difficulty.


mastercoms (approved):
Looks good to me!

PR #1001, opened 2019-01-05
i18n: Commands E-O

Pr0methean:
Localizes commands E through O. Also properly logs when /op or /deop fails with an exception, with a localized message, replacing `ex.printStackTrace()`.

Once this is merged, 20 of all 49 commands will have been i18nized, up from 8.

Also deploys the utility method `CommandUtils.toNamespaced` to append the `minecraft:` prefix when no namespace prefix is present, but make no change when an explicit namespace is present. This probably creates better error messages if a command is unsupported by Glowstone.


mastercoms (dismissed):
Hey, we discussed the namespaces a bit in the Discord, and I think an easy fix for now would be to check if a colon exists in the string before adding the Minecraft prefix. Also, please use Namespacedkey.MINECRAFT.

mastercoms (approved):

PR #1007, opened 2019-01-22
Parallelism - Pulsing entities and players in parallel divide and conquer manner.

Hagvan:
Utilizes CPU resources better.
I tested this using artificial load inside GlowEntity.pulse() method - each pulse every entity generates 10000 random integers.
Here are my benchmarks (CPU: Ryzen 2600X):
A message is shown every 120 ticks inside each world thread.

**Single threaded**
single thread

**ForkJoinPool**
forkjoinpool

Possible concerns:
I know in the future AI for mobs will be added and right now I don't know for sure if it will affect it.
What I did is just speed up iteration inside the GlowWorld.pulse() method.


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -464,7 +467,7 @@ public ChunkLock newChunkLock(@NonNls String desc) {
      */
     public void pulse() {
         List<GlowEntity> allEntities = new ArrayList<>(entityManager.getAll());
-        List<GlowPlayer> players = new LinkedList<>();
+        List<GlowPlayer> players = new ArrayList<>();
```
mastercoms, line 470: Why was this changed? Perhaps `Collections.synchronizedList()` would be better?

Hagvan: I used ArrayList because of O(1) access, which is required by ForkPulse and ForkPulsePlayers.
There is no need to use Collections.synchronizedList() because each entity is pulsed by only one thread anyway. There can't be memory consistency errors, at least I think so right now.

PR #1010, opened 2019-01-28
Use cloudflare DNS for session server IP lookup

heisluft:
This PR fixes an issue where the client would be kicked with the message: internal error during login. The server complains about not being able to reach the mojang session server, failing with a timeout exception. This is because netty's ```DefaultDnsServerAddressStreamProvider``` which uses Googles DNS servers (8.8.8.8 and 8.8.4.4) for IP lookup. This PR changes the DNS servers tou cloudflare's (1.1.1.1) thus resolving the issue.

This has been tested many times (it was frustrating to not be able to play on my server for 2 months now) and is proven to solve my issue (Thanks for your fast support on Discord, that really helped :D)

### DIFF
--- a/src/main/java/net/glowstone/net/http/HttpClient.java
+++ b/src/main/java/net/glowstone/net/http/HttpClient.java
@@ -22,8 +22,9 @@ import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.handler.timeout.ReadTimeoutHandler;
-import io.netty.resolver.dns.DefaultDnsServerAddressStreamProvider;
 import io.netty.resolver.dns.DnsAddressResolverGroup;
+import io.netty.resolver.dns.SingletonDnsServerAddressStreamProvider;
+import io.netty.util.internal.SocketUtils;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.util.concurrent.TimeUnit;
@@ -36,7 +37,7 @@ public class HttpClient {
     private static DnsAddressResolverGroup resolverGroup = new DnsAddressResolverGroup(
         GlowServer.EPOLL ? EpollDatagramChannel.class : GlowServer.KQUEUE
             ? KQueueDatagramChannel.class : NioDatagramChannel.class,
-        DefaultDnsServerAddressStreamProvider.INSTANCE);
+        new SingletonDnsServerAddressStreamProvider(SocketUtils.socketAddress("1.1.1.1", 53)));
 
     /**
      * Opens a URL.


smartboyathome:
Nice find, but is it possible to use the system's configured DNS server? I don't believe we should be ignoring what is configured in the system. Alternatively, if it's not possible to use the system's configured DNS, we should change these to be configurable rather than hardcoding defaults to a single provider that, for one reason or another, someone may object to.

heisluft:
Will address that as soon as i got time again (hopefully tomorrow). It seems like ```DefaultDnsServerAddressStreamProvider``` first tries to resolve the systems dns servers and uses google as fallback:
excerpt from  ```DefaultDnsServerAddressStreamProvider```:
```java
 // Using jndi-dns to obtain the default name servers.
        //
        // See:
        // - http://docs.oracle.com/javase/8/docs/technotes/guides/jndi/jndi-dns.html
        // - http://mail.openjdk.java.net/pipermail/net-dev/2017-March/010695.html
        Hashtable<String, String> env = new Hashtable<String, String>();
        env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.dns.DnsContextFactory");
        env.put("java.naming.provider.url", "dns://");
        try {
            DirContext ctx = new InitialDirContext(env);
            String dnsUrls = (String) ctx.getEnvironment().get("java.naming.provider.url");
            // Only try if not empty as otherwise we will produce an exception
            if (dnsUrls != null && !dnsUrls.isEmpty()) {
                String[] servers = dnsUrls.split(" ");
```
But this raises the question why it wouldn't be able to resolve the session servers:
curling them gives me back correct results, so it can't be my DNS config (or something is REALLY broken on my end).

mastercoms:
> These defaults do not respect your host's machines defaults.

[The docs](https://netty.io/4.1/api/io/netty/resolver/dns/DefaultDnsServerAddressStreamProvider.html) suggest otherwise.

mastercoms (changes requested):
I would recommend using a `MultiDnsServerAddressStreamProvider` collection of the following:

1. `SequentialDnsServerAddressStreamProvider` initialized from `glowstone.yml` optional user overrides.
2. `UnixResolverDnsServerAddressStreamProvider` to attempt to use the system DNS servers on Linux.
3. `SequentialDnsServerAddressStreamProvider` of `1.1.1.1`, `1.0.0.1`, `8.8.8.8`, `8.8.4.4`.

heisluft:
Appearantly, I was able to reach the sessionserver both with using cloudflare and google dns servers... The error seems to be the java excerpt I posted above. I looked into the matter and there is literally no way of solving the issue on Window using pure Java. If we decide it to be worth the effort I  _could_ write a C lib and do some JNI magic... However I'm not familiar with Maven at all and don't know how I tell Maven to compile this mess.

mastercoms:
That can be done later, but for now, I think what I proposed above is fine.

mastercoms:
Superseded by https://github.com/GlowstoneMC/Glowstone/commit/d486808b2c2f78da5d00aef1f33b90dff4970255.

PR #1014, opened 2019-02-03
Implement BlockPistonRetractEvent

Red-Teapot:
Addresses #922 

Although I have a couple of questions.

BlockPistonRetractEvent.java, lines 23-24:
```java
     * Get an immutable list of the blocks which will be moved by the
     * extending.
```
Should there be 'retracting' instead of 'extending'?

Is the list of moved blocks as the event parameter needed at all? Isn't just one block enough? Pistons can't move more than one block when retracting anyways.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockPiston.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockPiston.java
@@ -17,6 +17,7 @@ import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.event.block.BlockPistonExtendEvent;
+import org.bukkit.event.block.BlockPistonRetractEvent;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.PistonBaseMaterial;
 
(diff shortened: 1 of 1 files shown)


mastercoms:
Hi, I believe the comment should be retracting!

Also, slime blocks allow pistons to pull multiple blocks: 

> When a slime block is pushed or pulled by a piston it will attempt to move, in the same direction, all adjacent blocks.

https://minecraft.gamepedia.com/Slime_Block#Pistons

Red-Teapot:
Forgot about slime blocks, thanks. Although it seems like slime blocks are not pushed properly yet. Maybe I'll try to implement slime pushing then but I'm not sure if I can do that.

mastercoms:
You don't have to, would be out of the scope of this PR currently. Just wanted to let you know why the list exists :wink: 

But it's your choice if you want to include that functionality, I'm happy to accept it as part of this PR.

Red-Teapot:
I can't fix the comment since it's in Glowkit.

I think I'll create a separate PR when/if I implement slime pushing so it would be named properly and won't hang here.

mastercoms (approved):

PR #1015, opened 2019-02-03
Implemented PigZapEvent

heisluft:
Implements PigZapEvent (#922).
It relies on a custom spawn method in GlowWorld. I am open to any suggestions on how to solve this, although I think that this Method will be needed for any system where an Entity has to be mutated before it is spawned (or the spawn is cancelled) via an Event other than Creature/Item/EntitySpawnEvent.

### DIFF
--- a/src/main/java/net/glowstone/GlowWorld.java
+++ b/src/main/java/net/glowstone/GlowWorld.java
@@ -1411,6 +1411,38 @@ public class GlowWorld implements World {
         return null; // TODO: work on type mismatches
     }
 
+    //Needed for Events which mutate Entities before they are spawned
+    public <T extends GlowEntity> T spawn(Location location, T entity, SpawnReason reason)
+            throws IllegalArgumentException {
+        checkNotNull(entity).setRawLocation(location, false);
+        checkNotNull(reason);
+        checkNotNull(location);
+        try {
+            EntitySpawnEvent spawnEvent = null;
+            if (entity instanceof LivingEntity) {
+                spawnEvent = EventFactory.getInstance()
+                    .callEvent(new CreatureSpawnEvent((LivingEntity) entity, reason));
+            } else if (!(entity instanceof Item)) { // ItemSpawnEvent is called elsewhere
+                spawnEvent = EventFactory.getInstance().callEvent(new EntitySpawnEvent(entity));
+            }
+            if (spawnEvent != null && spawnEvent.isCancelled()) {
+                // TODO: separate spawning and construction for better event cancellation
+                entity.remove();
+            } else {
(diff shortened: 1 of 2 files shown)


src/main/java/net/glowstone/entity/GlowLightningStrike.java
```diff
@@ -113,6 +119,17 @@ public void pulse() {
                         ((Damageable) entity)
                             .damage(5, this, EntityDamageEvent.DamageCause.LIGHTNING);
                     }
+                    if (entity instanceof Pig) {
+                        GlowPig pig = (GlowPig) entity;
+                        GlowPigZombie zombie = EntityStorage.create(GlowPigZombie.class, location);
```
Pr0methean, line 124: Should other attributes like custom name be copied? Should an equipped saddle be dropped?

src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -1411,6 +1411,38 @@ public double getHumidity(int x, int z) {
...
+                // TODO: separate spawning and construction for better event cancellation
+                entity.remove();
+            } else {
+                List<Message> spawnMessage = entity.createSpawnMessage();
+                final GlowEntity finalEntity = entity;
+                getRawPlayers().stream().filter(player -> player.canSeeEntity(finalEntity))
+                    .forEach(player -> player.getSession().sendAll(spawnMessage
+                        .toArray(new Message[spawnMessage.size()])));
+            }
+        } catch (NoSuchMethodError | IllegalAccessError e) {
```
Pr0methean: Where are these thrown?

mastercoms:
I'm going to hold off on this PR until after we make the spawning code less miserable.

src/main/java/net/glowstone/entity/GlowLightningStrike.java
heisluft: Tested this in a SinglePlayer vanilla save w/ /entitydata commanf, only the customName is preserved. Potion effects, Inventory, Motion and Attributes are overwritten.

src/main/java/net/glowstone/GlowWorld.java
heisluft: CopyPasta fail 🤦‍♂️

heisluft:
> I'm going to hold off on this PR until after we make the spawning code less miserable.

Have you yet started working on / thinking about the rewrite?
Just asking whether you want to make this a personal issue or whether I could think about it and propose a PR as soon as I'm done.

mastercoms:
No plans at this moment, so feel free!

PR #1017, opened 2019-02-04
Implement slime blocks pushing and proper handling of non-movable blocks by pistons

Red-Teapot:
Added PistonMoveBehavior enum and corresponding methods in MaterialValueManager to store information about blocks push/pull behavior (e.g. bedrock does not push and pull, flowers drop when pushed and can't be pulled). All blocks can be pushed and pulled by default.

Pistons (including sticky variant) handle the behavior.

I haven't filled materialValues.yml yet so only bedrock and cobweb work for now.


src/main/java/net/glowstone/block/PistonMoveBehavior.java
```diff
@@ -0,0 +1,12 @@
+package net.glowstone.block;
+
+/**
+ * Behavior of a block when moved (pushed/pulled) py a piston.
+ *
+ * <p>MOVE - move as usual.</p>
+ * <p>BREAK - break block and drop necessary items (e.g. flowers).</p>
+ * <p>DONT_MOVE - don't move the block (e.g. bedrock).</p>
+ */
```
Pr0methean, line 10: Doesn't the Slime Block need a fourth mode?

Red-Teapot: What mode are you talking about? Special mode to determine if the block is slime? I don't know then, there is only one block with such properties for now.

And the Wiki says this:

> The types of blocks that can be moved are the same as those that can be pulled by a sticky piston. Blocks that cannot be pulled by a sticky piston (i.e. all the blocks listed on the table on the pistons page) stay in place.' 

So, I think that pistonPullBehavior property is enough for that.

Pr0methean: I mean to implement the behavior where a slime block moves other blocks when moved by a piston. 
https://gamepedia.cursecdn.com/minecraft_gamepedia/9/94/SlimeBlockFlyingMachineAnimation300lrgr.gif?version=b8b2274f33e41ad881e1878965f314de

Red-Teapot: Isn't it enough to just check if the material is slime block? So another variant in the enum is not needed.

I understand about the slime logic itself and I'm gonna work on it.

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -1,26 +1,41 @@
...
+import org.bukkit.*;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.event.block.BlockPistonExtendEvent;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.PistonBaseMaterial;
 
 public class BlockPiston extends BlockDirectional {
+    @AllArgsConstructor
+    private static class BlockInfo {
```
aramperes: I think you can use the `@Data` annotation, which will be equivalent to AllArgsConstructor+G/S.

src/main/java/net/glowstone/block/PistonMoveBehavior.java
smartboyathome: I would encourage you to rethink hard coding the slime block's "sticky" behavior. Even though slime blocks are the only instance of this behavior today doesn't mean it'll always be that way. Shouldn't be too much more work to abstract this behavior into some form of flag.

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -186,6 +216,92 @@ public void onRedstoneUpdate(GlowBlock me) {
         setType(me.getRelative(pistonBlockFace), 0, 0);
     }
 
+    // FIXME Remove this
+    private static Logger logger = Logger.getLogger(BlockPiston.class.getName());
```
Pr0methean: Unused.

src/main/java/net/glowstone/block/PistonMoveBehavior.java
garantiertnicht: Isn't there glazed terracotta as a special case that can only be pushed but not pulled?

mastercoms: Isn't that covered by separate pull and push behavior?

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -1,26 +1,41 @@
...
 import org.bukkit.event.block.BlockPistonExtendEvent;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.PistonBaseMaterial;
 
 public class BlockPiston extends BlockDirectional {
+    @AllArgsConstructor
+    private static class BlockInfo {
+        @Getter
+        @Setter
+        private GlowBlock block;
```
Pr0methean: Why not just use these two fields as parameters instead of needing this inner class? It'll be simpler as long as there are only two or three of them.

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -158,6 +216,92 @@ public void onRedstoneUpdate(GlowBlock me) {
...
+                case MOVE:
+                    blocksToMove.add(block);
+                    break;
+                case BREAK:
+                case DONT_MOVE:
+                    return true;
+            }
+        }
+
+        if (block.getType() == Material.SLIME_BLOCK) {
```
Pr0methean: Extract this into a subroutine. Add a PistonMoveBehavior.STICKY_MOVE and make it the value for Material.SLIME_BLOCK instead of checking for SLIME_BLOCK directly; that way, we can add the behavior to other blocks in the future.

Pr0methean (changes requested):

src/main/java/net/glowstone/block/PistonMoveBehavior.java
Red-Teapot: @garantiertnicht I've got separate move behaviors for pushing and pulling so it's not a problem. We can set Terracota's pushing behavior to MOVE and pulling behavior to DONT_MOVE.

Red-Teapot:
I think I've fixed all the stuff I needed to fix. Material push/pull behaviors are not filled yet though. Works only for cobweb and bedrock just as an example of non-pushable/drop-on-move blocks.

src/main/java/net/glowstone/block/blocktype/BlockPiston.java
```diff
@@ -156,23 +181,104 @@ public void onRedstoneUpdate(GlowBlock me) {
...
+            return addBlock(piston, movementDirection,
+                nextBlock, movementDirection.getOppositeFace(), blocksToMove, blocksToBreak);
+        } else {
+            return true;
         }
+    }
 
-        // remove piston head after retracting
-        setType(me.getRelative(pistonBlockFace), 0, 0);
+    private void breakBlock(GlowBlock block) {
```
aramperes, line 277: This logic doesn't take into account stuff like the `doTileDrops` gamerule and breaking containers (dropping chest contents). This kind of logic is already implemented in the [DiggingMessage handler](https://github.com/GlowstoneMC/Glowstone/blob/669ad8d45de9bea6ebe53a54bd6906fe9b4c5c87/src/main/java/net/glowstone/net/handler/play/player/DiggingHandler.java#L172-L185). Could you somehow merge this code?

If not, please leave a TODO and open an issue to support this.

Pr0methean (commented):
What are the correct piston behaviors for water and lava? Also, do we need special handling for blocks that can fall after being pushed (sand, gravel, concrete powder, dragon egg)?

mastercoms:
Merging this because it's a great start in our incomplete server implementation. Will iterate further based on reviews in this PR.

Red-Teapot:
@Pr0methean Movement behaviors for liquids are specified in material values. The only difference is that `getMinedDrops()` for liquids returns item stack containing liquid block (the method is not overridden). I'm not sure if I should override that method to return an empty stack so I made a [workaround](https://github.com/Red-Teapot/Glowstone/blob/9019fd6d91080b13d25d572b2efada22043810f2/src/main/java/net/glowstone/block/blocktype/BlockPiston.java#L298). It seems like liquids are destroyed on push and do not move on pull in vanilla game.

Regarding gravity-affected stuff, it's hard to say because it is hardly working already, e.g. blocks don't fall when a block below them is destroyed. Maybe I'll try to find out how to make blocks fall when I get some free time.

PR #1019, opened 2019-02-05
Regional difficulty

### LINKED ISSUE #545: Regional difficulty needs to be implemented
Currently, Glowstone has no concept of regional difficulty. RD and Clamped RD affects how some entities should behave.
There does not seem to be any api which needs to be implemented for this.

See:
https://minecraft.gamepedia.com/Difficulty#Regional_difficulty
https://minecraft.gamepedia.com/Difficulty#Other_effects

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48173712-regional-difficulty-needs-to-be-implemented?utm_campaign=plugin&utm_content=tracker%2F14691067&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14691067&utm_medium=issues&utm_source=github).
</bountysource-plugin>

VaiTon:
Implements regional difficulty for chunks following [this](https://minecraft.gamepedia.com/Difficulty#Clamped_regional_difficulty) algorithm.
Resolves issue #545

### DIFF
--- a/src/main/java/net/glowstone/chunk/GlowChunk.java
+++ b/src/main/java/net/glowstone/chunk/GlowChunk.java
@@ -659,29 +659,29 @@ public class GlowChunk implements Chunk {
         return heightMap[z * WIDTH + x] & 0xff;
     }
 
-    public float getRegionalDifficulty() {
-        final float moonPhase = world.getMoonPhase();
+    public double getRegionalDifficulty() {
+        final double moonPhase = world.getMoonPhase();
         final long worldTime = world.getFullTime();
         final Difficulty worldDifficulty = world.getDifficulty();
 
-        float totalTimeFactor;
+        double totalTimeFactor;
         if (worldTime > (21 * TickUtil.TICKS_PER_HOUR)) {
-            totalTimeFactor = 0.25f;
+            totalTimeFactor = 0.25;
         } else if (worldTime < TickUtil.TICKS_PER_HOUR) {
             totalTimeFactor = 0;
         } else {
-            totalTimeFactor = (float) (worldTime - TickUtil.TICKS_PER_HOUR) / 5760000;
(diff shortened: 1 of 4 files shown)


src/main/java/net/glowstone/chunk/GlowChunk.java
```diff
@@ -652,6 +659,65 @@ public int getHeight(int x, int z) {
         return heightMap[z * WIDTH + x] & 0xff;
     }
 
+    public float getRegionalDifficulty() {
+        final float moonPhase = world.getMoonPhase();
```
mastercoms: Why are these floats and not doubles?

mastercoms (changes requested):

src/main/java/net/glowstone/chunk/GlowChunk.java
VaiTon: At first I thought that using float was the right choice with java math, but now i did some researches and I understand that I should use double for this kind of things. Should I update [this method](https://github.com/GlowstoneMC/Glowstone/pull/1019/files/56fccd7a621490c98c9e881b3917a605093a73cc#diff-cea1d25e5518f2495b11e2d20021b0a0R751) too?

mastercoms: Yes please :)

aramperes (approved):
Looks good. In the future, avoid reformatting lines that aren't relevant to the PR, it makes the review process harder.

VaiTon:
Yeah, I'm sorry for that. It's just my IDE doing all when I commit. I think I will disable that :-)

mastercoms:
@momothereal Add `&w=1` to the files changed URL (also in diff settings, hide whitespace changes).

mastercoms (approved):

PR #1024, opened 2019-02-13
Portal Mechanics (Closes #130)

### LINKED ISSUE #130: Cannot create nether portals
Right now fireNetherPortal() is called when you use flint and steel on obsidian, but that function is empty. I'm gonna start filling that function to fire the nether portal. It would be a bit more complex, at least I think, than most things in java I've done before, but I think I can do it.
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29213197-cannot-create-nether-portals?utm_campaign=plugin&utm_content=tracker%2F14691067&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14691067&utm_medium=issues&utm_source=github).
</bountysource-plugin>

heisluft:
This PR implements the creation of nether portals and implements the actual travelling to the nether via portal as well as generating portals in the nether. It also implements any Bukkit / Spigot / PaperSpigot Event related to Portals and their mechanics.
This will fix one of Glowstone´s oldest issues while greatly expanding it´s functionality as a vanilla server.

What is implemented:
- Creation of nether portals by players
- Destroying Nether Portals
- Zombie Pigman Portal spawns
- Stubs of a TravelAgent (Thank you, @VaiTon )
- Generating and linking portals
- Travelling to the nether
- Bukkit Portal related Event implementations
- BUGFIX: Players somehow cannot see portal blocks


heisluft:
This is where I'm at now
[State of development](https://heisluft.tk/gh/GlowstoneMC/Glowstone/pr_1024_01.mp4)

VaiTon:
> This is where I'm at now
> [State of development](https://heisluft.tk/gh/GlowstoneMC/Glowstone/pr_1024_01.mp4)

Good job man!

heisluft:
[Next sneak peak](https://heisluft.tk/gh/GlowstoneMC/Glowstone/pr_1024_02.mp4) (chance of zombie pigman spawning increased by 1000% to shorten video length :))

heisluft:
Build will probably fail as I'm too tired to fix checkstyle warnings. will do that tomorrow.

VaiTon (commented):
I think it would be better if we implemented two TravelAgents, one for nether worlds and one for end ones.

heisluft:
My take on this is that each GlowWorld registers one TravelAgentImpl which then deals with portal implementation.
This also would allow us to handle Portal gen in the overworld - as Overworlds need to register both a Nether and an EndTravelAgent. getTravelAgent() in Entity/PlayerPortalEvent is not strictly required to return a consistent travelAgent but two implementations would just increase complexity (as the actual porting method is the same in both nether and end portals). Of course we could define a shared superclass but managing 3 classes + different handlers in the World class looks cumbersome to me compared to one, capable-of-everything class which can be referred to by each world in the same manner. Let me know if I missed something important in my logic

mastercoms:
@heisluft Is this still WIP or ready for review?

heisluft:
I’m sorry, I kind of lost track of this PR. After loosing some of my work to faulty git usage of mine I lost motivation to actually finish this PR. There are still features to implement  concerning the actual travel to the nether and I fear this PR is getting too big if I tried to mimic exact vanilla behavior. 

Don’t get me wrong though, I will continue my work on this today in order to get it working, although it will be more of a quick and dirty hack as I would have to make changes to GlowKit and seemingly unrelated code making this PR too confusing and difficult to review.

I look, however, into making a follow-up PR to GlowKit and GlowStone fixing the shortcomings of my hacky implementation.

TL;DR: No, this PR is not yet ready for review, but I continue contributing and will request review as soon as I’m done

heisluft:
[Next (And probably last sneak peak)](https://heisluft.tk/gh/GlowstoneMC/Glowstone/pr_1024_03.mp4).
I decided not to check wheter the bounding box of the portal block intersects the one of the player for now, so players get Teleported even if they technically are not standing in a portal. This has no other reason than not having enough time today.

EDIT: Done

heisluft:
Finished Work on this PR, ready for review

src/main/java/net/glowstone/block/blocktype/BlockPortal.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        PortalShape shape = new PortalShape(block.getLocation(), left);
+        if (shape.validate()
+                && shape.getPortalBlockCount() == shape.getHeight() * shape.getWidth()) {
+            return;
+        }
+        block.setType(Material.AIR);
+    }
+
+    public static BoundingBox getBoundingBox(GlowBlock block) {
```
aramperes: This needs documentation, specifying it only applies to portal blocks.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -605,6 +615,40 @@ public void pulse() {
...
+                    ? server.getWorld("world") : server.getWorld("world_nether");
+                TravelAgent agent = w.getTravelAgent();
+                boolean destIsNether = w.getEnvironment().equals(Environment.NETHER);
+                int destX = destIsNether ? location.getBlockX() / 8 : location.getBlockX() * 8;
+                int destY = destIsNether ? location.getBlockY() / 8 : location.getBlockY() * 8;
+                int destZ = destIsNether ? location.getBlockZ() / 8 : location.getBlockZ() * 8;
+                Location requested = new Location(w, destX, destY, destZ);
+                if (agent.getCanCreatePortal() || agent.findPortal(requested) != null) {
+                    Location teleportLocation = agent.findOrCreate(requested);
+                    EntityPortalEvent p = EventFactory.getInstance().callEvent(
```
aramperes: Avoid 1-letter variable. Rename this to `event` and the GlowWorld instance to `world`.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -605,6 +615,40 @@ public void pulse() {
                 }
             }
         }
+        GlowBlock block = world.getBlockAt(location);
+        if (block.getType().equals(Material.PORTAL)
+            && BlockPortal.getBoundingBox(block).intersects(boundingBox)) {
+            if (!isInPortal) {
```
aramperes, line 635: Could you document this code block overall? It is a bit hard to follow.

src/main/java/net/glowstone/util/GlowTravelAgent.java
```diff
@@ -0,0 +1,324 @@
...
+    private boolean canCreatePortal = true;
+    private World world;
+
+    /**
+     * Construct a new TravelAgent.
+     * @param world the world this agent operates in
+     */
+    public GlowTravelAgent(World world) {
+        if (world.getEnvironment() == World.Environment.THE_END) {
+            throw new UnsupportedOperationException("Not supported yet.");
```
aramperes: Change the message to something more descriptive, like `End portal teleportation is not supported yet.`

src/main/java/net/glowstone/util/pattern/PortalShape.java
```diff
@@ -0,0 +1,248 @@
...
+            if (offsetImmutable(bottomLeft, left, -j).add(0, height, 0).getBlock().getType()
+                != Material.OBSIDIAN) {
+                height = 0;
+                break;
+            }
+        }
+
+        if (height <= 21 && height >= 3) {
+            return height;
+        } else {
```
aramperes: The else statement is unnecessary because you are returning in the if-statement.

src/main/java/net/glowstone/util/pattern/PortalShape.java
```diff
@@ -0,0 +1,248 @@
...
+
+        for (int j = 0; j < width; ++j) {
+            if (offsetImmutable(bottomLeft, left, -j).add(0, height, 0).getBlock().getType()
+                != Material.OBSIDIAN) {
+                height = 0;
+                break;
+            }
+        }
+
+        if (height <= 21 && height >= 3) {
```
aramperes: I see `3`, `21`, and `22` come up a lot in your calculations. Can you extract those in constants? It seems like the minimum and maximum dimensions for portals.

src/main/java/net/glowstone/util/GlowTravelAgent.java
```diff
@@ -0,0 +1,324 @@
...
+        if (!EventFactory.getInstance().callEvent(event).isCancelled()) {
+            event.getBlocks().forEach(block ->
+                block.setTypeIdAndData(Material.PORTAL.getId(), meta, false));
+            return true;
+        }
+        return false;
+    }
+
+    private boolean canBuildPortal(Location loc, BlockFace facing) {
+        // TODO: Need impl
```
aramperes: Can you document what this method is supposed to do?

src/main/java/net/glowstone/util/GlowTravelAgent.java
```diff
@@ -0,0 +1,324 @@
...
+        int random = ThreadLocalRandom.current().nextInt(4);
+        int worldHeight = world.getEnvironment() == World.Environment.NETHER ? 128 : 256;
+
+        //try to find a suitable area for building the portal in range between
+        // x - 16, z - 16 and x + 16, z + 16.
+
+        //first attempt: find a portal spot that not only guarantees space for the portal, but also
+        //space around it
+        // iteration from x - 16 to x + 16
+        for (int tempX = reqX - 16; tempX <= reqX + 16; ++tempX) {
```
aramperes, line 121: This method is huge. Do you think you can break it down to multiple methods that can be tested individually?

src/main/java/net/glowstone/util/pattern/PortalShape.java
```diff
@@ -0,0 +1,248 @@
...
+        PortalCreateEvent event = new PortalCreateEvent(portalBlocks, bottomLeft.getWorld(),
+            PortalCreateEvent.CreateReason.FIRE);
+        if (!EventFactory.getInstance().callEvent(event).isCancelled()) {
+            // Dirty hack: directly calculate the metadata; Bukkit does not have portal material data :/
+            byte meta = (byte) (left == BlockFace.WEST ? 1 : 2);
+            event.getBlocks().forEach(block -> {
+                block.setType(Material.PORTAL);
+                block.setData(meta);
+            });
+
```
aramperes: Remove whitespace.

src/main/java/net/glowstone/util/pattern/PortalShape.java
```diff
@@ -0,0 +1,248 @@
...
+     *
+     * @param location the location from where to check. Not modified in process.
+     * @param face     the side of the edge to check for
+     * @return the distance or -1 in case of an invalid shape
+     */
+    private int getDistanceUntilEdge(Location location, BlockFace face) {
+        // Clone the location as Location objects are mutable
+        Location destLoc = location.clone();
+
+        int i;
```
aramperes: Rename `i` to `distance` in this method.

src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
```diff
@@ -13,6 +15,7 @@
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.util.Vector;
 
+
```
aramperes: Remove whitespace.

aramperes (changes requested):

aramperes:
This is some great work! I was hoping for more unit tests though, or at least setting up the ground work by separating the calculations into readable units.

src/main/java/net/glowstone/util/GlowTravelAgent.java
heisluft: Removed in next commit

src/main/java/net/glowstone/util/pattern/PortalShape.java
heisluft: Applied

src/main/java/net/glowstone/util/pattern/PortalShape.java
heisluft: Applied

src/main/java/net/glowstone/block/itemtype/ItemFlintAndSteel.java
heisluft: Applied

src/main/java/net/glowstone/util/pattern/PortalShape.java
heisluft: Applied

src/main/java/net/glowstone/entity/GlowEntity.java
heisluft: Applied

src/main/java/net/glowstone/entity/GlowEntity.java
heisluft: Applied

src/main/java/net/glowstone/block/blocktype/BlockPortal.java
heisluft: Applied

src/main/java/net/glowstone/util/GlowTravelAgent.java
heisluft: Applied

src/main/java/net/glowstone/util/GlowTravelAgent.java
heisluft: @momothereal I don't really know. If you want me to I will wrap my head around this. There are only two blocks (the first two) that are similar enough to have me try to extract them to a method

Pr0methean: This probably needs to be rewritten with functional syntax; for example, the nested search loops for X and Z (which repeat each other a bit) could become a subroutine taking an IntBinaryOperator scoring function.

heisluft: the problem with this is that the whole thing is highly state dependent the way it is currently written and i will have to change that, which is hat i am currently working on

Pr0methean: It occurs to me that it might actually be better to combine the nested for-loops into one (keep track of whether the best-so-far passed first attempt, or second attempt, or so on), which could be an easier place to optimize from and maybe even a candidate for ForkJoinPool parallelization (with the best-so-far variable possibly becoming an AtomicReference, in order for short-circuiting to still be possible).

heisluft: Never got into ForkJoinPool stuff. If it isn't that complicated I will learn it. If it isn't it might be an idea to merge this pr and then optimize it the way you want it.

src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -616,37 +616,47 @@ public void pulse() {
...
                 EventFactory.getInstance().callEvent(
                     new EntityPortalEnterEvent(this, block.getLocation()));
+                // prevent EnterEvent being fired again
                 isInPortal = true;
             }
+            // only try porting if the nether is enabled
             if (server.getAllowNether() && portalCooldown <= 0) {
-                GlowWorld w = getWorld().getEnvironment().equals(Environment.NETHER)
+                // determine destination world
+                GlowWorld destination = getWorld().getEnvironment().equals(Environment.NETHER)
```
aramperes, line 645: It might be necessary to ensure that there is such a world, even if the `allowNether` configuration is enabled.

src/main/java/net/glowstone/util/pattern/PortalShape.java
```diff
@@ -201,14 +216,15 @@ private int calculatePortalHeight() {
...
-            width = 0;
-            height = 0;
-            return 0;
         }
+
+        bottomLeft = null;
+        width = 0;
+        height = 0;
+        return 0;
+
```
aramperes: Remove whitespace.

momothereal: Remove whitespace.

src/main/java/net/glowstone/util/pattern/PortalShape.java
```diff
@@ -15,6 +15,21 @@
  */
 public class PortalShape {
 
+    /**
+     * Max portal width or height.
+     */
+    private static final int MAX_PORTAL_WIDTH_HEIGHT = 21;
```
aramperes, line 21: Could you document whether these constants are inclusive or exclusive?

aramperes (changes requested):
Final stretch!

src/main/java/net/glowstone/util/GlowTravelAgent.java
aramperes: I don't think the ForkJoinPool stuff is in the scope of this PR. But improving the readability of the logic is a big plus.

src/main/java/net/glowstone/entity/GlowEntity.java
Pr0methean: This logic would allow portals from the End to the Nether without cheats, which aren't possible in vanilla.

VaiTon: Should we just add a check for the type of world or something else?

smartboyathome: Perhaps look into adding a method to GlowWorld along the lines of:

```
boolean isPortalDestinationValid(BlockType type, GlowWorld destination)
```

This way, we aren't hardcoding Overworld to Nether, or Overworld to End, allowing for more extensibility later on (including for plugins).

src/main/java/net/glowstone/util/GlowTravelAgent.java
```diff
@@ -0,0 +1,319 @@
...
+    public Location findPortal(Location destination) {
+        Location minLoc = null;
+        double minDistance = Double.MAX_VALUE;
+
+        final int blockX = destination.getBlockX();
+        final int blockZ = destination.getBlockZ();
+
+        for (int x = blockX - searchRadius; x < (blockX + searchRadius); x++) {
+            for (int z = blockZ - searchRadius; z < (blockZ + searchRadius); z++) {
+                for (int y = 127; y >= 0; y--) {
```
Phoenix616: I think the start y value should be depending on the world type like it is in Vanilla Minecraft (127 for the nether, 255 for the rest)

heisluft: Thanks, totally missed that one

src/main/java/net/glowstone/entity/GlowEntity.java
heisluft: Applied the suggestions. Thanks for the input

mastercoms:
@aramperes What's left for this?

src/main/java/net/glowstone/util/GlowTravelAgent.java
aramperes: @mastercoms I think this was the last remaining comment to resolve

mastercoms: Portal creation logic is inherently complex, I will let this slide for now.

PR #1025, opened 2019-02-26
Animal feeding improvements

Pr0methean:
- Implements use of food for taming, horse healing, growth, or to set Love Mode.
- Animals won't consume food when it has none of these effects.
- `GlowAbstractHorse.canBreed()` can no longer return true if untamed.
- Implements the duration limit and the particle effect for Love Mode.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowAnimal.java
+++ b/src/main/java/net/glowstone/entity/GlowAnimal.java
@@ -59,6 +59,7 @@ public class GlowAnimal extends GlowAgeable implements Animals {
             boolean successfullyUsed = false;
             if (getBreedingFoods().contains(type)) {
                 if (canBreed()) {
+                    setInLove(1000); // TODO get the correct duration
                     // TODO set love mode if possible and spawn particles
                     // and don't set successfullyUsed if love mode is not possible
                     player.incrementStatistic(Statistic.ANIMALS_BRED);
--- a/src/main/java/net/glowstone/entity/GlowAnimal.java
+++ b/src/main/java/net/glowstone/entity/GlowAnimal.java
@@ -45,25 +45,26 @@ public class GlowAnimal extends GlowAgeable implements Animals {
     }
 
     /**
-     * Determines whether this entity can eat an item, if so, applies the effects of eating it.
+     * Determines whether this entity can eat an item while healthy, and if so, applies the effects
+     * of eating it.
      *
      * @param player the player feeding the entity, for statistical purposes
      * @param type an item that may be food
      * @return true if the item should be consumed; false otherwise
      */
     protected boolean tryFeed(Material type, GlowPlayer player) {
-        if (getBreedingFoods().contains(type)) {
-            if (canBreed() && (getInLove() <= 0)) {
-                setInLove(1000); // TODO get the correct duration
-                player.incrementStatistic(Statistic.ANIMALS_BRED);
-                return true;
-            } else {
-                int growth = computeGrowthAmount(type);
-                if (growth > 0) {
-                    grow(growth);
-                    return true;
-                }
-            }
+        if (!getBreedingFoods().contains(type)) {
+            return false;
+        }
+        if (canBreed() && getInLove() <= 0) {
+            setInLove(1000); // TODO get the correct duration
+            player.incrementStatistic(Statistic.ANIMALS_BRED);
+            return true;
+        }
+        int growth = computeGrowthAmount(type);
+        if (growth > 0) {
+            grow(growth);
+            return true;
         }
         return false;
     }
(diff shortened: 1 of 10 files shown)


aramperes:
What kind of live testing is required for this? Could you describe some testing steps?

Pr0methean:
Test that:

- Animals descending from GlowAbstractHorse eventually become tamed when fed repeatedly
- Baby animals grow when fed
- Injured animals heal when fed
- Adult animals, except untamed GlowAbstractHorse, enter Love Mode when fed (currently doesn't affect behavior, but should create heart particles)
- More than one of those things can happen with the same piece of food
- Food isn't consumed if none of those things can happen
- Love Mode eventually expires

FlorentClarret:
I tried your test and actually noticed some stuff:
- Heart particles are indeed created when an animal is fed, but the particles appear on their feet instead of above them, which looks a bit strange. 
- Food is still consumed even if nothing else happens. I think it's because the canBreed method return always true if the animal is adult. I tested on cows and pigs.

The other things seem to work, I will also try to take a look soon to improve animal feeding

Pr0methean:
@FlorentClarret For your test where food was falsely consumed, would the cow and pig have been expected to enter Love Mode? If not, why not?

FlorentClarret:
I just gave wheat to a cow, which was not in love mode and the cow entered in love mode as expected, but after that the wheat is still consumed if I keep giving it wheat. I tried on vanilla an it's not the case, the wheat is not consumed if the cow is already in love mode.

Pr0methean:
Both issues should be fixed now.

FlorentClarret (approved):
Looks good to me now. Thanks @Pr0methean

src/main/java/net/glowstone/entity/GlowAnimal.java
```diff
@@ -44,6 +44,30 @@ protected int getAmbientDelay() {
...
 
+    /**
+     * Determines whether this entity can eat an item, if so, applies the effects of eating it.
+     *
+     * @param player the player feeding the entity, for statistical purposes
+     * @param type an item that may be food
+     * @return true if the item should be consumed; false otherwise
+     */
+    protected boolean tryFeed(Material type, GlowPlayer player) {
+        if (getBreedingFoods().contains(type)) {
```
aramperes: Remove a nesting level by inverting this if-statement.

src/main/java/net/glowstone/entity/GlowAnimal.java
```diff
@@ -44,6 +44,30 @@ protected int getAmbientDelay() {
...
+            if (canBreed() && (getInLove() <= 0)) {
```
aramperes: The parentheses for the second condition are unnecessary.

src/main/java/net/glowstone/entity/GlowAnimal.java
```diff
@@ -44,6 +44,30 @@ protected int getAmbientDelay() {
...
+                setInLove(1000); // TODO get the correct duration
+                player.incrementStatistic(Statistic.ANIMALS_BRED);
+                return true;
+            } else {
```
aramperes: The else-statement is unnecessary.

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
```diff
@@ -35,6 +36,19 @@
             .put(Material.HAY_BLOCK, TickUtil.minutesToTicks(3))
             .build();
 
+    @Override
+    protected boolean tryFeed(Material food, GlowPlayer player) {
+        if (isAdult() && !isTamed()) {
```
aramperes: Remove a nesting level by inverting this if-statement.

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
```diff
@@ -104,6 +123,11 @@ private int getHorseFlags() {
         return BREEDING_FOODS;
     }
 
+    protected int computeTamingAmount(Material material) {
+        // TODO
```
aramperes, line 137: Please document what's the intended functionality of the method.

aramperes (dismissed):

src/main/java/net/glowstone/entity/GlowAnimal.java
ghost: Good code is a balance of function & readability.

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
Pr0methean: Done, and renamed to `computeDomestication` to match the Bukkit nomenclature. Parameter renamed to `food`.

src/main/java/net/glowstone/entity/passive/GlowAbstractHorse.java
Pr0methean: Done.

src/main/java/net/glowstone/entity/GlowAnimal.java
Pr0methean: Removing it in this case doesn't hurt readability.

src/main/java/net/glowstone/entity/GlowAnimal.java
Pr0methean: Removed, although I would've thought they'd improve readability (since one would guess that `getInLove()` returned boolean without knowing otherwise).

src/main/java/net/glowstone/entity/GlowAnimal.java
Pr0methean: Done.

aramperes (approved):
Good work! Sorry for the delay.

PR #1028, opened 2019-03-05
Reduce some duplication in networking-related code

kashike:


src/main/java/net/glowstone/net/Networking.java
```diff
@@ -0,0 +1,68 @@
...
+import io.netty.channel.kqueue.KQueueSocketChannel;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.DatagramChannel;
+import io.netty.channel.socket.ServerSocketChannel;
+import io.netty.channel.socket.SocketChannel;
+import io.netty.channel.socket.nio.NioDatagramChannel;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.channel.socket.nio.NioSocketChannel;
+
+public final class Networking {
```
aramperes, line 22: Could you add some Javadoc for the class its the public members?

aramperes (approved):

PR #1031, opened 2019-03-06
Set up CI with Azure Pipelines

gdude2002:
This is just the initial file, we'll need to modify this.


Pr0methean:
Make sure the environment doesn't require software to be "cgroup-aware", as many newer container platforms do. This means commands such as `free` and `top` must report only *our* share of the host's memory. We have something somewhere in our build pipeline that's not cgroup-aware and runs out of memory on the CircleCI container platform (so we're now using their traditional VM instead), and I found it to be beyond reach of JVM args; maybe that's happening here too.

gdude2002:
Well, so far so good! I think we'll be alright, given that this runs just fine without the memory limiting specified in the CircleCI env vars. Still more work to do to finish this thing off, though.

gdude2002:
This should now be functionally equivalent to the CircleCI pipeline.

mastercoms (approved):

Paulomart:
Why is the trigger set to master? Are feature branches no longer tested?

gdude2002:
Oh right, I forgot to update that from the template. I'll fix it in #1032

PR #1033, opened 2019-03-12
Test Java 8 too on Azure Pipelines

Pr0methean:
Since Azure Pipelines allows up to 10 concurrent jobs, we can easily afford to run a second one per push, and it can help us ensure we're not breaking compatibility with JDK 8 (which I think some plugins still require). This PR adds a second job on JDK 8, and renames the jars from that second job; I'm not sure why there's only one coverage report.


azure-pipelines.yml
```diff
@@ -50,21 +59,21 @@ jobs:
 
     inputs:
       artifactName: 'Glowstone'
-      targetPath: 'target/glowstone.jar'
+      targetPath: 'target/glowstone$(artifact_suffix).jar'
```
gdude2002: No, you don't change the path, you change the artifact name

azure-pipelines.yml
```diff
@@ -12,6 +12,15 @@ jobs:
 - job: build
   displayName: "Build"
 
+  strategy:
+    matrix:
+      Java8:
+        jdk_version: 1.8
+        artifact_suffix: '-Java8'
```
gdude2002: I would prefer the suffix to be `" (Java 8)"` since it appears in the name of the artifact, not the filename

gdude2002 (dismissed):

azure-pipelines.yml
Pr0methean: It's used in the filename as well. I tried changing it but got an error: https://dev.azure.com/glowstonemc/Glowstone/_build/results?buildId=60&view=logs&jobId=d59b01f9-f6ba-5c44-01e4-578439ff0ce1&taskId=99fccf4a-57f7-54e6-97b7-8a840e9f0e41&lineStart=9&lineEnd=10&colStart=1&colEnd=1

In follow-up PRs, we can try fancier formatting solutions like using different versions of the suffix in different places. But for now, it looks like perfect is the enemy of good.

gdude2002 (commented):
I think this looks good - RE code coverage, do we need to be running coverage on anything but the latest JDK version? Can we skip it on the JDK8 build? It doesn't seem like we really need to be running it twice.

What do you think?

Pr0methean:
Turned off JDK8 coverage.

etc/azure-templates/build.yml
```diff
@@ -0,0 +1,16 @@
...
+    mavenPomFile: "pom.xml"
+
+    javaHomeOption: "JDKVersion"
+    jdkVersionOption: "$(jdk_version)"
+    jdkArchitectureOption: "x64"
+    publishJUnitResults: true
+    testResultsFiles: '**/surefire-reports/TEST-*.xml'
+
+    goals: "package source:jar javadoc:jar"
+    options: "--batch-mode"
```
gdude2002: Missing newline at end of file

azure-pipelines.yml
```diff
@@ -45,23 +40,4 @@ jobs:
...
-      artifactName: 'JavaDoc JAR'
-      pathtoPublish: 'target/glowstone-javadoc.jar'
-
-  - task: PublishBuildArtifacts@1
-    displayName: "Artifact: Sources JAR"
-
-    inputs:
-      artifactName: 'Sources JAR'
-      pathtoPublish: 'target/glowstone-sources.jar'
+  - template: etc/azure-templates/publish-artifacts.yml
```
gdude2002: Missing newline at end of file.

azure-pipelines.yml
```diff
@@ -9,32 +9,27 @@ trigger:
     - "*"
 
 jobs:
-- job: build
-  displayName: "Build"
-
+- job: build8
```
gdude2002, line 12: This file is now very dense - please space it out a little more.

gdude2002 (dismissed):
Good work - resolve this review, and I'll be happy to merge this.

It's a shame we couldn't do this using the version matrix, but that's okay. This should work nicely.

azure-pipelines.yml
Pr0methean: Done: added a blank line between jobs, and between steps and configuration for each job.

azure-pipelines.yml
Pr0methean: Fixed.

etc/azure-templates/build.yml
Pr0methean: Fixed.

gdude2002 (approved):

PR #1034, opened 2019-03-14
Replace ad-hoc Validator class with Java Predicate

socram8888:
Hello. This PR is a really tiny improvement for the `TaxicabBlockIterator` class I made - basically, replacing the `Validator` ad-hoc class I made pre-Java-8 with `Predicate`, which is essentially the same except built into Java already.


VaiTon (approved):

mastercoms:

PR #1044, opened 2019-06-21
Make login DNS server configurable in glowstone.yml

smartboyathome:
This should allow the DNS servers used by our login system to be configurable. A lot of refactoring had to be done in order to allow injecting the configs into the various protocols. Unfortunately, this meant that they could no longer be singleton enums.

Currently, three's a bug that causes the handshake to become recursive.


smartboyathome:
Build failures are due to the build not picking up the changes I made in flow network

mastercoms (approved):

PR #1056, opened 2019-09-24
Implement blockdata command

clabe45:
Implements the `/blockdata` for 1.12.

Since `BlockEntity#loadNbt` isn't fully implemented, this command does not work. When `loadNbt` is done, this should work, but I can update it if not.


clabe45:
I now understand that `BlockEntity#loadNbt` is meant to be overriden, and that `ChestEntity#loadNbt` is not implemented. Since I was only testing with chests, it didn't work.

I will test with blocks that do implement it.

aramperes:
`ChestEntity#loadNbt` is inherited from `ContainerEntity#loadNbt`, which properly reads the "Items" tag. I'm not aware of any issue with `loadNbt`.

PR #1065, opened 2019-10-24
Implement EntityDamageByBlockEvent firing

ytnoos:
I created the new method `getTouchingBlocks()` to get every block a player was colliding with to put it on the EntityDamageByBlockEvent.

I created the `canTakeDamage()` method without any parameters to avoid code duplication on the new `damageBlock()` method.

I implemented[ `damageBlock()` method because default damage LivingEntity's method does not allow Blocks as a damage source.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowEntity.java
@@ -982,20 +982,40 @@ public abstract class GlowEntity implements Entity {
                 }
             }
         } else {
-            // bounding box-based calculation
-            Vector min = boundingBox.minCorner;
-            Vector max = boundingBox.maxCorner;
-            for (int x = min.getBlockX(); x <= max.getBlockX(); ++x) {
-                for (int y = min.getBlockY(); y <= max.getBlockY(); ++y) {
-                    for (int z = min.getBlockZ(); z <= max.getBlockZ(); ++z) {
-                        if (world.getBlockTypeIdAt(x, y, z) == material.getId()) {
-                            return true;
-                        }
-                    }
+            for (Block touchingBlock : getTouchingBlocks()) {
+                if (touchingBlock.getType().getId() == material.getId()) {
+                    return true;
                 }
             }
         }
         return false;
+    }
+
+    /**
+     * Gets a list containing all the blocks a player is touching
+     *
+     * <p>If the entity has not a defined bounding box, the list will be empty.
+     *
+     * @return the list of blocks a player is touching
+     */
+    public List<Block> getTouchingBlocks() {
+        if(boundingBox == null) {
+            return Collections.emptyList();
+        }
+
+        List<Block> blocks = new ArrayList<>();
+
+        // bounding box-based calculation
+        Vector min = boundingBox.minCorner;
+        Vector max = boundingBox.maxCorner;
+        for (int x = min.getBlockX(); x <= max.getBlockX(); ++x) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/entity/GlowEntity.java
```diff
@@ -1442,6 +1462,9 @@ public void damage(double amount, DamageCause cause) {
         damage(amount, null, cause);
     }
 
+    public void damageBlock(double amount, Block block, DamageCause cause) {
```
mastercoms: Could we rename this one to `damage`?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -936,10 +960,50 @@ public void setHealth(double health) {
...
+        // armor damage protection
+        // formula source: http://minecraft.gamepedia.com/Armor#Damage_Protection
+        double defensePoints = getAttributeManager().getPropertyValue(Key.KEY_ARMOR);
+        double toughness = getAttributeManager().getPropertyValue(Key.KEY_ARMOR_TOUGHNESS);
+        amount = amount * (1 - Math.min(20.0,
+                Math.max(defensePoints / 5.0,
+                        defensePoints - amount / (2.0 + toughness / 4.0))) / 25);
+
+        // fire event
+        EntityDamageByBlockEvent event = EventFactory.getInstance().onEntityDamage(
```
mastercoms, line 994: It's really non-ideal that we have to do this. Could we extract some of the damage code into two parts, so that both the entity and block pathway can call it appropriately?

mastercoms (changes requested):

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -936,10 +960,54 @@ public void setHealth(double health) {
...
+            noDamageTicks = maximumNoDamageTicks;
+        }
+
+        amount = getEffectiveDamage(amount);
+
+        // fire event
+        EntityDamageByBlockEvent event = EventFactory.getInstance().onEntityDamage(
+                new EntityDamageByBlockEvent(block, this, cause, amount));
+
+        if (event.isCancelled()) {
```
mastercoms: Still some duplicate code here. Couldn't we pass in the `EntityDamageEvent` to a new function?

mastercoms (changes requested):

src/main/java/net/glowstone/entity/GlowLivingEntity.java
ytnoos: Duplicated code is really short and it isn't worth to make a new function to allow both events, they are much different and their handling is chopped

mastercoms (changes requested):
Could you fix the checkstyle errors as reported by our CI? The rest LGTM.

mastercoms:
```
[ERROR] src/main/java/net/glowstone/entity/GlowLivingEntity.java:[77] (imports) CustomImportOrder: Wrong lexicographical order for 'org.bukkit.event.entity.EntityDamageByBlockEvent' import. Should be before 'org.bukkit.event.entity.EntityDamageByEntityEvent'.
[ERROR] src/main/java/net/glowstone/entity/GlowLivingEntity.java:[1019] (coding) OverloadMethodsDeclarationOrder: Overload methods should not be split. Previous overloaded method located at line '965'.
```

PR #1067, opened 2019-11-12
Add missing options to time command

DenyCZ:
Added missing time options to time command. Times are from the official minecraft wikipedia.


mastercoms (changes requested):
Could you add a TODO comment for query?

mastercoms (approved):

PR #1069, opened 2019-11-29
Migrate enum entries to upstream

KisaragiEffective:
Fix those entries since Bukkit has changed those name.


KisaragiEffective:
There must be misunderstood.

PR #1073, opened 2020-01-14
[WIP] Ran a Node script to remove the damage property from the recipes.yml.…

birbe:
Removed the damage property from the recipes with a Node script. This also makes it easier to deal with as in many places we no longer return a map that includes a material along with its damage but simply the material namespaced key alone.

Someone still needs to remove the non-existent blocks from the recipes.

### DIFF
--- a/src/main/java/net/glowstone/inventory/crafting/CraftingManager.java
+++ b/src/main/java/net/glowstone/inventory/crafting/CraftingManager.java
@@ -12,7 +12,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import net.glowstone.block.MaterialUtil;
-import net.glowstone.constants.ItemIds;
 import net.glowstone.i18n.ConsoleMessages;
 import net.glowstone.inventory.GlowCraftingInventory;
 import net.glowstone.io.nbt.NbtSerialization;
@@ -468,8 +467,11 @@ public final class CraftingManager implements Iterable<Recipe> {
 
         // shaped
         for (Map<?, ?> data : config.getMapList("shaped")) { // NON-NLS
-            ItemStack resultStack =
-                    ItemStack.deserialize((Map<String, Object>) data.get("result")); // NON-NLS
+            Map<String, Object> _recipe = (Map<String, Object>) data.get("result");
+            _recipe.put("v", 1); //Not legacy.
+
+            ItemStack resultStack = ItemStack.deserialize(_recipe);
+
             NamespacedKey key = readKey(data, resultStack);
             ShapedRecipe recipe = new ShapedRecipe(key, resultStack);
             List<String> shape = (List<String>) data.get("shape"); // NON-NLS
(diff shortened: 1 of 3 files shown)


birbe:
Building has failed as I haven't updated code to suit the new format.

mastercoms (changes requested):
This also gets rid of the amount specifier. I wouldn't change the format and instead just remove the damage IDs.

birbe:
> 
> 
> This also gets rid of the amount specifier. I wouldn't change the format and instead just remove the damage IDs.

My bad. Will fix.

birbe:
I've updated the result.

birbe:
The YML parser seemed to have messed up re-serializing some of the arrays and they are now just keyed maps, am fixing.

src/main/java/net/glowstone/inventory/crafting/CraftingManager.java
```diff
@@ -478,40 +480,55 @@ private void loadRecipes() {
...
             ItemStack resultStack
-                    = ItemStack.deserialize((Map<String, Object>) data.get("result")); // NON-NLS
+                    = ItemStack.deserialize((Map<String, Object>) data.get("result")); // NON-NLS;
+
             NamespacedKey key = readKey(data, resultStack);
-            float xp = ((Number) data.get("xp")).floatValue(); // NON-NLS
-            furnaceRecipes.add(
-                    new FurnaceRecipe(key, resultStack, inputStack.getType(),
-                            inputStack.getDurability(), xp, 200));
+            //float xp = ((Number) data.get("xp")).floatValue(); // NON-NLS
```
mastercoms, line 526: Seems like xp was removed from the original recipes.yml. Is that intentional? If not, could you add it back?

src/main/java/net/glowstone/inventory/crafting/CraftingManager.java
```diff
@@ -478,40 +480,55 @@ private void loadRecipes() {
...
 
         // furnace
         for (Map<?, ?> data : config.getMapList("furnace")) { // NON-NLS
+
+            ((Map<String, Object>) data.get("input")).put("v", 1); //Non-legacy.
+            ((Map<String, Object>) data.get("result")).put("v", 1); //Non-legacy.
+
             ItemStack inputStack
-                    = ItemStack.deserialize((Map<String, Object>) data.get("input")); // NON-NLS
+                    = ItemStack.deserialize((Map<String, Object>) data.get("input")); // NON-NLS;
```
mastercoms, line 521: Remove semicolon here.

mastercoms (changes requested):

mastercoms:
@Birbe could you possibly send us the Node script under a compatible license for us to continue this work?

birbe:
I probably won't be able to find it; I might be able to rewrite it though.

mastercoms:
Ok, if you could do that, it would be most helpful. Thank you!

birbe:
Oof, really sorry to leave you hanging like this. I got the script; I'm cleaning up my computer and I found it, not sure if it's still relevant. 
[rx2j.zip](https://github.com/GlowstoneMC/Glowstone/files/5732379/rx2j.zip)

As for the license:

I donate the attached files and code as a part of the GlowstoneMC project, you may use it however you'd like.

mastercoms:
Thanks so much!

PR #1075, opened 2020-01-17
Fix wrong coordinates in isSnowy

Muqsit:
The values of x and z variables over here are absolute world coordinates (the coordinates aren't relative to the chunk). Adding sourceX and sourceZ (the absolute coordinates of the chunk's beginning) sets the value far too high.

Possibly, the author was trying to do something like this: https://github.com/GlowstoneMC/Glowstone/blob/dev/src/main/java/net/glowstone/generator/objects/Lake.java#L95

But in Lake.java, x and z are coordinates relative to the chunk so it makes sense in Lake.


mastercoms (approved):

mastercoms:

PR #1076, opened 2020-01-21
BlockData implementation

mosemister:
This is my first PR for any public project so sorry if the format isn't correct. 

This PR is to solve the issue of https://github.com/GlowstoneMC/1.13-board/issues/28 As this is my first issue any and all comments will be appciated to make the code more to the Glowstone standard.

Implementations currently done:

- [x] NoteBlock
- [X] Anvil
- [X] Bed
- [X] Beetroot
- [X] EnderRod
- [x] BrewingStand
- [x] BubbleColumn
- [x] Cactus
- [x] Cake
- [x] Carrots
- [x] Cauldron
- [x] Chest
- [x] ChorusFlower
- [x] ChorusFruit
- [x] CobblestoneWall
- [x] Cocoa
- [x] Command
- [x] Comparator
- [x] Conduit
- [x] CoralDead
- [x] CoralFan
- [x] CoralFanWall
- [x] CoralPlant
- [x] DaylightDetector
- [x] Detector Rail
- [x] DirtSnow
- [x] Dispenser
- [x] Door
- [x] Dropper
- [x] EnderChest
- [x] EnderPortalFrame
- [x] Fence
- [x] FenceGate
- [x] Fire
- [x] FloorSign
- [x] Fluid
- [x] Frosted Ice
- [x] Furnace (original)
- [x] GlazedTerracotta
- [x] Grass
- [x] Hay
- [x] Hopper
- [x] HugeMushroom
- [x] IronBars
- [x] Jack-O-lantern
- [x] JukeBox
- [x] Kelp
- [x] Ladder
- [x] Leaves
- [x] Lever
- [x] Log
- [x] MinecartDetector
- [x] MinecartTrack
- [x] Mycel
- [x] NetherWart
- [x] Observer
- [x] Piston
- [x] PistonExtension
- [x] PistonMoving
- [x] PlayerSkull
- [x] Portal
- [x] Potatoes
- [x] PowerRails
- [x] PressurePlate Binary
- [x] PressurePlateWeighted
- [x] RedstoneLamp
- [x] RedstoneOre
- [x] RedstoneRepeater
- [x] RedstoneTorchStanding
- [x] RedstoneTorchWall
- [x] RedstoneWire
- [x] Reeds
- [x] Sapling
- [x] SeaPickle
- [x] ShulkerBlock
- [x] Skull
- [x] Snow
- [x] Soil
- [x] StainedGlassPane
- [x] Stairs
- [x] StandingBanner
- [x] Stem
- [x] StoneButton
- [x] Structure
- [x] TNT
- [x] TallPlantFlower
- [x] TallSeaGrass
- [x] TrapDoor
- [x] TrappedChest
- [x] Tripwire
- [x] TripwireHook
- [x] TurtleEgg
- [x] Vine
- [x] WallBanner
- [x] WallPlayerSkull
- [x] WallSign
- [x] WallTorch
- [x] WallWitherSkull
- [x] WitherSkull
- [x] WoodenButton


mastercoms:
Hi, welcome to Glowstone! I really appreciate your initiative on this issue. One thing I would like to point out is that statically typed properties like this will prove to be cumbersome when initializing block data from IDs, unless we were to create class specific initializing for every material, or used slow reflection to product block data. Neither approach is preferable, so I'd recommend taking a different approach to tackle this problem. Admittedly, this issue is one of our most difficult, as signified by the high bounty. If this is your first PR, I would recommend tackling one of the smaller issues first, and communicating with us on [our Discord ](https://discord.gg/TFJqhsC) to get help on how to get your way around Glowstone!

smartboyathome:
To add to what @mastercoms said, this solution does not provide a mechanism for translating blockdata to the appropriate over-the-network ID, which is stored in the generated block report (see the linked issue on 1.13-board). Your solution will need to take this into account for that.

mosemister:
Thats fair. Because bukkit uses interfaces, such as Ageable, Directional, etc. I didn't think there would be another way to implement BlockData. 

Ill give it a think and work on the network specific ids or something. Im not too fussed about the money, more motivated on seeing a 1.13 glowstone

mosemister:
Thats fair. I am very familiar with Sponge API that uses Key Value pairing for this and I know the glowstone project doesn't want anything based off another project. 

However I have just seen the classes in the flattening folder, in particular, the BlockData implementation there that uses Maps. Ill expand on that keeping into account the network api

src/main/java/net/glowstone/block/data/AbstractBlockData.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public @NotNull String getAsString(){
+        String entries = null;
+        for(StateValue<?> value : this.values){
+            if(entries == null){
+                entries = value.getGenerator().getId() + "=" + value.getValueAsString();
+            }else{
+                entries = entries + ", " + value.getGenerator().getId() + ": " + value.getValueAsString();
```
smartboyathome: Did you mean to give the top line a different id-value separator from the bottom line?

src/main/java/net/glowstone/block/data/AbstractBlockData.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public @NotNull BlockData merge(@NotNull BlockData blockData) {
+        ((AbstractBlockData)blockData).getStateValues().forEach(s -> applyStateValue(s));
+        return this;
+    }
+
+    @Override
+    public boolean matches(@Nullable BlockData blockData) {
+        return Objects.equals(blockData, this);
+    }
```
smartboyathome: According to the docs, the beavior for these two methods change depending on in the block data has been parsed using `Server.createBlockData(String)` or not. This seems like it'll require another flag be added to this class, to be set by the aforementioned method. In addition, for `matches(BlockData)`, the docs mention that only explicitly checked states should be matched, so it seems like your `StateValue` class will need a flag to determine whether it was initialized with an explicit state or using the default value.

src/main/java/net/glowstone/block/data/AbstractBlockData.java
mosemister: Nope. Haha. Thanks for noticing that

src/main/java/net/glowstone/block/data/AbstractBlockData.java
mosemister: Must have missed that. Thanks for the tip

src/main/java/net/glowstone/block/data/IBlockData.java
```diff
@@ -0,0 +1,21 @@
...
+
+import net.glowstone.block.data.state.StateGenerator;
+import net.glowstone.block.data.state.StateValue;
+import org.bukkit.block.data.BlockData;
+
+import java.util.Set;
+
+public interface IBlockData extends BlockData {
+
+    Set<StateValue<?>> getStateValues();
```
smartboyathome: Should this be a map instead of a set? Seems like, quite often, we're fetching things by ID, which requires a scan. Admittedly, this won't be a huge performance impact, since there won't be many pieces of state, but it still would be good practice from a design perspective.

src/main/java/net/glowstone/block/data/IBlockData.java
```diff
@@ -0,0 +1,21 @@
...
+
+    default <T extends Object> StateValue<T> getStateValue(StateGenerator<T> generator){
+        return (StateValue<T>) this.getStateValues().stream().filter(s -> s.getGenerator().equals(generator)).findAny().orElse(null);
+    }
+
+    default StateValue<?> getStateValue(String id){
+        return this.getStateValues().stream().filter(s -> s.getGenerator().getId().equals(id)).findAny().orElse(null);
+    }
```
smartboyathome: Nitpick: Personally, I prefer using Optionals instead of nulls when a value may or may not be present, since it's easy to tell at a glance that this is the intended behavior. That's just me, though.

mosemister: Personally I also prefer to use Optionals too, but because the Bukkit API didnt, I chose not to in the name of keeping coherence. But I am willing to change it to optional in the name of pushing for modern API

src/main/java/net/glowstone/block/data/IBlockData.java
mosemister: Never knew about that design perspective. Just wanted it to be easy to maintain so I didnt allow for the possibility of the two IDs being unsynced by a plugin. 

Ill change it

src/main/java/net/glowstone/block/data/state/StateGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+import org.bukkit.Instrument;
+import org.bukkit.block.BlockFace;
+import org.bukkit.block.data.Bisected;
+import org.bukkit.block.data.Rail;
+import org.bukkit.block.data.type.*;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public interface StateGenerator<T> {
```
smartboyathome: What is this interface being used for? Is it attempting to generate the index in the block report's property value arrays within the state files?

mosemister: The StateGenerator class is for the common class between all generators. So the Generators are used to generate all common values for the states, such as network codes, the ID of the StateValue. 

So for example, the NETHER_PORTAL block data has a state of AXIS that can be either X or Z. The StateValue will contain the current value of the AXIS, as in it will either hold the enum of Axis.X or Axis.Z, while the generator will hold all the possible values for the AXIS nether portal, as well hold the ID of AXIS.

minecraft:nether_portal{axis=x}
[material id]{[generator]=[value]}

smartboyathome: Our plan is currently to utilize the blocks report generated by the vanilla server JAR at runtime. Due to this, there will likely be a separate manager handling the downloading of the vanilla server and generation of the report (if not already present), and desalinizing of the JSON model. I've got a prototype manager written already that reads and deserializes this report. How would this model interface with such a mechanism for determining the network IDs?

mosemister: From what I can remember about those JSON files. It contains the state value as its network id as well as the material. Right? 

If so, then you would create the default BlockData interface for your material, get the generators for the BlockData and then create the StateValue of the network ID from the generator and replace the default value in the BlockData's StateValue for that ID to the newly created StateValue. 

I havent fully got the network system in place yet, but you can see some of it such as 

    StateGenerator.deserialize(int networkId)

smartboyathome: I documented the structure of that file here using JSON Schema: https://github.com/GlowstoneMC/1.13-board/issues/28#issuecomment-575973492

Basically, each combination of property values is mapped to its own block ID. Within this file, all values are strings, so the most efficient way to find the block IDs would be to look them up based on the block type and a Map<String, String> of properties.

mosemister: Wouldn't it be more effectient to have a map value of Map<String, BlockData>? Im assuming when you mean Map<String, String> you mean the unique id that these property values and the mapped string is the typical string id, as in "minecraft:END_ROD{face=NORTH}"?

smartboyathome: No, in your example I literally mean the structure is (not real data from the file):

 ```
{
    "minecraft:end_rod": {
        "properties": {
            "face": ["NORTH", "SOUTH", "EAST", "WEST"]
        },
        "states": [
            {
                "properties": {
                    "face": "NORTH"
                 },
                 "id": 123,
                 "default": true
            },
            {
                "properties": {
                    "face": "SOUTH"
                 },
                 "id": 124
            },
            {
                "properties": {
                    "face": "EAST"
                 },
                 "id": 125
            },
            {
                "properties": {
                    "face": "WEST"
                 },
                 "id": 126
            },
        ]
    }
}
```

In my explanation above, the Map<String, String> corresponded to obj["minecraft:end_rod"].states[0].properties, for example. While, yes, I could parse this out all the way into BlockData, there would be quite a lot of operations being done just to figure out every state of every block (again, all combinations for each block are represented within this file). Instead, my code just parses it into the bare minimum needed for you to be able to match a block's state to its network ID

Comment (smartboyathome) [reply src/main/java/net/glowstone/block/data/state/StateGenerator.java]: Oh, and by the way, I couldn't even use the NBT functions to create the BlockData for each state in this file, since the file represents all values using JSON, not NBT. We'd need to maintain a separate function, just for unfolding this.

Comment (mastercoms) [reply src/main/java/net/glowstone/block/data/state/StateGenerator.java]: We have a Mojangson class that can help.

Comment (smartboyathome) [reply src/main/java/net/glowstone/block/data/state/StateGenerator.java]: The problem isn't with parsing this stuff, the problem is with translating the model from the raw form above into the BlockData that Bukkit expects. We'd be having to maintain a separate method full of switch cases for doing this mapping.

Comment (mosemister) [reply src/main/java/net/glowstone/block/data/state/StateGenerator.java]: Right. Ok, I can handle that

Comment (mosemister) [reply src/main/java/net/glowstone/block/data/state/StateGenerator.java]: Is this what you had in mind? @smartboyathome 

3efe441c7d4cde4ff57368231508b600dd8d46af

Edit:
Just seen it is too big to view the whole thing, here is the premis:

        builder.put("obj[minecraft:grass_block].state[0].properties", "minecraft:grass_block[snowy=true]");
        builder.put("obj[minecraft:grass_block].state[1].properties", "minecraft:grass_block[snowy=false]");
        builder.put("obj[minecraft:podzol].state[0].properties", "minecraft:podzol[snowy=true]");
        builder.put("obj[minecraft:podzol].state[1].properties", "minecraft:podzol[snowy=false]");

**Comment by mastercoms** [comment]
Hey, I just wanted to drop by and say I really like how this PR is going! I'll have some more feedback on it soon, but just wanted to let you know you're some awesome work!

File: src/main/java/net/glowstone/block/data/state/StateGenerator.java (hunk 5 above)
Comment (smartboyathome) [reply src/main/java/net/glowstone/block/data/state/StateGenerator.java]: Not quite what I had in mind. When I get a chance, I'll send you over the interfaces I was coding against. With this method, I am seriously concerned we'll end up past Java's 64KB block limit eventually.

**Comment by smartboyathome** [comment]
I saw you removed the WIP tag. I will see about checking out your code and reviewing it here soon. :)

**Comment by mosemister** [comment]
I mearly removed the [wip] due to all the blocks being added (except noteblock ... Still working out how to do that one). I still need to make sure the code compiles, make sure all the names are correct and all that. That wont change the code that much

**Comment by mosemister** [comment]
The updated context push fixes the noteblock and missed code. The only check i need to do now is to make sure the ids are as they are in mc

File: src/main/java/net/glowstone/block/data/state/StateUtil.java [hunk 6]
```diff
@@ -0,0 +1,21 @@
+package net.glowstone.block.data.state;
+
+import org.bukkit.block.BlockFace;
+
+public interface StateUtil {
+
+    BlockFace[] SIXTEEN_BLOCK_FACES = {BlockFace.SOUTH, BlockFace.SOUTH_SOUTH_WEST, BlockFace.SOUTH_WEST, BlockFace.WEST_SOUTH_WEST, BlockFace.WEST, BlockFace.WEST_NORTH_WEST, BlockFace.NORTH_WEST, BlockFace.NORTH_NORTH_WEST, BlockFace.NORTH, BlockFace.NORTH_NORTH_EAST, BlockFace.NORTH_EAST, BlockFace.EAST_NORTH_EAST, BlockFace.EAST, BlockFace.EAST_SOUTH_EAST, BlockFace.EAST, BlockFace.EAST_SOUTH_EAST, BlockFace.SOUTH_EAST, BlockFace.SOUTH_SOUTH_EAST};
```
smartboyathome: There are 18 block faces here, not 16. Due to this,w ill your logic actually work as intended?

smartboyathome: Looks like EAST and EAST_SOUTH_EAST are duplicated.

mosemister: Thanks for spotting that. 

From the cleanup ive been doing locally, Ive been focusing on the ids and default values of the blocks

mosemister:
Should be complete now. All the ids and defaults match. No implementations I have missed.

The network stuff can be removed if needed, ill leave that to whoever wants to take on that issue (it maybe me)

smartboyathome (approved):
Sorry for the delay in reviewing. I have been in discussions with @mastercoms on how we're approaching merging this PR. You did a lot of excellent work here, which is why I am approving it after a lengthy analysis of it. Seriously, great job with this!

Based on your work, I created an automatically generated version, based in part on your solution here, which you can find [here](https://github.com/smartboyathome/redstone-transformer), at least until the repo is created within the Glowstone org. Would love your input on this, to see if there's anything I missed. This is probably the version we will be moving forward with, since it covers a few different aspects of this and related issues, and is less laborious to maintain as we upgrade between versions, and it won't require generating a block report at runtime.

Mastercoms is currently figuring out how the bounty will be divided up so that you get paid for the substantial amount of work that went into this. Thank you for your patience with this PR.

mosemister:
I will take a deep look in a few days as I have stuff to do. You guys have been doing an amazing job with glowstone, just glad there is code based on my code that will be in it. 

Please take your time reviewing it, in no rush. 

As for my brief look on the your code, it looks amazing, easy enough to follow. As mentioned before, ill take a good look in a few days.

smartboyathome:
Closing, as its superseded by #1087.

Issue #975, opened 2018-11-08
Gradle building


Zaxar163:
Why Gradle is better then Maven? It has easy plugin interface and flexible build scripts. You can write custom tasks and include scripts...

mastercoms:
Although we can, do we need to? A declarative build system like Maven is simpler and gets the job done, fast. We don't need a build language to handle our builds.

c-edw:
There are more important things to be focusing on than the build system, in my opinion.

hazae41:
> Although we can, do we need to? A declarative build system like Maven is simpler and gets the job done, fast. We don't need a build language to handle our builds.

Simpler? Glowstone build is done with about 400 lines of Maven while it could be done in about 100 lines of Gradle

Fast? It has been proved that [Gradle](https://gradle.org/gradle-vs-maven-performance/) is between 10 and 100 times faster than Maven

Not knowing how to use it doesn't mean it is complicated ;)

aramperes:
@Hazae41 

- Simplicity is not the same as conciseness. Maven is just fine for what we need.
- We don't need the flexibility of a dynamic language like Groovy, in the current state of the project. Our current tools get the job done, so why change them for the sake of change?
- While a performance improvement would be welcome, we're satisfied with the current length of our builds. I'd argue we should focus on the architecture of our application rather than the build tools to improve that.

mastercoms:
Speaking of performance, Glowstone specifically was much slower on Gradle compared to Maven. Maybe gradle.org has a bit of bias?

smartboyathome:
Just wanted to add one other reason for sticking with Maven right now. Currently, we're using Maven's dependency resolution internally for our runtime library dependencies. If we moved to Gradle, this would probably also need to be moved as well in order to keep the dependency resolution consistent. between both systems. That would likely be a sizeable amount of work for very little gain, so I would also recommend against moving.

Pr0methean:
@Zaxar163 If you're still convinced we should be using Gradle, we'll consider a pull request with a complete implementation. But it must address @smartboyathome's comment (i.e. it must replace Naether with a Gradle-based counterpart).

aki-ks:
Linkstone uses our own Maven plugin. If we switch to Gradle and want to stay consistent between all repos, we would have to rewrite it.

gdude2002:
**WARNING:** Super big comment (10,000+ chars). Caveat emptor.

---

[...]

PR #1077, opened 2020-01-22
Move to Gradle

gdude2002:
This PR converts the build from Maven to Gradle. No doubt there will be a few teething issues - I'm currently working on getting the build working.

More info is available [in this issue](https://github.com/GlowstoneMC/Glowstone/issues/975#issuecomment-576940807).


gdude2002:
In case I forgot to say, by the way, this PR is ready for review.

The only thing I'm planning on looking at is the Azure build cache, and that's rumoured to not work properly at the moment.

yannicklamprecht:
What's the reason to switch to Gradle, when Maven is working fine?

gdude2002:
> What's the reason to switch to Gradle, when Maven is working fine?

There is a genuinely massive issue (#975) with that discussion in it - it's also linked in the initial comment of this PR.

gdude2002:
Some further work needs to be done here for 1.13 to function due to some _ridiculously_ large generated classes requiring some extra args to be passed to the compiler, something which is impossible with the current set of Gradle plugins available.

I'll probably have to write one, but it should be pretty simple. We'll see.

EDIT: Just haven't had time for this lately, but it'll get done.

gdude2002:
I am having too much trouble writing a Gradle plugin in a language that isn't Groovy, and the overall plugin architecture doesn't really make sense to me.

As it is currently impossible to work around the stack size requirement (even though we have no good reason to be hitting it) and I don't have the experience required to write a new Lombok plugin that fixes the problem, I'm closing this PR.

I'll leave the branch in place in case someone else wants to take a stab at it.

mastercoms:
I wouldn't recommend anyone working on it further for the time being. The statically generated class in 1.13 causing the stack overflow will eventually be replaced hopefully.

mastercoms:
The aforementioned class has been replaced, thanks @smartboyathome!

PR #1081, opened 2020-01-24
Remove duplicate options

mio-19:
these options are already in the configuration file

### DIFF
--- a/src/main/java/net/glowstone/GlowServer.java
+++ b/src/main/java/net/glowstone/GlowServer.java
@@ -564,45 +564,6 @@ public class GlowServer implements Server {
                 case "--configfile":
                     configFileName = args[++i];
                     break;
-                case "--port":
-                case "-p":
-                    parameters.put(Key.SERVER_PORT, Integer.valueOf(args[++i]));
-                    break;
-                case "--host":
-                case "-H":
-                    parameters.put(Key.SERVER_IP, args[++i]);
-                    break;
-                case "--onlinemode":
-                case "-o":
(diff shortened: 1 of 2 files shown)


mastercoms:
Hi, thank you for your contribution to Glowstone! I think it is useful for people to be able to override their configuration file using command line options. Are there specific benefits to removing this functionality, for example some sort of error or incompatibility?

mio-19:
@mastercoms I think people can always easily modify the configuration file, so this functionality is not necessary

aramperes:
This is useful for running Glowstone using scripts that can't edit YAML files. They are also supported in Spigot: https://www.spigotmc.org/wiki/start-up-parameters/

mio-19:
@momothereal Scripts can create a YAML configuration file or use command-line YAML processor

mastercoms:
Why should we force people to complicate their scripts with YAML processors (and in some cases, including YAML will not be available), for the benefit of a 50 lines of code cleanup and nanosecond level possible startup time improvement?

mio-19:
It's not for performance, it's for consistencies and simplicity. Some configurations can specify different values in two places, and some configurations can only be specified in one place

mastercoms:
While consistency in configuration is a noble cause, these options are commonly used in other server software (despite configuration being available), like CraftBukkit and Spigot. Specifically, many server panels and hosts rely on these options being available to easily manage their server hosting. I don't think we should break compatibility and consistency with other software for this kind of consistency. It doesn't seem valuable enough to me.

PR #1084, opened 2020-01-24
add GravityAccel to GlowAnimal

mio-19:


plexus:
I noticed animals spawned in mid-air do not fall. Does this fix that?

mastercoms:
Probably, but I haven't tested it in game yet, will do so tomorrow.

plexus:
Tried it now and it works! I was trying to generate falling chickens but they were stuck in mid-air, with this patch they fall down as they should :) thanks!

mastercoms (approved):

PR #1086, opened 2020-02-04
Fix checkstyle violations

plexus:
So the build is no longer red.


mastercoms:
The checkstyle rules should be upgraded instead, so that this code which conforms to our style passes checkstyle.

mastercoms (changes requested):
Alright, I investigated further and fixed the brace rule. However, the import spacing rule is apparently intended, since the checkstyle was intended to enforce no spaces between groups. I guess we will have to roll with it for now, as any changes to the import order rules is going to cause much more breakage (1000 files) instead of this relatively small fix.

So, could you please revert the brace changes, and then I'll merge the import changes. Thanks for contributing!

plexus:
I undid the brace changes and rebased. Thanks!

mastercoms (approved):

mastercoms:

PR #1087, opened 2020-02-24
BlockData implementation using new redstone-transformer libs

smartboyathome:
Supersedes #1076, as its based on that PR's code.

The [redstone-transformer](https://github.com/GlowstoneMC/redstone-transformer) library implements Bukkit's interfaces using the block report generated by Minecraft's vanilla server. This allows us to have an easier time maintaining this data, and especially the network IDs, which may change between versions. All materials are mapped by the aforementioned library, and exposed through BlockDataManager. I added this manager to GlowServer, and exposed a getter for other classes that need it. This does have its shortcomings, but my plan is to eventually refactor Glowstone to start to integrate Inversion of Control, making passing items like this easier.

This will implement GlowstoneMC/1.13-board#27 and GlowstoneMC/1.13-board#28.


src/main/java/net/glowstone/constants/GlowParticle.java
```diff
@@ -101,13 +99,15 @@
             return new Float[]{red, green, blue, size};
         });
         registerParticleDataHandler(ItemStack.class, PARTICLE_HANDLER_RETURN_SELF);
-        registerParticleDataHandler(BlockData.class, o -> {
+        // TODO: Figure out what the correct result for this is, since not all BlockData can be converted to a single
```
mastercoms: I thought every BlockData had an integer ID. For which cases did you find this not to be true?

smartboyathome: Fixed it in the latest commit.

mastercoms (approved):
Awesome work, thanks!

PR #1090, opened 2020-05-06
Fix potion effects display to self

SHADOWDANCH:
As players now have real entity id for self intead 0 (https://github.com/GlowstoneMC/Glowstone/commit/c1fa66060d1e9c5efafed1fcea6f22fa43f2b355) there is no need for separated packet with entity id 0. Also implement `LivingEntity#getPotionEffect(PotionEffectType)`

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -1112,12 +1112,7 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
         EntityEffectMessage msg = new EntityEffectMessage(getEntityId(), effect.getType().getId(),
                 effect.getAmplifier(), effect.getDuration(), effect.isAmbient());
         for (GlowPlayer player : world.getRawPlayers()) {
-            if (player == this) {
-                // special handling for players having a different view of themselves
-                player.getSession().send(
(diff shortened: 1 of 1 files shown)


aramperes:
Could you link some documentation that the entity ID changed? Was this done in 1.12 or later versions?

SHADOWDANCH:
> Could you link some documentation that the entity ID changed? Was this done in 1.12 or later versions?

Sorry for my poor english. Entity id not changed in minecraft. Before this commit https://github.com/GlowstoneMC/Glowstone/commit/c1fa66060d1e9c5efafed1fcea6f22fa43f2b355  in glowstone all players have entity id 0 for self at clientside. And after it players at clientside have real entity id like it on original minecraft server. And so after this commit there is no need for sending separated packet with 0 entity id

aramperes:
The client expects/expected an ID of 0 to refer to itself. This may have changed, but it would be good to know in which version this changed.

SHADOWDANCH:
> The client expects/expected an ID of 0 to refer to itself. This may have changed, but it would be good to know in which version this changed.

I am not sure in what version this changed. But 1.12.2 client not uses 0 id for self. It receives self id from server in `SPacketJoinGame`

aramperes:
I asked on IRC and you are right, it is preferable to use the real entity ID. Historically, ID 0 worked because from the client's perspective, its own character is the first entity in the worls (so 0 index).

aramperes (approved):

PR #1092, opened 2020-05-06
Implement Beacon logic

SHADOWDANCH:
Implement vanilla beacon logic and call of `BeaconEffectEvent`

### DIFF
--- a/src/main/java/net/glowstone/block/entity/state/GlowBeacon.java
+++ b/src/main/java/net/glowstone/block/entity/state/GlowBeacon.java
@@ -13,6 +13,8 @@ import org.bukkit.potion.PotionEffectType;
 
 public class GlowBeacon extends GlowContainer implements Beacon {
 
+    private static final int RADIUS_MULTIPLIER = 10;
+
     /**
      * Creates an entity for the given beacon block.
      *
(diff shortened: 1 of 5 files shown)


src/main/java/net/glowstone/block/entity/state/GlowBeacon.java
```diff
@@ -39,7 +28,7 @@ private BeaconEntity getBlockEntity() {
 
     @Override
     public Collection<LivingEntity> getEntitiesInRange() {
-        return Collections.emptyList();
+        return getWorld().getNearbyEntitiesByType(Player.class, getLocation(), getTier() * 10 + 10, getWorld().getMaxHeight());
```
aramperes: Could you extract `10` as a class constant?

src/main/java/net/glowstone/block/entity/state/GlowBeacon.java
```diff
@@ -28,7 +30,7 @@ private BeaconEntity getBlockEntity() {
 
     @Override
     public Collection<LivingEntity> getEntitiesInRange() {
-        return getWorld().getNearbyEntitiesByType(Player.class, getLocation(), getTier() * 10 + 10, getWorld().getMaxHeight());
+        return getWorld().getNearbyEntitiesByType(Player.class, getLocation(), getTier() * RADIUS_MULTIPLIER + 10, getWorld().getMaxHeight());
```
aramperes: Assuming the formula is equivalent to `(tier + 1) * RADIUS_MULTIPLIER`, it should be 
```suggestion
        return getWorld().getNearbyEntitiesByType(Player.class, getLocation(), getTier() * RADIUS_MULTIPLIER + RADIUS_MULTIPLIER, getWorld().getMaxHeight());
```

aramperes (approved):

mastercoms:

Issue #1095, opened 2020-05-07
Some cauldron-item interactions missing


emreozcan:
Title: Some cauldron-item interactions missing

Glowstone build: c4c59a7ebd263cf220a88382e882073bca7d267d

Interacting with a cauldron while holding an empty bucket does not fill the bucket. Also, interacting with a cauldron while holding a water bottle does not fill the cauldron. These are the expected results, as in a 1.12.2 single player world the interactions give these expected results.

PR #1097, opened 2020-05-08
Adds missing cauldron-item interactions.

emreozcan:
This PR adds the missing cauldron-item interactions listed in issue #1095:
- Cauldron &ndash; Empty bucket
- Cauldron &ndash; Water bottle

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
@@ -100,15 +100,19 @@ public class BlockCauldron extends BlockNeedsTool {
             return;
         }
         if (player.getGameMode() != GameMode.CREATIVE) {
-            Map<Integer, ItemStack> dummyInventory = player.getInventory()
-                    .addItem(new ItemStack(Material.WATER_BUCKET));
-            if (!dummyInventory.isEmpty()) {
-                player.getWorld()
-                        .dropItemNaturally(player.getLocation(), new ItemStack(Material.WATER_BUCKET));
+            if (player.getItemInHand().getAmount() == 1) {
+                player.getItemInHand().setType(Material.WATER_BUCKET);
+            } else {  // player has at least 2 empty buckets
+                Map<Integer, ItemStack> drops = player.getInventory()
+                        .addItem(new ItemStack(Material.WATER_BUCKET));
+                if (!drops.isEmpty()) {
+                    player.getWorld()
+                            .dropItemNaturally(player.getLocation(), new ItemStack(Material.WATER_BUCKET));
+                }
+                player.getItemInHand().setAmount(player.getItemInHand().getAmount() - 1);
             }
         }
 
-        player.getItemInHand().setAmount(player.getItemInHand().getAmount() - 1);
     }
 
     private void fillBottle(GlowPlayer player, GlowBlock block) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
```diff
@@ -61,36 +75,74 @@ public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face,
...
+        // fired when a player fills an empty bucket from the cauldron
+        if (block.getData() != LEVEL_FULL) {
+            return;
+        }
+        if (!setCauldronLevel(block, LEVEL_EMPTY, player,
+                CauldronLevelChangeEvent.ChangeReason.BUCKET_FILL)) {
+            return;
+        }
+        if (player.getGameMode() != GameMode.CREATIVE) {
+            Map<Integer, ItemStack> dummyInventory = player.getInventory()
```
mastercoms: Could we skip this if the bucket amount is 1, and just set the type to water bucket?

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
```diff
@@ -61,36 +75,74 @@ public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face,
...
+                    .addItem(new ItemStack(Material.WATER_BUCKET));
+            if (!dummyInventory.isEmpty()) {
+                player.getWorld()
+                        .dropItemNaturally(player.getLocation(), new ItemStack(Material.WATER_BUCKET));
             }
         }
+
+        player.getItemInHand().setAmount(player.getItemInHand().getAmount() - 1);
```
mastercoms: Is the bucket still consumed in creative mode?

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
mastercoms: Can we rename the variable to `drops`, for consistency?

mastercoms (changes requested):

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
emreozcan: Of course! Will comply in next commit.

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
emreozcan: Yes, oops. Will be fixed in next commit.

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
emreozcan: Certainly. Doing the same thing to [Line 124](https://github.com/GlowstoneMC/Glowstone/pull/1097/files/c0bebd5ec90cbbcce286c10c667b053ddf1c98a5#diff-b80570db289b0cfacd9df3c246576452R124) too.

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
```diff
@@ -36,11 +38,23 @@ public boolean blockInteract(GlowPlayer player, GlowBlock block, BlockFace face,
...
-                fillCauldron(player, block);
+                emptyBucket(player, block);
+                return true;
+
+            case BUCKET:
+                fillBucket(player, block);
+                return true;
+
+            case POTION:
+                emptyBottle(player, block);
```
mastercoms: Change emptyBottle to return boolean then change this to

```java
if (emptyBottle(player, block) {
  return true;
}
break;
...
} // switch statement end
return super.blockInteract(player, block, face, clickedLoc);
```

Return false if PotionType != WATER.

mastercoms (changes requested):

src/main/java/net/glowstone/block/blocktype/BlockCauldron.java
emreozcan: [Line 44](https://github.com/GlowstoneMC/Glowstone/pull/1097/files/c317b67b28b66bd0b06040f30cfe305cdb97a01b#diff-b80570db289b0cfacd9df3c246576452R44) already does this. Am I missing something?

mastercoms: I was suggesting an alternative to that logic, which is more extensible for the future if we have more edge case checks we need to do for each item type.

emreozcan: Okay, makes sense.

emreozcan:
The failed check is [caused by a file transfer/connectivity issue on Azure's end](https://dev.azure.com/glowstonemc/Glowstone/_build/results?buildId=525&view=logs&j=f93145ac-9471-53b7-dc09-8530b564757f&t=ad3b922e-7e52-5934-c164-59fdffd4c0f4&l=1438). Obviously I don't have access to retrying failed jobs but I hope someone who does can verify this isn't an issue with the PR.

mastercoms (approved):

PR #1098, opened 2020-05-08
Update EntityEffectMessage. Optimize PotionEffect's pulsing

SHADOWDANCH:
Changes in this PR:
- Updated `EntityEffectMessage` according to 1.12.2 protocol specification from [wiki.vg](https://wiki.vg/index.php?title=Protocol&oldid=14204#Entity_Effect)
- Removed updating of potion effects to clients every pulse. Duration of potion effect autimatically decreases on client. Server need to only remove it when duration time is out
- Added updating of `MetadataIndex.POTION_COLOR` and `MetadataIndex.POTION_AMBIENT` metadata values on potion effect added/removed
- Improvements to `EffectCommand`

### DIFF
--- a/src/main/java/net/glowstone/command/minecraft/EffectCommand.java
+++ b/src/main/java/net/glowstone/command/minecraft/EffectCommand.java
@@ -90,7 +90,7 @@ public class EffectCommand extends GlowVanillaCommand {
                 }
             }
 
-            int amplifier = 1;
+            int amplifier = 0;
             if (args.length >= 4 && args[3] != null) {
                 try {
                     amplifier = Integer.parseInt(args[3]);
@@ -109,9 +109,9 @@ public class EffectCommand extends GlowVanillaCommand {
                     commandMessages.getResourceBundle());
             for (GlowPlayer player : players) {
                 player.addPotionEffect(
-                    new PotionEffect(effectType, duration, amplifier, false, hideParticles));
+                    new PotionEffect(effectType, duration, amplifier, false, !hideParticles));
                 doneMessage.send(sender, effectType.getName(),
-                        effectType.getId(), amplifier, player.getName(), duration / 20);
+                        effectType.getId(), amplifier + 1, player.getName(), duration / 20);
             }
             return true;
         }
--- a/src/main/java/net/glowstone/entity/GlowLivingEntity.java
+++ b/src/main/java/net/glowstone/entity/GlowLivingEntity.java
@@ -53,6 +53,7 @@ import net.glowstone.util.RayUtil;
 import net.glowstone.util.SoundUtil;
 import net.glowstone.util.loot.LootData;
 import net.glowstone.util.loot.LootingManager;
+import org.bukkit.Color;
 import org.bukkit.EntityAnimation;
 import org.bukkit.EntityEffect;
 import org.bukkit.GameMode;
@@ -257,6 +258,10 @@ public abstract class GlowLivingEntity extends GlowEntity implements LivingEntit
      * The ticks an entity stands adjacent to fire and lava.
      */
     private int adjacentBurnTicks;
+    /**
+     * If potion effects of entity changed
+     */
+    private boolean potionEffectsChanged;
 
     /**
      * Creates a mob within the specified world.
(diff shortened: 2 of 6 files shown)


src/main/java/net/glowstone/command/minecraft/EffectCommand.java
```diff
@@ -109,9 +109,9 @@ public boolean execute(CommandSender sender, String label, String[] args,
                     commandMessages.getResourceBundle());
             for (GlowPlayer player : players) {
                 player.addPotionEffect(
-                    new PotionEffect(effectType, duration, amplifier, false, hideParticles));
+                    new PotionEffect(effectType, duration, amplifier, false, !hideParticles));
                 doneMessage.send(sender, effectType.getName(),
-                        effectType.getId(), amplifier, player.getName(), duration / 20);
+                        effectType.getId(), amplifier + 1, player.getName(), duration / 20);
```
mastercoms, line 114: This should just `amplifier` instead of `amplifier + 1`. The command is 0 to 255 inclusive, and potion effects start at 0. It is only the display on the client which is + 1, which does not need to be handled in the message.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -1109,8 +1107,10 @@ public boolean addPotionEffect(PotionEffect effect, boolean force) {
 
         potionEffects.put(effect.getType(), effect);
 
+        updatePotionEffectsMetadata();
```
mastercoms: Should the update calls be done in `pulse` since messages are held until then anyway, instead of in `addPotionEffect` and `removePotionEffect`?

src/main/java/net/glowstone/entity/GlowLivingEntity.java
```diff
@@ -1166,6 +1168,11 @@ public void clearActivePotionEffects() {
         }
     }
 
+    protected void updatePotionEffectsMetadata() {
+        metadata.set(MetadataIndex.POTION_COLOR, potionEffects.size() > 0 ? 1 : 0); //TODO: calculate potion particles color
```
mastercoms: I believe using `Color.mixColors()`, mapping each `potionEffect.getColor()` to an array, and then calling `Color#asRGB` on the resulting mixed color will work here.

mastercoms (changes requested):

src/main/java/net/glowstone/command/minecraft/EffectCommand.java
SHADOWDANCH: But in vanilla minecraft server this messages displays `amplifier + 1` and not from 0

src/main/java/net/glowstone/entity/GlowLivingEntity.java
SHADOWDANCH: I think there is no need for recalculation of potion effect color every pulse

mastercoms: There is also no need to recalculate potion effect color every time a potion is added (`addPotionEffects` or multiple `addPotionEffect` calls), before it is made relevant on the network. Perhaps we can queue an update on pulse using a boolean flag, `potionEffectsChanged`.

src/main/java/net/glowstone/entity/GlowLivingEntity.java
SHADOWDANCH: `Color#mixColors(Color...)` is not static. I'am not sure on what color i should call it? White?

mastercoms: Perhaps the first color?

mastercoms: With that, it would be preferable to no longer use streams and just have to use a  i=1 -> length for loop, unfortunately.

SHADOWDANCH: I retrieve base particle colors of all potion effect from bukkit with this code:
```java
for (PotionEffectType type : PotionEffectType.values()) {
    if (type != null) {
        System.out.println("type=" + type.getName() + ", color=" + type.getColor().asRGB());
    }
}
```
And then mix all colors with color of first effect of entity and result looks not like vanilla color.

Jump effect particle from glowstone:
glowstone_jump
Jump effect particle from vanilla:
vanilla_jump

Jump + strength effect particle from vanilla:
vanilla_strength+jump
Jump + strength effect particle from glowstone:
glowstone_strength+jump

Also tried mixing with white/black and still not very similar to vanilla

mastercoms: I'm not sure how else to mix it, we will have to get someone to document it for us from vanilla. But for now, this can be good enough functionality with a TODO

mastercoms (approved):

PR #1099, opened 2020-05-25
Implement WrittenBook open. Fix WritableBook sign

SHADOWDANCH:
Changes in this pull request:

- Signed books now can be opened
- Writable books now can be signed again

### DIFF
--- a/src/main/java/net/glowstone/block/itemtype/ItemWrittenBook.java
+++ b/src/main/java/net/glowstone/block/itemtype/ItemWrittenBook.java
@@ -1,7 +1,10 @@
 package net.glowstone.block.itemtype;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
 import net.glowstone.block.GlowBlock;
 import net.glowstone.entity.GlowPlayer;
+import net.glowstone.net.GlowBufUtils;
 import net.glowstone.net.message.play.game.PluginMessage;
 import org.bukkit.block.BlockFace;
 import org.bukkit.inventory.EquipmentSlot;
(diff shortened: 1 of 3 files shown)


src/main/java/net/glowstone/block/itemtype/ItemWrittenBook.java
```diff
@@ -30,6 +31,10 @@ public void rightClickBlock(
     }
 
     private void openBook(GlowPlayer player) {
-        player.getSession().send(new PluginMessage("MC|BOpen", EMPTY));
+        ByteBuf buf = Unpooled.buffer();
+        GlowBufUtils.writeHand(buf, player.getMainHand());
```
mastercoms: This VarInt isn't for left or right, but for main or offhand. You can use a book in your offhand if there is nothing in your main hand when you right click.

mastercoms (approved):

mastercoms (changes requested):
Apologies -- I meant to select request changes for the previous review.

src/main/java/net/glowstone/block/itemtype/ItemWrittenBook.java
```diff
@@ -20,19 +20,19 @@ public Context getContext() {
 
     @Override
     public void rightClickAir(GlowPlayer player, ItemStack holding) {
-        openBook(player);
+        openBook(player, EquipmentSlot.HAND);
```
mastercoms: I was still able to right click with my offhand in air. If this is due to a larger incomplete implementation of hand/offhand usage in Glowstone, please add a TODO.

mastercoms (changes requested):

mastercoms (approved):

PR #1105, opened 2021-01-27
Add GlowAdvancementDisplay flags

x4e:
Just implementing a TODO, added flag definitions for the AdvancementDisplay packet writing using wiki.vg.

### DIFF
--- a/src/main/java/net/glowstone/advancement/GlowAdvancement.java
+++ b/src/main/java/net/glowstone/advancement/GlowAdvancement.java
@@ -75,7 +75,7 @@ public class GlowAdvancement implements Advancement {
         }
         buf.writeBoolean(display != null);
         if (display != null) {
-            display.encode(buf);
+            display.encode(buf, true, true, false);
         }
         ByteBufUtils.writeVarInt(buf, criteriaIds.size());
         for (String criteriaId : criteriaIds) {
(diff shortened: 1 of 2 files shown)


aramperes (changes requested):
The TODO here wasn't just to name the flags, but to have a way to pass them when creating the message.

I think this could be 3 new boolean fields: `hasBackgroundTexture`, `showToast`, `hidden` which then get encoded as byte flags.

x4e:
@aramperes Thank You for clarifying, I have implemented this in 6ddaaadf.

I think there are probably some issues with the achievement display writing (specifically the icon slot field doesn't seem to be written correctly), but I will leave these for now since you said the dev branch is not really active.

x4e:
Checkstyle does not like my empty description for the `hidden` parameter - unfortunately I do not know the purpose of this parameter due to a lack of documentation on [wiki.vg](https://wiki.vg).

aramperes:
From [Minecraft Wiki](https://minecraft.gamepedia.com/Advancement/JSON_format#File_format):

> hidden: Can be true or false. Whether or not to hide this advancement and all its children from the advancement screen until this advancement have been completed. Has no effect on root advancements themselves, but still affects all their children. Defaults to false.

x4e:
@aramperes Thanks! Implemented in 5ec1a5a

aramperes (approved):

PR #1106, opened 2021-01-28
Allow blocks to fall when placed directly on a non solid block

x4e:
Previously placing a piece of sand on a torch for example would result in the sand not falling.
This change fixes that by considering all non solid blocks as not supporting the block, not just empty blocks.

### DIFF
--- a/src/main/java/net/glowstone/block/blocktype/BlockFalling.java
+++ b/src/main/java/net/glowstone/block/blocktype/BlockFalling.java
@@ -56,17 +56,11 @@ public class BlockFalling extends BlockType {
         }
     }
 
+    /**
+     * @param material A Material
+     * @return true if material will support this block and prevent it from falling or breaking
+     */
     private boolean supportingBlock(Material material) {
-        switch (material) {
-            case AIR:
-            case FIRE:
-            case WATER:
-            case STATIONARY_WATER:
-            case LAVA:
-            case STATIONARY_LAVA:
-                return false;
-            default:
-                return true;
-        }
+        return material.isSolid();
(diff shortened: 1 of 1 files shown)


src/main/java/net/glowstone/block/blocktype/BlockFalling.java
```diff
@@ -56,17 +56,11 @@ protected void transformToFallingEntity(GlowBlock me) {
         }
     }
 
+    /**
+     * @param material A Material
+     * @return true if material will support this block and prevent it from falling or breaking
+     */
     private boolean supportingBlock(Material material) {
```
aramperes, line 63: Is it worth just removing this function and replacing the usage with `material.isSolid()`?

x4e: A similar `supportingBlock` method is defined here: https://github.com/GlowstoneMC/Glowstone/blob/d72e7f0273085796273c3bb9086729fcd935f7ef/src/main/java/net/glowstone/entity/objects/GlowFallingBlock.java#L181
My thought was that keeping the function would clarify the difference between supporting a stationary falling block and a, well, falling falling block.
I think that the name of the function also makes the code a bit clearer and more descriptive.

Tbh there's not much of a reason to keep or remove the function, so let me know what you prefer and it's best we go wit hthat.

aramperes: I agree it makes the code clearer with this function. Nothing actionable

aramperes (approved):

PR #1107, opened 2021-01-29
Add vine physics

x4e:
This PR adds physics to vines.

This means that every time a block updates near a vine the vine will break if it is not below another vine **and** is not currently on a solid block.

Previously placing a vine on a block, then destroying the block would result in a floating vine. Same for hanging a vine off another vine and destroying the parent vine.


x4e:
Should be good to go now, only potential thing to look at in the future is support for vines being on top/bottom of blocks, this seems to have some sort of weird support in the vanilla edition that maybe should be replicated.

aramperes (approved):
Code looks good, however I'm not sure if this will be merged because there is a soft code-freeze on `dev` until we can merge the `1.16` branch.

The patch is fairly small so I'm personally not opposed to merging now. cc. @mastercoms

PR #1115, opened 2021-07-09
Added some missing tags to ArrowStore and ProjectileStore and added the new Int-Array representation of UUIDs to CompoundTag

VaiTon:
feat: added missing tags to ArrowStore and ProjectileStore
feat: added new Int-Array representation of UUIDs in CompoundTag


src/main/java/net/glowstone/util/nbt/CompoundTag.java
```diff
@@ -728,35 +727,53 @@ public boolean readString(@NonNls String key, Consumer<? super String> consumer)
     /**
      * Applies the given function to a UUID extracted from the given pair of long subtags, if they
      * both exist.
+     * <p>
```
mastercoms: use `<br>` or close the `<p>` tag.

mastercoms (changes requested):

VaiTon:
I'm reopening this with some more :)

mastercoms (approved):

PR #1116, opened 2021-07-11
Utilize RecipeManager and TagManager from Redstone Transformer

smartboyathome:
This PR swaps out the tags and recipe manager implementations for their counterparts from the redstone transformer. This allows us to get this data from the default datapack easily without keeping that data around and without having to hardcode these recipes. The implementations differ slightly from their Glowstone counterparts, but usually only in name. We may want to switch away from using the MaterialSetTag class to our own wrapper around SubTagTrackingTag in order to allow for modifications to these custom tags to be tracked in the future.


src/main/java/net/glowstone/inventory/crafting/CraftingManager.java
```diff
@@ -1,567 +0,0 @@
...
-        materials.forEach(wood -> furnaceFuels.put(wood, duration));
-    }
-
-    /**
-     * Load default recipes from built-in recipes.yml file.
-     */
-    @SuppressWarnings("unchecked")
-    private void loadRecipes() {
-        // Load recipes from recipes.yml file
-        InputStream in = getClass().getClassLoader().getResourceAsStream("builtin/recipes.yml");
```
mastercoms, line 483: We can delete this file.

mastercoms (changes requested):

mastercoms:
Also, I am not sure how relevant `materialValues.yml` and `fishingRewards.yml` will be once Redstone Transformer is done, but if they will still be relevant, could you add tag support to both of them?

smartboyathome:
materialValues.yml currently doesn't have an equivalent in the Redstone Transformer. The data this file contains isn't available from Mojang itself through data packs, but it looks like it is available in [PrismarineJS's minecraft-data](https://github.com/PrismarineJS/minecraft-data).

For fishingRewards.yml, this should eventually be replaced by the fishing loot table in Mojang's official data pack.

mastercoms:
@smartboyathome Could you resolve merge conflicts? Should be good after that.

mastercoms (approved):

PR #1117, opened 2021-07-12
Gradle 7.1

### LINKED ISSUE #975: Gradle building

VaiTon:
Closes https://github.com/GlowstoneMC/Glowstone/issues/975


build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+
+tasks.jar {
+    enabled = false
+    manifest {
+        attributes(
+                "Launcher-Agent-Class" to "net.glowstone.util.ClassPathAgent",
+                "Main-Class" to "net.glowstone.GlowServer",
+
+                "Implementation-Title" to project.name,
+                "Implementation-Version" to "${project.version}.git-${project.name}-${getGitHash()}",
```
smartboyathome, line 103: This seems fine for pipeline versions, but when developing, the version built could override the commit you're building upon? Basically, the issue Git doesn't assign a hash for uncommitted changes.

build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+
+    id("com.github.johnrengelman.shadow") version "7.0.0"
+}
+
+val buildExtras = System.getenv("BUILD_EXTRAS")?.toBoolean() ?: false
+val apiVersion = "1.16.5-R0.1-SNAPSHOT"
+
+repositories {
+    mavenLocal()
+    mavenCentral()
```
smartboyathome: We don't pull dependencies from Maven Central, our repo.glowstone.net repository instead proxies to it, sot his should probably be removed.

smartboyathome (changes requested):
Those two changes are my only gripes with this, but I would prefer if @gdude2002 also reviewed this.

build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+
+    maven("https://repo.glowstone.net/repository/maven-public/")
+    maven("https://repo.glowstone.net/repository/snapshots/")
+    maven("https://libraries.minecraft.net")
+}
+
+dependencies {
```
gdude2002, line 22: This won't do - use [version catalogs](https://docs.gradle.org/current/userguide/platforms.html#sub:version-catalog).

We make use of them at Kord Extensions, and it's much, much easier to manage deps this way. For example:

* [settings.gradle.kts](https://github.com/Kord-Extensions/kord-extensions/blob/root/settings.gradle.kts)
* [libs.version.toml](https://github.com/Kord-Extensions/kord-extensions/blob/root/libs.versions.toml)
* [build.gradle.kts](https://github.com/Kord-Extensions/kord-extensions/blob/root/kord-extensions/build.gradle.kts)

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-bin.zip
```
gdude2002: Nitpick, but may as well use 7.1.1.

gdude2002 (changes requested):
Pretty good, just a couple of things. You can consider this a "+1" to the other review comments as well.

build.gradle.kts
VaiTon: I have a question regarding this. How is the support of GitHub dependabot on this? I would prefer putting versions in settings.gradle.kts and let dependabot update them. What do you think?

build.gradle.kts
VaiTon: I don't seem to understand how could this be a problem :/

build.gradle.kts
gdude2002: Gradle support in dependabot is already fairly lacking, although I do wonder if it's that much of a deal in our case - I don't find myself getting many emails for it here, if any

VaiTon: I'm doing it! :P

build.gradle.kts
smartboyathome: It just could give someone the wrong idea about what version a given implementation was based on if it was built locally.

gdude2002 (approved):

smartboyathome (approved):

PR #1125, opened 2021-09-14
Inefficient Usages of Java Collections

FastAtlas:
Hi,

We find that there are several inefficient usages of Java Collections:

1. The contains method is invoked upon a list object. We recommend replacing it with a HashSet.
2. There is no iteration occurring upon a TreeMap and LinkedHashMap, thus the insertion order does not matter. We recommend replacing it with a HashMap.
3. ArrayList is inserted before an iteration, while multiple memory reallocation might occur when the size of the list exceeds its capacity. We recommend replacing it with a LinkedList.

We discovered the above inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

Bests

Ditto

### DIFF
--- a/src/main/java/net/glowstone/generator/objects/IceSpike.java
+++ b/src/main/java/net/glowstone/generator/objects/IceSpike.java
@@ -4,6 +4,8 @@ import org.bukkit.Material;
 import org.bukkit.World;
 import org.bukkit.block.Block;
 
+import java.util.Set;
+import java.util.HashSet;
 import java.util.Arrays;
 import java.util.Random;
 
@@ -25,6 +27,8 @@ public class IceSpike implements TerrainObject {
         }
         boolean succeeded = false;
         int stemRadius = Math.max(0, Math.min(MAX_STEM_RADIUS, tipRadius - 1));
+        Set<Material> materialSet = new HashSet<>(Arrays.asList(MATERIALS));
+        
         for (int x = -stemRadius; x <= stemRadius; x++) {
             for (int z = -stemRadius; z <= stemRadius; z++) {
                 int stackHeight = MAX_STEM_HEIGHT;
@@ -33,7 +37,7 @@ public class IceSpike implements TerrainObject {
                 }
                 for (int y = tipOffset - 1; y >= -3; y--) {
                     Block block = world.getBlockAt(sourceX + x, sourceY + y, sourceZ + z);
-                    if (Arrays.asList(MATERIALS).contains(block.getType())
+                    if (materialSet.contains(block.getType())
                         || block.getType() == Material.PACKED_ICE) {
                         block.setType(Material.PACKED_ICE);
                         stackHeight--;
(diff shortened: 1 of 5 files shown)


src/main/java/net/glowstone/util/GlowHelpMap.java
```diff
@@ -59,7 +58,7 @@
      */
     public GlowHelpMap(GlowServer server) {
         this.server = server;
-        helpTopics = new TreeMap<>(NAME_COMPARE);
+        helpTopics = new HashMap<>();
```
mastercoms, line 61: This is iterated upon in Bukkit (a Maven dependency of ours, see net.glowstone.glowkit), in HelpCommand.

src/main/java/net/glowstone/util/library/Library.java
```diff
@@ -157,8 +157,7 @@ public static Library fromConfigMap(Map<?, ?> configMap) {
      * @return A map that is able to be serialized into a config.
      */
     public Map<?, ?> toConfigMap() {
-        // Using LinkedHashMap to keep the props in order when written into the config file.
-        Map<String, Object> configMap = new LinkedHashMap<>();
+        Map<String, Object> configMap = new HashMap<>();
```
mastercoms, line 160: This isn't called within Glowstone (yet?), but the interface guarantees ordering for a config file.

src/main/java/net/glowstone/generator/objects/IceSpike.java
```diff
@@ -25,6 +27,8 @@ public boolean generate(World world, Random random, int sourceX, int sourceY, in
         }
         boolean succeeded = false;
         int stemRadius = Math.max(0, Math.min(MAX_STEM_RADIUS, tipRadius - 1));
+        Set<Material> materialSet = new HashSet<>(Arrays.asList(MATERIALS));
```
mastercoms, line 30: Would much rather have the MATERIALS variable be directly converted to a HashSet.

src/main/java/net/glowstone/generator/objects/trees/RedwoodTree.java
```diff
@@ -45,6 +47,7 @@ protected final void setLeavesHeight(int leavesHeight) {
 
     @Override
     public boolean canPlace(int baseX, int baseY, int baseZ, World world) {
+        Set<Material> overridableSet = new HashSet<>(overridables);
```
mastercoms, line 50: I think we should convert the base class' overridables to a HashSet, to prevent allocations every call.

mastercoms (changes requested):

mastercoms:
Also, can the CLA be signed by @DittoTool?

src/main/java/net/glowstone/util/GlowHelpMap.java
```diff
@@ -28,7 +28,6 @@
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.TreeSet;
```
mastercoms, line 31: Seems like TreeSet was removed instead of TreeMap? TreeSet is still used, while the usage of TreeMap was removed.

src/main/java/net/glowstone/util/library/Library.java
FastAtlas: Yeah. Our tool finds that the order is actually not used because the map object is not traversed. Of course, you can maintain it as a LinkedHashMap if you want to assure the extensibility of the program. For the current version, the HashMap is a better choice.

src/main/java/net/glowstone/generator/objects/IceSpike.java
FastAtlas: Yes. I agree with you.

src/main/java/net/glowstone/generator/objects/trees/RedwoodTree.java
FastAtlas: You can redesign the class to achieve better optimization. Currently, such allocation also improves the efficiency of the single function indeed.

src/main/java/net/glowstone/util/GlowHelpMap.java
FastAtlas: Sorry, I will update it. It is my fault.

src/main/java/net/glowstone/generator/objects/IceSpike.java
imDaniX: EnumSet instead of HashSet?

PR #1126, opened 2021-10-03
Changed port validator to allow all valid ports

matthi4s:
Currently, only ports below 49152 are accepted, otherwise, Glowstone falls back to 25565. This produces this warning:

`20:08:46 [WARNING] Invalid config value for 'server.port' (54224), resetting to default (25565)`

While the ports 49152 to 65535 are considered private/dynamic ports and probably aren't ideal to use under many circumstances, they are still valid ports and should be accepted by Glowstone.

At Aternos, we use a large port range including that private range to reduce the chance of conflicting ports while trying to keep the server port consistent for the user.


mastercoms (commented):
Thanks, looks good! Could you rebase this onto 1.12 so we can deploy it to currently running servers?

PR #1131, opened 2021-12-12
ref: use default GlowKit BoundingBox & nullability annotations

VaiTon:
1. Use the default GlowKit/Bukkit implementation of BoundingBox instead of using a custom one.
2. Added some more nullability annotations.


src/main/java/net/glowstone/GlowWorld.java
```diff
@@ -15,24 +15,11 @@
...
 import net.glowstone.chunk.GlowChunk.Key;
 import net.glowstone.chunk.GlowChunkSnapshot.EmptySnapshot;
-import net.glowstone.constants.GameRules;
-import net.glowstone.constants.GlowBiome;
-import net.glowstone.constants.GlowBiomeClimate;
-import net.glowstone.constants.GlowEffect;
-import net.glowstone.constants.GlowParticle;
-import net.glowstone.constants.GlowSound;
-import net.glowstone.constants.GlowTree;
+import net.glowstone.constants.*;
```
mastercoms: No wildcard imports.

mastercoms (changes requested):
Everything else looks good to me, though I'll have to do a second pass later

mastercoms (changes requested):
Are the java. import ordering changes correct?

VaiTon:
@mastercoms they should be as I'm using checkstyle. I'm merging the conflicts!

PR #1133, opened 2021-12-18
Updated GlowSheep.java imports to pass maven checkstyle plugin

dogeared:
checkstyle was failing because of unused import and wildcard import.


mastercoms:
Thank you!

PR #1147, opened 2025-04-23
Fix: Add placeholder for missing API event triggers (WIP)

vegaale313:
## Overview

This PR introduces a placeholder file (`TODOEventHooks.java`) to begin documenting and organizing missing event trigger logic within Glowstone’s Bukkit-compatible API. Based on issue [#922](https://github.com/GlowstoneMC/Glowstone/issues/922), several gameplay interactions currently do not fire the expected events (e.g., PlayerMoveEvent, InventoryClickEvent), causing incompatibility with plugins that rely on these hooks.

## Motivation

The goal of this commit is to establish a central point to track these gaps and gradually improve coverage and consistency with Bukkit behavior. This file will act as a staging ground for identifying missing hooks and associating them with relevant gameplay flows before full implementations are made.

## Next Steps

- Map in-game actions to expected Bukkit events
- Locate appropriate injection points in the event pipeline
- Gradually implement and test new hooks

### DIFF
--- a/src/main/java/net/glowstone/TODOEventHooks.java
+++ b/src/main/java/net/glowstone/TODOEventHooks.java
@@ -1,0 +1,13 @@
+package net.glowstone;
+
+/**
+ * Temporary class to track missing Bukkit event hook points.
+ * 
+ * TODO: Hook PlayerInteractEvent in entity.PlayerInteractionManager
+ * TODO: Add InventoryClickEvent trigger inside inventory.InventoryMonitor
+ * 
+ * This class is for internal planning use only and will be removed after integration.
+ */
+public class TODOEventHooks {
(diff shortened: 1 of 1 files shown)


mastercoms:
Hi, thanks for your contribution and writeup! I'm open to this idea, are we thinking this class will be constructed at those entry points so we can find usages later on and figure out how to integrate the event? I feel like most of the workload is associated with that locating places to put the events, and actually creating and using an event has been pretty straightforward. So maybe you're proposing to do planning in the code file structure? In my opinion it's probably better documented in the issue you linked, to post planning regarding where an event generally should be going. I'm not entirely sure if I understand entirely so I would appreciate some more clarification and apologize for not getting it.

mastercoms:

Issue #503, opened 2017-06-24
[#396] Entities: Pathfinding [$50]


gdude2002:
[![Bountysource bounty](https://api.bountysource.com/badge/issue?issue_id=46548046)](https://www.bountysource.com/issues/46548046-396-entities-pathfinding)

This issue is part of [this story](https://github.com/GlowstoneMC/Glowstone/issues/396), *Completing entities in Glowstone* (#396). To make the work easier and bounties fairer, we've split it into multiple issues.

Anyone can participate in the completion of this issue, which will be a major turning point in the development and the possible use-cases for Glowstone as a Minecraft server alternative.

---

**Before getting started!** If you are unfamiliar with Glowstone, please remember that Glowstone is a complete rewrite of the Minecraft vanilla server, aimed to be a better alternative to other software like Spigot. You are by no means allowed to push code to this repository based on Mojang's own platform (whether it be Forge, MCP, decompiled binaries, etc.) Doing so will cause in the rejection of your Pull Request. You are allowed to use other resources like the numerous wikis that exist on the Internet (Minecraft Wiki, wiki.vg, etc.) as well as trial-and-error.

---

Pathfinding
========

Glowstone currently lacks a 3D pathfinding tool for entities to reach a specific target. Keep in mind that dozens of entities may be running pathfinding actions simultaneously, which may hinder the performance on the server. This pathfinding algorithm would need to be able to not affect performance too much, while being a rough equivalent to the Vanilla functionality.

* Decide on an appropriate pathing algorithm, which may be A* (currently used in Vanilla) or one of its many derivatives. Based on my personal work and research on this, A* seems much more efficient and appropriate for 2D environments, but less so with layered 3D environments. The decided algorithm would have to be implemented with the concept of a 3D voxel world in mind (not a closed environment!)

Salil999:
Interested in working on this. If A* is used, what heuristic are we looking at? Manhattan distance is what I'm presuming...

aramperes:
I would think that Euclidean distance heuristic would be used. Even though the world is just a set of cubes, mobs (entities subject to Pathfinding) don't have to follow the grid. They can go from A to B diagonally.

smartboyathome:
[...]

Issue #502, opened 2017-06-24
[#396] Entities: AI [$65]


gdude2002:
[![Bountysource bounty](https://api.bountysource.com/badge/issue?issue_id=46548045)](https://www.bountysource.com/issues/46548045-396-entities-ai)

This issue is part of [this story](https://github.com/GlowstoneMC/Glowstone/issues/396), *Completing entities in Glowstone* (#396). To make the work easier and bounties fairer, we've split it into multiple issues.

Anyone can participate in the completion of this issue, which will be a major turning point in the development and the possible use-cases for Glowstone as a Minecraft server alternative.

---

**Before getting started!** If you are unfamiliar with Glowstone, please remember that Glowstone is a complete rewrite of the Minecraft vanilla server, aimed to be a better alternative to other software like Spigot. You are by no means allowed to push code to this repository based on Mojang's own platform (whether it be Forge, MCP, decompiled binaries, etc.) Doing so will cause in the rejection of your Pull Request. You are allowed to use other resources like the numerous wikis that exist on the Internet (Minecraft Wiki, wiki.vg, etc.) as well as trial-and-error.

---

Entity AI
======

Entity AI (or Entity Tasking) makes up the core of the interactions between the entities, the world and the player. They may consist of multiple triggers and will result in certain actions depending on the entity type. Entities may also execute random actions. These triggers and actions need to be implemented, in order to be on an equivalent level as the Vanilla server.

For example, entities may randomly wander around, look around, find targets, etc.

phylib:
Hi,
I would like to implement movements for chicken. Could you give me an advice on where to start? Should this be integrated in the `GlowChicken.pulse(..)` Method?
Thanks
Phylib

mastercoms:
Hi, thank you for your interest in Glowstone AI!

We currently have an Entity AI package here: https://github.com/GlowstoneMC/Glowstone/tree/dev/src/main/java/net/glowstone/entity/ai.

States essentially determine what tasks should be performed. Tasks run in a tick.

All GlowLivingEntitys have a TaskManager instance, in a field called taskManager

waterWang:
I claim this bounty. I will implement the Entity AI system with additional tasks (wandering, melee attack, fleeing) and register them for all monster types and animals.

waterWang:
I have submitted a PR (#1162) that implements the Entity AI system with:
- WanderTask: random wandering movement
- MeleeAttackTask: hostile mob attack behavior
[...]

PR #1155, opened 2026-05-16
Implement entity AI system with pathfinding integration

singhvishalkr:
I noticed the entity AI system (#502) and pathfinding (#503) issues while looking at Glowstone. The existing framework was solid but needed more behaviors and pathfinding integration, so I went ahead and implemented them.

The AI now has priority-based task execution. Tasks run in order from 0 (critical survival like swimming) through 10 (idle behaviors like looking around). This means a drowning cow will swim before it decides to look at a player.

New behaviors added:
- WanderTask picks a random walkable spot within 10 blocks and pathfinds there
- SwimTask applies upward velocity when submerged
- PanicTask makes animals flee when hurt
- MeleeAttackTask handles combat movement and attack cooldowns
- AvoidEntityTask makes passive mobs flee from monsters and sprinting players
- FindTargetTask scans for players within line of sight

I reworked the A* implementation to use proper f-score ordering with Manhattan-based heuristics. Added iteration limits to prevent server hangs on impossible paths. Also wrote a Jump Point Search variant that skips redundant nodes in open terrain.

Zombies now wander when idle instead of standing still, switch to targeting state when they spot a player, and use pathfinding for movement. Animals panic when attacked and avoid monsters.

I tested this manually by spawning zombies and watching them chase me through obstacles. The pathfinding handles elevation changes and goes around obstacles correctly. Performance stayed reasonable even with 20+ entities pathfinding simultaneously.

### DIFF
--- a/src/main/java/net/glowstone/entity/GlowAnimal.java
+++ b/src/main/java/net/glowstone/entity/GlowAnimal.java
@@ -42,8 +42,14 @@ public class GlowAnimal extends GlowAgeable implements Animals {
     public GlowAnimal(Location location, EntityType type, double maxHealth) {
         super(location, type, maxHealth);
         if (type != null) {
+            EntityDirector.registerEntityMobState(type, MobState.IDLE, "swim");
+            EntityDirector.registerEntityMobState(type, MobState.IDLE, "avoid_entity");
+            EntityDirector.registerEntityMobState(type, MobState.IDLE, "panic");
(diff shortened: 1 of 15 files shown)


singhvishalkr:
Light check-in: branch is ready on my side for the bounty-scoped AI work. If something needs to shift to match how you want the server to evolve, say the word and I will adjust.

mastercoms:

PR #1158, opened 2026-05-17
Allow chests next to existing double chests

decade10:
## What changed
- Allow a new chest to be placed next to an existing double chest without trying to attach it as a third chest.
- Keep the old guard against connecting to more than one attachable single chest.
- Add BlockChest placement tests for the double-chest adjacency case and the two-single-chest rejection case.

Addresses GlowstoneMC/1.13-board#20.

## Testing
- Added unit coverage in BlockChestTest.
- Local Gradle test execution was blocked by the repository's Gradle 6.1/Groovy compatibility with the installed Java runtime.


mastercoms:

PR #1159, opened 2026-05-18
feat: allow chest placement adjacent to double chests (fix #20)

malsony:
This change modifies BlockChest.java to allow placing chests directly next to double chests without requiring a block between them, as requested in issue #20.


mastercoms:

PR #1161, opened 2026-05-25
Allow chests next to existing double chests

duongynhi000005-oss:
Fixes GlowstoneMC/1.13-board#20.\n\n- allow single chests beside existing double chests\n- preserve triple-chest guard\n- add regression tests


mastercoms:

